PR #20, opened 2024-11-11
feat(airbyte-cdk): add gzipjson decoder

artem1205:
## What

resolving amazon ads low-code async migration

add custom decoder to implement gzip jsonl decoder

## How

add custom decoder

## Review guide

1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml`
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py`
3.  `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py`

- [ ] YES 💚
- [ ] NO ❌

## Summary by CodeRabbit

- **New Features**
	- Introduced `CustomDecoder` and `GzipJsonDecoder` for enhanced data decoding capabilities.
	- Updated retrievers to utilize new decoding strategies, improving flexibility in data extraction.
	- Expanded `ModelToComponentFactory` to support new decoders.

- **Bug Fixes**
	- Improved error handling in the authentication process within the component factory.

- **Tests**
	- Added tests for `GzipJsonDecoder` to ensure proper handling of gzip-compressed JSON responses.

### DIFF
--- a/airbyte_cdk/sources/declarative/decoders/__init__.py
+++ b/airbyte_cdk/sources/declarative/decoders/__init__.py
@@ -3,9 +3,9 @@
 #
 
 from airbyte_cdk.sources.declarative.decoders.decoder import Decoder
-from airbyte_cdk.sources.declarative.decoders.json_decoder import JsonDecoder, JsonlDecoder, IterableDecoder
+from airbyte_cdk.sources.declarative.decoders.json_decoder import JsonDecoder, JsonlDecoder, IterableDecoder, GzipJsonDecoder
 from airbyte_cdk.sources.declarative.decoders.noop_decoder import NoopDecoder
 from airbyte_cdk.sources.declarative.decoders.pagination_decoder_decorator import PaginationDecoderDecorator
 from airbyte_cdk.sources.declarative.decoders.xml_decoder import XmlDecoder
 
-__all__ = ["Decoder", "JsonDecoder", "JsonlDecoder", "IterableDecoder", "NoopDecoder", "PaginationDecoderDecorator", "XmlDecoder"]
+__all__ = ["Decoder", "JsonDecoder", "JsonlDecoder", "IterableDecoder", "GzipJsonDecoder", "NoopDecoder", "PaginationDecoderDecorator", "XmlDecoder"]
--- a/airbyte_cdk/sources/declarative/decoders/json_decoder.py
+++ b/airbyte_cdk/sources/declarative/decoders/json_decoder.py
@@ -33,11 +33,15 @@ class JsonDecoder(Decoder):
             body_json = response.json()
             yield from self.parse_body_json(body_json)
         except requests.exceptions.JSONDecodeError:
-            logger.warning(f"Response cannot be parsed into json: {response.status_code=}, {response.text=}")
+            logger.warning(
+                f"Response cannot be parsed into json: {response.status_code=}, {response.text=}"
+            )
             yield {}
 
     @staticmethod
-    def parse_body_json(body_json: Mapping[str, Any] | list) -> Generator[Mapping[str, Any], None, None]:
+    def parse_body_json(
+        body_json: Mapping[str, Any] | list,
+    ) -> Generator[Mapping[str, Any], None, None]:
         if not isinstance(body_json, list):
             body_json = [body_json]
         if len(body_json) == 0:
--- a/airbyte_cdk/sources/declarative/decoders/json_decoder.py
+++ b/airbyte_cdk/sources/declarative/decoders/json_decoder.py
@@ -5,7 +5,7 @@
 import logging
 from dataclasses import InitVar, dataclass
 from gzip import decompress
-from typing import Any, Generator, Mapping, MutableMapping, List
+from typing import Any, Generator, Mapping, MutableMapping, List, Optional
 
 import requests
 from airbyte_cdk.sources.declarative.decoders.decoder import Decoder
@@ -86,7 +86,7 @@ class JsonlDecoder(Decoder):
 
 @dataclass
 class GzipJsonDecoder(JsonDecoder):
-    encoding: str = "utf-8"
+    encoding: Optional[str] = "utf-8"
 
     def decode(self, response: requests.Response) -> Generator[MutableMapping[str, Any], None, None]:
         raw_string = decompress(response.content).decode(encoding=self.encoding)
--- a/airbyte_cdk/sources/declarative/decoders/json_decoder.py
+++ b/airbyte_cdk/sources/declarative/decoders/json_decoder.py
@@ -1,11 +1,11 @@
 #
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
-
+import codecs
 import logging
 from dataclasses import InitVar, dataclass, field
 from gzip import decompress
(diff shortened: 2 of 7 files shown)


aaronsteers:
Hi, @artem1205 - sorry for the mixup. New slash command is "/autofix" as you found. :)

airbyte_cdk/sources/declarative/decoders/json_decoder.py
```diff
@@ -70,8 +77,17 @@ class JsonlDecoder(Decoder):
...
-    def decode(self, response: requests.Response) -> Generator[Mapping[str, Any], None, None]:
+    def decode(self, response: requests.Response) -> Generator[MutableMapping[str, Any], None, None]:
         # TODO???: set delimiter? usually it is `\n` but maybe it would be useful to set optional?
         #  https://github.com/airbytehq/airbyte-internal-issues/issues/8436
         for record in response.iter_lines():
             yield orjson.loads(record)
+
+
+@dataclass
+class GzipJsonDecoder(JsonDecoder):
```
maxi297, line 88: Should we make this more generic? It feels like we should have a GzipDecoder that takes another decoder as once decompressed, the format could be anything. The issue right now is that the interface relying on `request.Response`. Hence, it feels like there is an abstraction or a concept missing...

artem1205: > Should we make this more generic?

I'm not sure about it. All implementations are inherited from `Decoder`, which has only 1 method `decode`

>[!NOTE]   
 Decoder strategy to transform a requests.Response into a Mapping[str, Any]

We _could_ make the decoder truly generic by wrapping the response content in a byte stream (e.g., `BytesIO`). However, this would require implementing custom iterators for `lines` or `content`, significantly increasing code complexity.

maxi297: The downside is that we will have to create all possible combinations of decompression and format. I'm fine for leaving this as is for now but it doesn't seem like something that will scale though

airbyte_cdk/sources/declarative/decoders/__init__.py
```diff
@@ -3,9 +3,9 @@
 #
 
 from airbyte_cdk.sources.declarative.decoders.decoder import Decoder
-from airbyte_cdk.sources.declarative.decoders.json_decoder import JsonDecoder, JsonlDecoder, IterableDecoder
+from airbyte_cdk.sources.declarative.decoders.json_decoder import JsonDecoder, JsonlDecoder, IterableDecoder, GzipJsonDecoder
```
maxi297, line 6: nit: shouldn't those imports be in alphabetical order?

airbyte_cdk/sources/declarative/migrations/legacy_to_per_partition_state_migration.py
```diff
@@ -47,8 +52,10 @@ def __init__(
             self._cursor.cursor_field, parameters=self._parameters
         ).eval(self._config)
 
-    def _get_partition_field(self, partition_router: SubstreamPartitionRouter) -> str:
-        parent_stream_config = partition_router.parent_stream_configs[0]
+    def _get_partition_field(
+        self, partition_router: CustomPartitionRouter | SubstreamPartitionRouter
+    ) -> str:
+        parent_stream_config = partition_router.parent_stream_configs[0]  # type: ignore # custom partition will introduce this atribute if needed
```
maxi297, line 58: [This typing](https://github.com/airbytehq/airbyte-python-cdk/pull/20/files#diff-77adc5cbcd3c74a99502b696f8828068335505eaf9a9342db3218e229bad988dR39) does not make sense to me if we are to ignore there error here. Should we instead fail if the partition router is not of type `SubstreamPartitionRouter` to have a clear error for the dev to take action? Are there cases today where the `partition_router` is not a `SubstreamPartitionRouter`? 

Outside of the scope of this PR: It seems like this wouldn't work in the following cases:
* if it were to be used with `ListPartitionRouter`
* if the `parent_stream_config.partition_field` requires interpolation
* if the first parent is not the "right" parent

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -620,10 +629,10 @@ def create_legacy_to_per_partition_state_migration(
             )
 
         return LegacyToPerPartitionStateMigration(
-            declarative_stream.retriever.partition_router,
-            declarative_stream.incremental_sync,
+            partition_router,
+            declarative_stream.incremental_sync,  # type: ignore # was already checked. Migration can be applied only to incremental streams.
```
maxi297, line 633: Where was this checked? I only see if statements on `declarative_stream.retriever.partition_router`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -620,10 +629,10 @@ def create_legacy_to_per_partition_state_migration(
...
             config,
-            declarative_stream.parameters,
+            declarative_stream.parameters,  # type: ignore
```
maxi297, line 635: Why do we need to ignore here?

maxi297 (commented):
A couple of comments around typing but the rest seems fine

airbyte_cdk/sources/declarative/decoders/__init__.py
artem1205: I guess this will be fixed later

https://github.com/airbytehq/airbyte-python-cdk/blob/1e0f777924a902e6e8d2909992459d1188c9e6e6/pyproject.toml#L105

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
artem1205: different types are expected there `Mapping[str, Any] != Dict[str, Any]`

airbyte_cdk/sources/declarative/decoders/__init__.py
maxi297: Ok, that explains why. Thanks for the context! I'll resolve this comment

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: Yeah, the difference in typing is annoying... I'm fine with this "ignore" then. Can we add a comment to explain this?

artem1205: added

airbyte_cdk/sources/declarative/migrations/legacy_to_per_partition_state_migration.py
artem1205: agreed, reverted to `SubstreamPartitionRouter` for consistency.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
artem1205: added check for this

maxi297 (approved):
LGTM! :shipit:

PR #32, opened 2024-11-12
chore(refactor): Remove Partition.close

maxi297:
## What
Work as part of https://github.com/airbytehq/airbyte-internal-issues/issues/10552

Following [this conversation](https://github.com/airbytehq/airbyte/pull/45180#discussion_r1767385234), we have moved part of the state management to the concurrent read processor. However, the close is still done as part of the partition.  The reason this is annoying is that it adds a dependency between the partitions and the cursor and now every time there is a change in the cursor, it might affect the partitions.

## How
We can remove this dependency by calling `stream.cursor.close_partition(...)` instead of `partition.close` as part of the concurrent read processor.

## Release Notes

- **New Features**
	- Enhanced error handling in file-based stream processing, improving visibility into cursor state during exceptions.
	- Streamlined partition management by centralizing closure logic within the stream's cursor.

- **Bug Fixes**
	- Adjusted tests to ensure correct partition closure behavior and error handling.

- **Tests**
	- Updated test cases for `FileBasedStreamPartition` and `StreamPartition` to reflect changes in cursor management.
	- Added new parameterized tests to improve coverage of cursor behavior in file synchronization scenarios.


airbyte_cdk/sources/streams/concurrent/partitions/partition.py
```diff
@@ -40,21 +40,6 @@ def stream_name(self) -> str:
...
 
-    @abstractmethod
-    def close(self) -> None:
-        """
-        Closes the partition.
-        """
-        pass
-
-    @abstractmethod
-    def is_closed(self) -> bool:
```
brianjlai, line 51: i did a very quick search of our cdk repo and never saw an usage of this method. What was the original intention of having this method?

brianjlai (approved):
changes look good to me. I'm mainly just curious why we needed an `is_closed()` method in the first place.

PR #39, opened 2024-11-12
chore(refactor): refactor partition generator to take any stream slicer

maxi297:
## What
Work as part of https://github.com/airbytehq/airbyte-internal-issues/issues/10552

In order to easily use any type of stream slicing that is done in the Declarative CDK as part of the Concurrent CDK, we need the `StreamSlicer` abstraction to be available in the Concurrent CDK.

## How
* Have `airbyte_cdk.sources.streams.concurrent.Cursor` implement `StreamSlicer` and update `ConcurrentCursor` to generate `StreamSlice`s
* Replace `CursorPartitionGenerator` by a more generic partition generator that takes any stream slicer

## Release Notes

- **New Features**
  - Introduced a structured approach to partition management with new classes for `DeclarativePartitionFactory` and `StreamSlicerPartitionGenerator`.
  - Enhanced stream slicing capabilities with the addition of the `StreamSlicer` class for improved state checkpointing.

- **Improvements**
  - Updated type hints across various classes for better clarity.
  - Refined error handling and state management in the cursor and partition classes.
  - Simplified hashing logic for stream slices, improving performance.

- **Bug Fixes**
  - Improved handling of cursor states and slice boundaries, ensuring robustness in concurrent processing.

- **Tests**
  - Expanded test coverage for key components, including new tests for `DeclarativePartitionFactory` and enhanced scenarios for cursor handling.
  - Updated tests to reflect changes in method signatures and expected behaviors.


airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
```
maxi297, line 1: TODO: unit tests

airbyte_cdk/sources/streams/concurrent/cursor.py
```diff
@@ -299,7 +301,7 @@ def ensure_at_least_one_state_emitted(self) -> None:
         """
         self._emit_state_message()
 
-    def generate_slices(self) -> Iterable[Tuple[CursorValueType, CursorValueType]]:
+    def stream_slices(self) -> Iterable[StreamSlice]:
```
maxi297, line 304: This change is going to be breaking for source-salesforce as the slicing relies on `generate_slices`

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -3,7 +3,7 @@
 #
 
 import logging
-from typing import Any, Generic, Iterator, List, Mapping, Optional, Tuple, Union
+from typing import Any, Generic, Iterator, List, Mapping, Optional, Tuple, Union, Callable
```
maxi297, line 6: @aaronsteers is it expected that `poetry run ruff format` does not order the imports?

maxi297: Ignore that, it seems like I didn't have the right ruff command

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
```diff
@@ -0,0 +1,66 @@
...
+from airbyte_cdk.sources.streams.concurrent.partitions.stream_slicer import StreamSlicer
+from airbyte_cdk.sources.types import StreamSlice
+from airbyte_cdk.utils.slice_hasher import SliceHasher
+
+
+class DeclarativePartitionFactory:
+    def __init__(self, stream_name: str, json_schema: Mapping[str, Any], retriever: Retriever, message_repository: MessageRepository) -> None:
+        self._stream_name = stream_name
+        self._json_schema = json_schema
+        self._retriever = retriever  # FIXME: it should be a retriever_factory here to ensure that paginators and other classes don't share interal/class state
```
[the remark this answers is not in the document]
maxi297: I've fixed that by adding a `retriever_factory` instead of the retriever directly. I'll add a comment to explain the reasoning.

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        stream_name: str,
+        json_schema: Mapping[str, Any],
+        retriever_factory: Callable[[], Retriever],
+        message_repository: MessageRepository,
+    ) -> None:
+        """
+        The DeclarativePartitionFactory takes a retriever_factory and not a retriever directly. The reason is that out components are not
+        thread safe and classes like `DefaultPaginator` would not behave the way we want if multiple threads were to call their methods.
+        In order to avoid these problems, we will create one retriever per thread which should make the processing thread-safe.
```
brianjlai, line 26:
```suggestion
        The DeclarativePartitionFactory takes a retriever_factory and not a retriever directly. The reason is that our components are not
        thread safe and classes like `DefaultPaginator` may not work because multiple threads can access and modify a shared field across each other.
        In order to avoid these problems, we will create one retriever per thread which should make the processing thread-safe.
```

Small nit, but i think it helps to illustrate an exact race condition for the shared paginator

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -344,3 +337,34 @@ def _remove_concurrent_streams_from_catalog(
                 if stream.stream.name not in concurrent_stream_names
             ]
         )
+
+    def _retriever_factory(
+        self, stream_config: ComponentDefinition, source_config: Config, stream_state: StreamState
+    ) -> Callable[[], Retriever]:
+        def _factory_method() -> Retriever:
+            declarative_stream: DeclarativeStream = self._constructor.create_component(
```
brianjlai, line 345: I feel pretty mixed about the fact that we have to create a runtime class of `DeclarativeStream` from the model instead of just instantiating a `SimpleRetriever` runtime instance. Given that I think in a previous convo, it seems reasonable that we want to deprecate `DeclarativeStream` in favor of `DefaultStream`. Similar to the existing code where we don't rely on the low-code runtime component of `DatetimeBasedCursor`, it still feels relevant here too

I presume the reason why we're opting for instantiating the the runtime stream object instead of the retriever directly because we need fields from the DeclarativeStream passed down into the retriever? If so let's just add a comment mentioning that.

brianjlai (approved):
looks good to me. I think the changes make sense and agree in moving towards the consolidation under the `StreamSlicer` in how we generate partitions

airbyte_cdk/sources/streams/concurrent/cursor.py
maxi297: The interfaces under the Concurrent CDK are often marked as experimental. I'll set a minor version and update source-salesforce right after

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
maxi297: > I presume the reason why we're opting for instantiating the the runtime stream object instead of the retriever directly because we need fields from the DeclarativeStream passed down into the retriever?

`create_declarative_stream` has some logic to generate the retriever. When the DeclarativeStream will be deleted, this logic will not be deleted but will be used to instantiate a `DefaultStream` with a `StreamSlicerPartitionGenerator`. I haven't put too much time in doing that because today the `client_side_incremental_sync` and `cursor_used_for_stop_condition` depends on the previous interfaces which is blocked by https://github.com/airbytehq/airbyte-internal-issues/issues/10568

airbyte_cdk/utils/slice_hasher.py
```diff
@@ -0,0 +1,22 @@
...
+
+class SliceHasher:
+    @classmethod
+    def hash(cls, stream_name: str, stream_slice: Optional[Mapping[str, Any]] = None) -> int:
+        if stream_slice:
+            # Convert the slice to a string so that it can be hashed
+            s = json.dumps(stream_slice, sort_keys=True, cls=SliceEncoder)
+            return hash((stream_name, s))
+        else:
+            return hash(stream_name)
```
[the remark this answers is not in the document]
maxi297: Added this suggestion

PR #50, opened 2024-11-14
feat: add download_decoder + download_extractor

artem1205:
# What

resolving https://github.com/airbytehq/airbyte-internal-issues/issues/10429

## Summary by CodeRabbit

- **New Features**
  - Introduced `download_extractor` and `download_decoder` properties in the asynchronous data retrieval process, enhancing flexibility for record extraction and decoding.
  - Added a new `ResponseToFileExtractor` class for handling responses that require extraction and decompression.
  - Implemented a custom `CustomRecordExtractor` class for improved record extraction from JSON responses.
  - Enhanced testing capabilities with a fixture for handling large datasets in the `ResponseToFileExtractor` tests.

- **Bug Fixes**
  - Improved error handling for the `download_extractor` instantiation to ensure default functionality when not explicitly defined.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2024,16 +2023,34 @@ def create_async_retriever(
...
-            decoder=decoder,
+            decoder=download_decoder,
             config=config,
             name=job_download_components_name,
         )
         download_retriever = SimpleRetriever(
             requester=download_requester,
             record_selector=RecordSelector(
-                extractor=ResponseToFileExtractor(),
+                # extractor=ResponseToFileExtractor(),# old one
```
maxi297, line 2052: Can we remove this line?

maxi297 (commented):
Should we add a test to validate that custom decoders are supported?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
artem1205: sure, do we need to add `ResponseToFileExtractor` to the list of possible choices `Decoders` ?

maxi297: Do you mean "to the list of possible `extractors`"? I think that would make sense. I think we were reluctant at first because it was considered experimental since it wasn't in the Connector Builder but if we need it, I'm good with that

airbyte_cdk/sources/declarative/extractors/response_to_file_extractor.py
```diff
@@ -89,21 +94,18 @@ def _save_to_file(self, response: requests.Response) -> Tuple[str, str]:
         """
         # set filepath for binary data from response
         decompressor = zlib.decompressobj(zlib.MAX_WBITS | 32)
-        needs_decompression = True  # we will assume at first that the response is compressed and change the flag if not
```
maxi297, line 92: As mentioned by the comment, we were relying on the response to know if we need to decompress. The logic of determining compression on the fly was implemented for sendgrid because of [this](https://github.com/airbytehq/airbyte/pull/35343). Is there a reason why we want to pass that at the initialization of the object? Have we validate that there won't be impact of sendgrid?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1331,6 +1331,23 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
+  ResponseToFileExtractor:
+    title: CSV To File Extractor
+    description: Record extractor that downloads a headless
```
maxi297, line 1336: What is `headless`? My understanding of this component is that we need to use it when the responses could be too big in memory and we need to save them on disk to avoid OOM issues. Can we add something in the description to reflect that?

maxi297 (commented):
My main concern is regarding backward compatibility with source-sendgrid. Once we figure this out, I'll approve

airbyte_cdk/sources/declarative/extractors/response_to_file_extractor.py
artem1205: my bad, i'll revert this

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
artem1205: description updated

airbyte_cdk/sources/declarative/extractors/response_to_file_extractor.py
artem1205: reverted

maxi297 (approved):
LGTM! Thanks Artem. I'm eager to see this applied to amazon-ads

PR #62, opened 2024-11-18
feat: Remove low-code cursor usage

artem1205:
## What

Resolve https://github.com/airbytehq/airbyte-internal-issues/issues/10568

## How

- add `should_be_synced`

## Review guide

1. `airbyte_cdk/sources/streams/concurrent/cursor.py`

- [ ] YES 💚
- [ ] NO ❌

## Release Notes

- **New Features**
	- Enhanced cursor handling capabilities with new methods for determining record synchronization.
	- Added flexibility in pagination strategy by allowing different cursor types.
	- Improved record instantiation with additional parameters for better context.
	- Introduced a `stream_name` parameter in various components for enhanced traceability.

- **Bug Fixes**
	- Improved filtering logic to simplify record synchronization checks.
	- Enhanced error handling in various methods to enforce parameter requirements.

- **Tests**
	- Updated test cases to align with new filtering logic and cursor handling enhancements.
	- Refined assertions in tests to reflect changes in method parameters.
	- Added new test case for evaluating synchronization criteria based on cursor values.
	- Standardized test scenarios to utilize new `StreamSlice` representation.
	- Enhanced clarity in test code by using named parameters for `Record` instantiation.
	- Improved test coverage for `DatetimeBasedCursor` functionality across various scenarios.
	- Updated tests to reflect the new structure and usage of the `Record` class.
	- Added new tests for various components, including error handlers and authenticators, to ensure robust behavior under different configurations.

### DIFF
--- a/airbyte_cdk/sources/streams/concurrent/cursor.py
+++ b/airbyte_cdk/sources/streams/concurrent/cursor.py
@@ -178,6 +178,8 @@ class ConcurrentCursor(Cursor):
         self._most_recent_cursor_value_per_partition: MutableMapping[StreamSlice, Any] = {}
         self._has_closed_at_least_one_slice = False
         self._cursor_granularity = cursor_granularity
+        # Flag to track if the logger has been triggered (per stream)
+        self._should_be_synced_logger_triggered = False
 
     @property
     def state(self) -> MutableMapping[str, Any]:
@@ -457,8 +459,10 @@ class ConcurrentCursor(Cursor):
         try:
             record_cursor_value: CursorValueType = self._extract_cursor_value(record)  # type: ignore  # cursor_field is converted to an InterpolatedString in __post_init__
         except ValueError:
-            LOGGER.warning(
-                f"Could not find cursor field `{self.cursor_field.cursor_field_key}` in record. The incremental sync will assume it needs to be synced"
-            )
+            if not self._should_be_synced_logger_triggered:
+                LOGGER.warning(
+                    f"Could not find cursor field `{self.cursor_field.cursor_field_key}` in record. The incremental sync will assume it needs to be synced"
+                )
+                self._should_be_synced_logger_triggered = True
             return True
         return self.start <= record_cursor_value <= self._end_provider()


artem1205:
UPD:
pytest is green, but mypy still raises an error:
```
airbyte_cdk/sources/declarative/extractors/record_filter.py:80: error: Argument 1 to "should_be_synced" of "PerPartitionWithGlobalCursor" has incompatible type "Mapping[str, Any]"; expected "Record"  [arg-type]
airbyte_cdk/sources/declarative/extractors/record_filter.py:80: error: Argument 1 to "should_be_synced" of "GlobalSubstreamCursor" has incompatible type "Mapping[str, Any]"; expected "Record"  [arg-type]
airbyte_cdk/sources/declarative/extractors/record_filter.py:80: error: Argument 1 to "should_be_synced" of "DatetimeBasedCursor" has incompatible type "Mapping[str, Any]"; expected "Record"  [arg-type]
airbyte_cdk/sources/streams/concurrent/cursor.py:456: error: Returning Any from function declared to return "bool"  [no-any-return]
airbyte_cdk/sources/declarative/requesters/paginators/strategies/stop_condition.py:37: error: Argument 1 to "should_be_synced" of "ConcurrentCursor" has incompatible type "airbyte_cdk.sources.types.Record"; expected "airbyte_cdk.sources.streams.concurrent.partitions.record.Record"  [arg-type]
```

I guess it is out of scope of current PR to fix this, as we'll have to change throughout all codebase 
- `records: Mapping[str, Any]` -> `records: Record`
- merge `airbyte_cdk.sources.streams.concurrent.partitions.record.Record` && `airbyte_cdk.sources.types.Record`

@maxi297 , WDYT ?

artem1205:
> Can we confirm this hypothesis? We would need a client-side filtering stream with a parent stream for that.

Tested on mailchimp segment_members stream::: got an exception ::

So I see 2 options  here:

1. Refactor `RecordSelector` and all related components to make them work with class Record: https://github.com/airbytehq/airbyte-python-cdk/blob/f62c91eccfa658f529b416031ba3cc5ece93c9c6/airbyte_cdk/sources/declarative/extractors/record_selector.py#L84-L88


2. Use `patch` listed below to create `Record` dynamically during the filtering
```diff
Index: airbyte_cdk/sources/declarative/extractors/record_filter.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/airbyte_cdk/sources/declarative/extractors/record_filter.py b/airbyte_cdk/sources/declarative/extractors/record_filter.py
--- a/airbyte_cdk/sources/declarative/extractors/record_filter.py	(revision 6842cc40efe287250fb477a7438d6d76dab4cf19)
+++ b/airbyte_cdk/sources/declarative/extractors/record_filter.py	(date 1732126900725)
@@ -10,7 +10,7 @@
     PerPartitionWithGlobalCursor,
 )
 from airbyte_cdk.sources.declarative.interpolation.interpolated_boolean import InterpolatedBoolean
-from airbyte_cdk.sources.types import Config, StreamSlice, StreamState
+from airbyte_cdk.sources.types import Config, StreamSlice, StreamState, Record
 
 
 @dataclass
@@ -77,7 +77,7 @@
         records = (
             record
             for record in records
-            if (self._substream_cursor or self._date_time_based_cursor).should_be_synced(record)
+            if (self._substream_cursor or self._date_time_based_cursor).should_be_synced(Record(data=record, associated_slice=stream_slice))
         )
         if self.condition:
             records = super().filter_records(

```


Do we plan to initialize `Record`  as early as possible in our Selector?

maxi297:
I think I don't understand the proposals. It feels like in both case, we would be at risk of having a cursor with a different implementation that doesn't take the type of record. For example, if the cursor is the ConcurrentCursor, calling `.should_be_synced(Record(data=record, associated_slice=stream_slice))` as part of `record_filter.py` would probably break. Let me know if I'm missing something.

So for me right now, the only solutions are:
* Break the interface in the declarative package to set types that both declarative and concurrent know (something like `def should_be_synced(record_data: Mapping, stream_slice: Mapping)`). I don't like that because we are breaking the interface which will be complicated to manage
* Updating airbyte_cdk.sources.types.Record to an interface that does not depend on the `partition` but would fit the needs of the concurrent package. This involves:
  * Updating `record.partition` usages in `ConcurrentCursor` to rely on the slice, not the partition
  * Updating `record.partition` usages in `ConcurrentReadProcessor` so that the record has the stream name (we will need to add the stream name to the record)
  * Updating the declarative package to pass the stream name to `record` (I did not see any usage of `Record.__init__` in the sources but I would like a double check on that
  * As the concurrent package should not know about the declarative one, we should probably move the record to a different package but find a way to keep the import from 
I think I prefer the second one. It is more changes code wise in the CDK but create less issues in the sources. I'm also fine breaking interfaces in the concurrent framework as we agreed they were experimental until we complete the migration to one CDK. 

WDYT?

airbyte_cdk/sources/streams/concurrent/cursor.py
```diff
@@ -442,3 +445,12 @@ def _evaluate_upper_safely(self, lower: CursorValueType, step: GapType) -> Curso
             return lower + step
         except OverflowError:
             return self._end_provider()
+
+    def should_be_synced(self, record: Record) -> bool:
+        record_cursor_value:  CursorValueType = self._extract_cursor_value(record)  # type: ignore  # cursor_field is converted to an InterpolatedString in __post_init__
```
maxi297, line 450: I think `self._extract_cursor_value` will raise a ValueError if the record does not have a cursor value. Can we add a test for that to validate this? If needed, we should update this code

artem1205: ok, I'll add one. Do we need to handle this error (and proceed record as `True`), or raising an Exception is ok ?

maxi297: Proceeding the record as true seems like the less disruptive thing to do

airbyte_cdk/sources/declarative/extractors/record_filter.py
```diff
@@ -68,37 +67,19 @@ def __init__(
...
-        )
-        filter_date: datetime.datetime = self._get_filter_date(state_value)
         records = (
             record
             for record in records
-            if self._end_datetime
-            >= self._date_time_based_cursor.parse_date(record[self._cursor_field])
-            >= filter_date
+            if (self._substream_cursor or self._date_time_based_cursor).should_be_synced(
+                Record(data=record, associated_slice=stream_slice)
```
maxi297, line 81: Can we add a comment as to why it is fine to omit some of the parameters like `stream_name` here?

airbyte_cdk/sources/declarative/incremental/per_partition_cursor.py
```diff
@@ -325,8 +325,10 @@ def is_greater_than_or_equal(self, first: Record, second: Record) -> bool:
     @staticmethod
     def _convert_record_to_cursor_record(record: Record) -> Record:
         return Record(
```
maxi297, line 327: Small not to @tolik0 that this change might impact you (hopefully very slightly) for what you are working on

airbyte_cdk/sources/declarative/incremental/per_partition_with_global.py
```diff
@@ -194,9 +194,11 @@ def get_request_body_json(
         )
 
     def should_be_synced(self, record: Record) -> bool:
-        return self._global_cursor.should_be_synced(
-            record
-        ) or self._per_partition_cursor.should_be_synced(record)
+        return (
```
maxi297, line 197: Why change that to use the active cursor only? If we really need to use the active cursor, we should use `_get_active_cursor`

airbyte_cdk/sources/types.py
```diff
@@ -15,9 +16,17 @@
 
 
 class Record(Mapping[str, Any]):
-    def __init__(self, data: Mapping[str, Any], associated_slice: Optional[StreamSlice]):
+    def __init__(
+        self,
+        data: Mapping[str, Any],
+        associated_slice: Optional[StreamSlice] = None,
+        stream_name: Optional[str] = None,
```
maxi297, line 23: I think we should make this non optional and just on the filtering, we set a dummy value. I would like the typing to confirm that we pass this value wherever we can. WDYT?

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -477,9 +477,9 @@ def _extract_record(stream_data: StreamData, stream_slice: StreamSlice) -> Optio
             # Record is not part of `StreamData` but is the most common implementation of `Mapping[str, Any]` which is part of `StreamData`
             return stream_data
         elif isinstance(stream_data, (dict, Mapping)):
-            return Record(dict(stream_data), stream_slice)
+            return Record(data=dict(stream_data), associated_slice=stream_slice)
```
maxi297, line 480: Should we pass `stream_name` here?

airbyte_cdk/sources/declarative/extractors/record_filter.py
```diff
@@ -11,7 +10,7 @@
     PerPartitionWithGlobalCursor,
 )
 from airbyte_cdk.sources.declarative.interpolation.interpolated_boolean import InterpolatedBoolean
-from airbyte_cdk.sources.types import Config, StreamSlice, StreamState
+from airbyte_cdk.sources.types import Config, StreamSlice, StreamState, Record
```
maxi297, line 13: nit: is ruff validating order of import? It seems like it should be alphabetical here

airbyte_cdk/sources/streams/concurrent/adapters.py
```diff
@@ -294,7 +294,11 @@ def read(self) -> Iterable[Record]:
                     self._stream.transformer.transform(
                         data_to_return, self._stream.get_json_schema()
                     )
-                    yield Record(data_to_return, self)
+                    yield Record(
+                        data=data_to_return,
+                        stream_name=self.stream_name(),
+                        associated_slice=copy.deepcopy(self._slice),
```
maxi297, line 300: Why do we need to do a deepcopy here?

airbyte_cdk/sources/types.py
```diff
@@ -139,3 +148,6 @@ def __ne__(self, other: Any) -> bool:
 
     def __json_serializable__(self) -> Any:
         return self._stream_slice
+
+    def __hash__(self) -> int:
+        return hash(orjson.dumps(self._stream_slice))
```
maxi297, line 153: Is orjson ensuring order of fields? If so, what is the order? Note that we have `SliceEncoder` in `adapters.py` which seems to indicate that we have slices that would need more than just a `dumps` to work

maxi297 (commented):
Just a couple more comments. This is clean. I think we are hitting the last miles on this one. Thanks Artem for tackling this!

airbyte_cdk/sources/declarative/extractors/record_filter.py
artem1205: added

airbyte_cdk/sources/declarative/extractors/record_filter.py
artem1205: added sorting option in https://github.com/airbytehq/airbyte-python-cdk/pull/94

airbyte_cdk/sources/declarative/incremental/per_partition_with_global.py
artem1205:
```python
self._global_cursor.should_be_synced(
            record
        ) or self._per_partition_cursor.should_be_synced(record)
```

we might get an exception if `self._global_cursor...` returns `False`, we try to get it from `_per_partition_cursor` which cause an exception.

thanks! ref to use `_get_active_cursor`

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
artem1205: I guess yes, added

airbyte_cdk/sources/types.py
artem1205: hmm, i'll check it

airbyte_cdk/sources/streams/concurrent/cursor.py
```diff
@@ -442,3 +446,19 @@ def _evaluate_upper_safely(self, lower: CursorValueType, step: GapType) -> Curso
...
+        """
+        Determines if a record should be synced based on its cursor value.
+        :param record: The record to evaluate
+
+        :return: True if the record's cursor value falls within the sync boundaries
+        """
+        try:
+            record_cursor_value: CursorValueType = self._extract_cursor_value(record)  # type: ignore  # cursor_field is converted to an InterpolatedString in __post_init__
+        except ValueError:
+            LOGGER.warning(
```
maxi297, line 460: I was checking source-pipedrive for the migration to concurrency and this log might be very noisy. Can we only log once for that? Note that we will have the same issue in `observe` so we might want to add the try/except here too @darynaishchenko

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
artem1205: fixed in https://github.com/airbytehq/airbyte-python-cdk/pull/62/commits/7ab7e06df0a23ad5262cc73716f3413f498ff898;
mypy is much cleaner now

airbyte_cdk/sources/streams/concurrent/adapters.py
artem1205: no need, removed

airbyte_cdk/sources/types.py
artem1205: > Is orjson ensuring order of fields? If so, what is the order?

what do you mean by order? dict is already an ordered structure, we can pass `orjson.OPT_SORT_KEYS` to forcefully sort dict by dict_keys.

But I'm ok to use standart json library here, this is not a heavily used operation.

maxi297: Dict are ordered by when they were inserted. It means that dictionaries that are equal might not generate the same JSON representation. For example:
```
import json
x = {}
x[1] = 1
x[2] = 2
y = {}
y[2] = 2
y[1] = 1

print(f"JSON compare: {json.dumps(x) == json.dumps(y)}")
print(f"Python compare: {x == y}")
```

... will output:
```
JSON compare: False
Python compare: True
```

To fix the ordering problem, we would need to sort the keys.

airbyte_cdk/sources/streams/concurrent/cursor.py
maxi297: We can start once per stream for now and see if we need more eventually

airbyte_cdk/sources/declarative/incremental/per_partition_with_global.py
maxi297: What is the exception that is caused when calling with the per_partition_cursor? I would expect that both work fine

airbyte_cdk/sources/types.py
artem1205: converted to non optional

airbyte_cdk/sources/types.py
artem1205: gotcha, sort added

airbyte_cdk/sources/streams/concurrent/cursor.py
artem1205: added

airbyte_cdk/sources/declarative/incremental/per_partition_with_global.py
maxi297: Up on this. This is the only remaining question and we can close this after that!

artem1205: ok, so:
https://github.com/airbytehq/airbyte-python-cdk/blob/7ab7e06df0a23ad5262cc73716f3413f498ff898/unit_tests/sources/declarative/extractors/test_record_filter.py#L360-L364 (last 2 tests) failed due to:

``` python3
../airbyte_cdk/sources/declarative/incremental/per_partition_cursor.py:306: in should_be_synced
    return self._get_cursor(record).should_be_synced(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PerPartitionCursor(), record = {'id': 1, 'created_at': '2020-01-03'}

    def _get_cursor(self, record: Record) -> DeclarativeCursor:
        if not record.associated_slice:
            raise ValueError(
                "Invalid state as stream slices that are emitted should refer to an existing cursor"
            )
        partition_key = self._to_partition_key(record.associated_slice.partition)
        if partition_key not in self._cursor_per_partition:
>           raise ValueError(
                "Invalid state as stream slices that are emitted should refer to an existing cursor"
            )
E           ValueError: Invalid state as stream slices that are emitted should refer to an existing cursor
```

we should use active cursor only (do not check per-partition if global cursor returned `False`)

maxi297: Oh, that makes sense! It feels harsh to kill a sync because of this though. Should `should_be_synced` just return `True` if the partition does not exist? I'm curious what @tolik0 thinks about this one

maxi297 (approved):
LGTM! In terms of new version, can we confirm that the changes to `Record` will not be breaking for the sources? If not, I'm all good to merge this

artem1205:
Removed `Record` class is not imported/used directly so it should be safe to merge

PR #76, opened 2024-11-19
Christo/finalize cdk sdm publish flow

ChristoGrab:
## WAT WAT


.github/workflows/pypi_publish.yml
```diff
@@ -1,9 +1,12 @@
-name: Python Packaging
+name: Packaging and Publishing
 
 on:
   push:
-
   workflow_dispatch:
+    inputs:
+      version:
+        description: "The version to publish (e.g., v6.7.0dev4)"
```
aaronsteers, line 8: @ChristoGrab 

```suggestion
        description: "The version to publish (e.g., 6.7.0dev4)"
```

Since we are scraping "v" from the release tag, we may need to omit it here as well.

.github/workflows/pypi_publish.yml
```diff
@@ -16,32 +19,101 @@ jobs:
...
+
       - name: Attach Wheel to GitHub Release
         uses: svenstaro/upload-release-action@v2
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: dist/*.whl
-          tag: ${{ github.ref }}
+          tag: ${{ env.VERSION }}
           overwrite: true
           file_glob: true
```
aaronsteers, line 60: @ChristoGrab - This step will only work with release tag events.

You can add an "if" condition like `if: startsWith(github.ref, 'refs/tags/v')` to skip it for workflow dispatch.

The reason is that it's job is to attach the wheel to the release (not strictly necessary anyway). If using the `VERSION` env var, the "tag" format will probably need to be reconstructed: `refs/tags/v${VERSION}`.

ChristoGrab:
Closing in favour of a [fresh PR here](https://github.com/airbytehq/airbyte-python-cdk/pull/77)

PR #83, opened 2024-11-22
CI: move SDM image publish to dedicated step

aaronsteers:
The blast radius for SDM deploys is too large for us to constantly deploy. This moves the process to a single-step process that we can trigger.

How publishing would work with the workflow trigger:

1. If 'dry_run' is checked, all steps will be evaluated, except that nothing will be published to DockerHub.
1. You would need to run from a published version tag or define your own version tag.
2. If the version tag input is omitted and we aren't running from a release tag, then we'll just publish a SHA-tagged image. (Still useful for testing.)
3. If the version being release is _not_ a prerelease version (doesn't have 'rc' or 'beta' or 'dev' in the image name), then we'll also tag as "latest".

TODO:

- [ ] We have to basically merge this PR in order to be able to test it...
- [ ] Consider also kicking off the version bump PR to the Builder (`airbyte-platform-internal` repo) when this flow with run on a non-prelease version.
- [ ] Make sure that if the workflow is run with an explicit version tag that we also tell Poetry dynamic versioning about it so the built package has the correct version attached. (Not necessary when running from a release tag.)
- [ ] Consider removing the version ref input entirely and just requiring a prerelease tag to be created in the repo for pre-relases.

## Normal Releases

The preferred way to run this (for production releases) would be to select an already-deployed release tag as in this screenshot:

image

In that case, you only have to select the release tag and run the workflow from there.

To add:

- In the same flow, we can auto-create the downstream PR for the Builder to use this version of the CDK.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new GitHub Actions workflow for publishing the Source-Declarative-Manifest (SDM) connector as a Docker image.

- **Bug Fixes**
	- Enhanced error handling and version management in the Docker publishing workflow.

- **Chores**
	- Removed the DockerHub publishing job from the PyPI publishing workflow, streamlining the process to focus solely on Python package publishing.


aaronsteers:
Findings: 

- New `AIRBYTE_ENTRYPOINT` env var needs to be "source-declarative-manifest" without "poerty run". This is because Orchestrator is overiding working directory.
- Actual entrypoint should match for consistency during testing.

ChristoGrab (approved):
LGTM! Will follow up with the update to the Dockerfile

PR #89, opened 2024-11-26
feat(Low-Code Concurrent CDK): Allow non-incremental substreams and list based partition router streams with parents to be processed by the concurrent cdk

brianjlai:
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/10552

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of streams with improved logic for substreams and concurrent processing.
	- Introduced two new streams: `arcana_personas_stream` and `palace_enemies_stream`.
	- Added a new parameter to manage resumable full refresh behavior in stream components.
	- Updated `as_airbyte_stream` method to include an `is_resumable` attribute for better stream configuration.

- **Bug Fixes**
	- Improved error handling in the testing framework for specific slice keys.

- **Tests**
	- Updated test cases to accommodate new streams and refined assertions for concurrent streams.
	- Added checks for the `is_resumable` attribute in multiple test scenarios to ensure consistency. 

- **Documentation**
	- Added comments to clarify the rationale behind changes in resumable full refresh behavior.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1334,6 +1336,8 @@ def _merge_stream_slicers(
                 if model.incremental_sync
                 else None
             )
+        elif self._disable_resumable_full_refresh:
+            return stream_slicer
```
brianjlai, line 1340: This was an interesting trick that took me a moment to find.

Right now since for low-code we automatically instantiate an RFR cursor. This was done to automatically release RFR to low-code connectors. However, since we don't want to support RFR for concurrent, we need a flag to signal not to apply it. And returning here works because we either return the `stream_slicer` if it exists or `None`, skipping over auto-creating the RFR cursor.

There is a very specific gotcha that I had originally gotten tripped up on, where we lost the state of of the PerPartitionCursor while instantiating child cursor during partition generation. And when we actually synced the partition, all the child cursors were lost. @maxi297, just tagging you to remind me to mention this in more detail to you, because its guaranteed to come up when incremental substreams is going to be worked on. But right now, this issue doesn't affect substreams that are full refresh

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -193,6 +201,17 @@ def _group_streams(
...
 
+                partition_router_component_definition = (
+                    name_to_stream_mapping[declarative_stream.name]
+                    .get("retriever")
+                    .get("partition_router")
+                )
+
+                is_substream_without_incremental = (
+                    partition_router_component_definition
+                    and not datetime_based_cursor_component_definition
```
maxi297, line 212: nit: The naming makes it a big confusing. Should we rename `datetime_based_cursor_component_definition` for `incremental_sync` to clarify that it can be custom incremental syncs?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: That makes sense to me. I'm even wondering: is there a reason with we keep the following condition?
```
        elif stream_slicer:
            # For the Full-Refresh sub-streams, we use the nested `ChildPartitionResumableFullRefreshCursor`
            return PerPartitionCursor(
                cursor_factory=CursorFactory(
                    create_function=partial(ChildPartitionResumableFullRefreshCursor, {})
                ),
                partition_router=stream_slicer,
            )
```
My understanding is that only sources with that inherits directly from `ManifestDeclarativeSource` would keep the RFR behavior, is that correct? Once we roll out this change, it would only be the Connector Builder, right?

maxi297 (commented):
This looks good to me. There only the RFR part that I think I don't fully understand and would like more clarity before I approve.

brianjlai:
Regression test results for `source-asana`: https://github.com/airbytehq/airbyte/actions/runs/12041043907/job/33572113642

- 18 out of 19 streams are running in concurrent (one stream `workspaces` is full refresh and not a substream)
- Record counts for full refresh match exactly (asana does not have any incremental streams)
- There are mismatches on the `is_resumable = False` not present on stream objects in the catalog in the new version. This is ultimately not a problem because its optional and it not being there implies false. But I'll put a fix in this PR for consistency
- Records are reported as mismatched because of properties missing from a record. However, I spot checked these and these field values were null. This was a change added in CDK v4. And this is getting flagged since we're going from v1 -> v6.

I'll push the catalog fix and max's comments, but aside from that, asana doesn't seem to show any issues running on concurrent candidate version vs the legacy flow.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: hmmm that's a good point, i'd thought to leave it in just in case we still wanted to keep RFR in for manifest sources (connector builder sources, or any that inherit ManifestDeclarativeSource instead of ConcurrentSource) that weren't ported to concurrent yet, but still wanted to continue to have RFR.

If we want to be more aggressive and just drop it entirely moving from 6.2.0+ onward, I can drop the entire conditional block and `self_disable_resumable_full_refresh` field, but my thinking is we can leave it in for now and remove it easily later

maxi297: Let's leave it for now and plan for the removal of RFR in the declarative framework so that we do it right. I've added a row in the tech spec for now

maxi297 (approved):
LGTM! Thanks for the implementation and the validation on source-asana

PR #91, opened 2024-11-27
Feat: destination vector db add option for embedding without field names

frankiebromage:
### Description
Responding to issue [#48627](https://github.com/airbytehq/airbyte/issues/48627#issue-2684912914)

Including field names when embedding text changes the semantic meaning of a vector, so it is useful to have the option of embedding only the text without the field names.

This is my first contribution, so please let me know what else is needed!

### Changes
Adds an additional config to allow user to choose whether or not to embed field names with a vector db destination.

### DIFF
--- a/airbyte_cdk/destinations/vector_db_based/config.py
+++ b/airbyte_cdk/destinations/vector_db_based/config.py
@@ -108,6 +108,12 @@ class ProcessingConfigModel(BaseModel):
         always_show=True,
         examples=["text", "user.name", "users.*.name"],
     )
+    omit_field_names_from_embeddings: bool = Field(
+        default=False,
+        title="Omit field names from embeddings",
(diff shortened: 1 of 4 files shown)


aaronsteers:
@frankiebromage - Thanks for creating this! 

Just a heads up: one challenge with rolling this out will be that most of the vector store destinations are several versions behind in their CDK pin. This may be related to the reason the Pinecone test failed in CI.

natikgadzhi:
Kicking the tires on CI again

natikgadzhi:
@aaronsteers while it's low(er) priority, I think we can make an epic with all our vector destinations + openai and unstructured and langchain dependency upgrades and run a project with community devs to update them to the latest versions, and add integration tests (use docker to standup DB where needed). 

Gauging your appetite for something like this.

PR #92, opened 2024-11-27
fix: add request count with eviction for not retriable requests to fix memory problem

maxi297:
# what

Addresses high memory consumption when enabling concurrency in https://github.com/airbytehq/oncall/issues/6821.
        
The `_request_attempt_count` attribute keeps growing as multiple requests are made using the same `http_client`.
To mitigate this issue, we evict keys for completed requests once we confirm that no further retries are needed.
This helps manage memory usage more efficiently while maintaining the necessary logic for retry attempts.

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling and memory management in the HTTP client, improving request retry logic and clarity of error messages.
	- Introduced a mechanism to evict completed requests from memory based on retry conditions.

- **Bug Fixes**
	- Improved error message clarity and ensured sensitive information is filtered out during logging.

- **Tests**
	- Added new tests to validate request eviction logic and enhanced existing tests for better error handling and robustness.

- **Chores**
	- Added a new development dependency (`pympler`) for improved memory profiling.


airbyte_cdk/sources/streams/http/expiring_dictionary.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def remove_expired_keys(self):
+        """Remove expired keys from the dictionary."""
+        current_time = time.time()
+        expired_keys = []
+        for key, (value, timestamp) in self._store.items():
+            if current_time - timestamp >= self._expiration_time:
+                expired_keys.append(key)
+            else:
+                # After Python 3.7, items will be presented in the order they were added.
```
maxi297, line 17: Based on this, I would assume that it is expiring the entries based on the "created_at" and not the "updated_at" which seems to be an issue as one request could retry for a long time while the others might be successful and we don't need to retry them.

Two proposals:
* Keep the logic based on the time find a way to order this by updated time without too much computation power. If you'd ask me right now, I would think that this is not an easy problem to solve but I'm not opposed in trying to solve it
* In `HttpClient._handle_error_resolution`, if error_resolution.response_action not in self._ACTIONS_TO_RETRY_ON, then evict the entry from the cache where `self._ACTIONS_TO_RETRY_ON = {ResponseAction.RETRY, ResponseAction.RATE_LIMITED}`

aldogonzalez8: I think [this commit](https://github.com/airbytehq/airbyte-python-cdk/pull/92/commits/5a4a6235e2513d8c508f65a5dfad8e715dde5db0) resolves the "updated_at" problem (+ tests), taking a look on your second comment.

aldogonzalez8: > In HttpClient._handle_error_resolution, if error_resolution.response_action not in self._ACTIONS_TO_RETRY_ON, then evict the entry from the cache where self._ACTIONS_TO_RETRY_ON = {ResponseAction.RETRY, ResponseAction.RATE_LIMITED}

Ok, I got it. This is good as we would usually keep a light store. It makes most of my test useless, but I will remove it for now while I get a new one.

maxi297: I think your comment would indeed solve the issue. Thinking again about it, I think the second solution is more interesting because:
* It is less complex in the sense that we don't need to maintain a new data structure
* It is more efficient as we don't have to check the new data structure on each HTTP request

I think there is a drawback which is that it isn't so clear that we are at risk of a memory leak if we don't have an explicit data structure that ensures that (so this would be a DX issue) but if we can get a good test to ensure this, I think I would be sold.

On your side, do you have a preference?

aldogonzalez8: > It is less complex in the sense that we don't need to maintain a new data structure

This could be a problem for unexpected behavior/scenarios.

> but if we can get a good test to ensure this, I think I would be sold.

Let me try to achieve this through the second solution

aldogonzalez8: Ok, I updated to second propsal which is cleaner and simpler IMO, also added unit test that keeps an eye on count store size.

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -116,10 +116,11 @@ def __init__(
         else:
             self._backoff_strategies = [DefaultBackoffStrategy()]
         self._error_message_parser = error_message_parser or JsonErrorMessageParser()
+        self._request_attempt_count: Dict[requests.PreparedRequest, int] = {}
         self._disable_retries = disable_retries
         self._message_repository = message_repository
 
-    def __del__(self):
+    def __del__(self) -> None:
```
maxi297, line 123: I know this is not part of the code you added but:
* Have we confirmed that this was necessary? I would assume it has no impact in the end and therefore am wondering if we can just remove this logic
* The same question would apply to `response.close()` that was added. I fear this one is more prone to problems given that it assumes that the response is fully consumed before we close it

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -332,6 +340,10 @@ def _send(
 
         return response  # type: ignore # will either return a valid response of type requests.Response or raise an exception
 
+    def _evict_key(self, prepared_request: requests.PreparedRequest) -> None:
```
maxi297, line 343: Can we add a comment to explain what is the consequence of not cleaning this? It seems like it wasn't obvious since it took us a bit of time to investigate and find the issue

maxi297 (commented):
LGTM for this part. I'm mostly wondering on the other changes in [maxi297/test_closing_requests_objects](https://github.com/airbytehq/airbyte-python-cdk/tree/maxi297/test_closing_requests_objects) now

airbyte_cdk/sources/streams/http/http_client.py
aldogonzalez8: Yes, I added.

airbyte_cdk/sources/streams/http/http_client.py
aldogonzalez8: It's gone and still flat.

[Connection Inspection](https://app.datadoghq.com/dashboard/fmc-w6h-unw/connection-inspection?fromUser=true&refresh_mode=sliding&tpl_var_ab_connection_id%5B0%5D=adee3b22-dfac-4273-a30c-cf4ca9b7cfb6&from_ts=1732670155841&to_ts=1732756555841&live=true)
[Profiler](https://app.datadoghq.com/profiling/explorer?query=service%3A%22airbyte%2Fsource-klaviyo%22%20env%3Aprod&agg_m=%40prof_python_cpu_cores&agg_m_source=base&agg_q=service&agg_q_source=base&agg_t=avg&attribute=file&env=prod&fromUser=true&my_code=disabled&profile_type=heap-live-size&service=airbyte%2Fsource-klaviyo&top_n=10&top_o=top&version=2.11.0-dev.de8e73a64a&viz=flame_graph&x_missing=true&start=1732742151248&end=1732756551248&paused=true)

aldogonzalez8: For reference we pinned a ws for testing with this:

- 2.11.1-dev.183f86a108 from branch aldogonzalez8/source-klaviyo/evit-request-count that uses
- [v6.6.8rc12](https://github.com/airbytehq/airbyte-python-cdk/releases/tag/v6.6.8rc12) CDK pre dev from aldogonzalez8/airbyte-cdk/add_request_count_with_expiration

maxi297 (commented):
There seems to be a test failing but apart from that, the change looks good to me. Thanks for hopping onto this one Aldo! (note that you will need to auto approve as I was the one opening this PR)

Following the release of this, can we release source-klaviyo with an updated CDK version?

aldogonzalez8:
> There seems to be a test failing but apart from that, the change looks good to me. Thanks for hopping onto this one Aldo! (note that you will need to auto approve as I was the one opening this PR)
> 
> Following the release of this, can we release source-klaviyo with an updated CDK version?

Auto approve is my passion.

Yes, I will look at the test and later follow up with a Klaviyo release.

aldogonzalez8:
I will release it on 2024-12-02.

PR #96, opened 2024-12-02
feat(concurrent): [ISSUE #10550] do not raise when record does not have cursor value in…

maxi297:
… concurrent cursor

## What

Following https://github.com/airbytehq/airbyte-internal-issues/issues/10550 change on source-pipedrive, we have seen issues when the cursor field was not provided. 

For now, we will assume that we need to sync those records but they won't influence the `most_recent_cursor_value`

## How
By not raising if the cursor value is not provided

## Summary by CodeRabbit

- **New Features**
	- Improved error handling and logging for cursor value extraction in the `ConcurrentCursor` class.

- **Bug Fixes**
	- Enhanced stability by ensuring that the system does not raise exceptions when no cursor value is present.

- **Tests**
	- Added a new test to verify that the `observe` method behaves correctly when no cursor value is provided, along with updates to existing tests for comprehensive coverage.

### DIFF
--- a/airbyte_cdk/sources/streams/concurrent/cursor.py
+++ b/airbyte_cdk/sources/streams/concurrent/cursor.py
@@ -186,7 +186,9 @@ class ConcurrentCursor(Cursor):
         self.start, self._concurrent_state = self._get_concurrent_state(stream_state)
         self._lookback_window = lookback_window
         self._slice_range = slice_range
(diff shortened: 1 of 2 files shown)


darynaishchenko (approved):
LGTM
Tested on source-pipedrive, records without cursor field were synced successfully

Issue #64, opened 2024-11-18
Ability to test manifest-only connectors directly from the CDK repo (via `--use-local-cdk` or similar)


aaronsteers:
In absence of a `--use-local-cdk` option for manifest-only connectors, we can't test them against dev branches in this repo.

We'll either need to update `airbyte-ci` to accept `--use-local-cdk` for manifest-only connectors, or we would need some other test flow to confirm downstream connectors are not breaking (or conversely to confirm their bugs are fixed) by changes here in this repo.

aaronsteers:
@alafanechere, @natikgadzhi, @ChristoGrab - Logging for visibility.

My best suggestion is that we add something like`--use-local-cdk` for manifest-only connectors, but another option would be adding `--use-cdk-ref` support. 

1. `--use-local-cdk` - In this former case, we'd run a manifest-only connector from an adjecently-cloned CDK repo. 
2. `--use-cdk-ref` - In this case, we'd point the test to a specific commit.

My rationale for preferring "option 1": while option 1 can subsume the requirements of "option 2" simply by cloning both and switching to the correct branches (we're already doing this for Python connectors), option 2 does not work for not-yet-committed changes. Which means: a developer making changes to the CDK could not test their changes are successful until/unless they commit and push their changes.

Maybe there is an "option 3" I'm not aware of (@alafanechere to confirm?), but I think we need to prioritize this highly (24-48 hours) even if a perfect options isn't yet available.

@alafanechere - does option 1 sound sensible to you as well? And if so, do you have any capacity to help us deliver this? I think this is in the "asap" category of post-migration action items.

PR #99, opened 2024-12-02
ci: run airbyte connector tests from master branch

aaronsteers:
This reverts a previous change that pulled airbyte-ci and connector images from a fixed branch where we were iterating on airbyte-ci improvements.

After this change:
- All connector definitions in the "Test Connectors" workflow will be pulled from `master` branch on the monorepo.
- Manifest-only connectors (PokeAPI and The Guardian) are removed from the matrix since they otherwise would represent a false positive success. The work to enable them again is logged here:
  - https://github.com/airbytehq/airbyte-python-cdk/issues/64

## Summary by CodeRabbit

- **Chores**
	- Updated workflow configuration for testing connectors to enhance clarity and functionality.
	- Improved comments regarding filters and job cancellation logic.
	- Refined connector references for better readability without altering functionality.

### DIFF
--- a/.github/workflows/connector-tests.yml
+++ b/.github/workflows/connector-tests.yml
@@ -74,6 +74,8 @@ jobs:
             cdk_extra: n/a
           - connector: source-chargebee
             cdk_extra: n/a
+          - connector: source-zendesk-support
+            cdk_extra: n/a
           - connector: source-s3
             cdk_extra: file-based
           - connector: destination-pinecone
--- a/.github/workflows/connector-tests.yml
+++ b/.github/workflows/connector-tests.yml
@@ -74,16 +74,18 @@ jobs:
             cdk_extra: n/a
           - connector: source-chargebee
             cdk_extra: n/a
-          - connector: source-zendesk-support
-            cdk_extra: n/a
           - connector: source-s3
             cdk_extra: file-based
           - connector: destination-pinecone
             cdk_extra: vector-db-based
           - connector: destination-motherduck
             cdk_extra: sql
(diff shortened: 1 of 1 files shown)


.github/workflows/connector-tests.yml
```diff
@@ -74,21 +74,18 @@ jobs:
             cdk_extra: n/a
           - connector: source-chargebee
             cdk_extra: n/a
-          # Currently not passing CI (unrelated)
```
maxi297, line 77: This should be fixed now. Why are we removing source-zendesk-support?

maxi297 (commented):
LGTM! One question regarding zendesk-support

.github/workflows/connector-tests.yml
aaronsteers: @maxi297 - Happy to restore it. I'll do that now.

aaronsteers: Done ✅

maxi297: It seems like `ticket_activities` is missing some records. I don't know if they expire. @DanyloGL would you be able to check this one?

.github/workflows/connector-tests.yml
```diff
@@ -116,8 +117,7 @@ jobs:
         if: steps.no_changes.outputs.status != 'cancelled'
         with:
           repository: airbytehq/airbyte
-          # TODO: Revert to `master` after Airbyte CI released:
-          ref: aj/airbyte-ci/update-python-local-cdk-code
+          ref: master
```
aldogonzalez8, line 120: 👍

PR #111, opened 2024-12-04
feat(Low-Code Concurrent CDK): Add ConcurrentPerPartitionCursor

tolik0:
## Summary by CodeRabbit

- **New Features**
  - Added support for concurrent cursor processing in declarative sources.
  - Enhanced incremental sync capabilities with per-partition and global cursor strategies.
  - Introduced new methods for creating concurrent cursors based on partitioned data streams.
  - Added a method to check if the timer is running in the Timer class.

- **Improvements**
  - Streamlined record filtering and cursor handling.
  - Improved error handling for cursor and record management.
  - Updated request option retrieval in retrievers.

- **Technical Updates**
  - Introduced new cursor classes for more flexible incremental synchronization.
  - Expanded test coverage for concurrent and incremental stream processing.
  - Updated tests to reflect changes in stream categorization and expected behaviors.
  - Added detailed unit tests for low-code connector streams and hierarchical substreams.
  - Enhanced the testing framework for `ConcurrentDeclarativeSource` and improved error handling in tests.


tolik0:
Regression tests: https://github.com/airbytehq/airbyte/actions/runs/12201672892/job/34040590334

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,273 @@
...
+        # The dict is ordered to ensure that once the maximum number of partitions is reached,
+        # the oldest partitions can be efficiently removed, maintaining the most recent partitions.
+        self._cursor_per_partition: OrderedDict[str, Cursor] = OrderedDict()
+        self._over_limit = 0
+        self._partition_serializer = PerPartitionKeySerializer()
+
+        self._set_initial_state(stream_state)
+
+    @property
+    def cursor_field(self) -> CursorField:
```
maxi297, line 86: Should we add this in the `Cursor` interface? If we put it there, it seems like the return type will need to be `Optional` because of the `FinalStateCursor`

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,273 @@
...
+        return self._cursor_field
+
+    @property
+    def state(self) -> MutableMapping[str, Any]:
+        states = []
+        for partition_tuple, cursor in self._cursor_per_partition.items():
+            cursor_state = cursor._connector_state_converter.convert_to_state_message(
```
maxi297, line 93: I have two confusions regarding this method:

**Why is Cursor.state public?**
I know it was already like that before you changes but I'm trying to understand where it is used and with it is public. Before your changes, it didn't seem like we needed it to be public. 

**Is there a way we can access the information we need without accessing private parameters?** 
It feel like what we need to make public here is the representation of the data that is being set as an AirbyteMessage of type state. In order words, the one in `ConcurrentCursor._emit_state_message` because when PerPartitionCursor will want to emit the state, it does not care about the implementation of the cursor but care about setting it in a state message

Based on the two points above, can we have the `state` method be more explicit (something like `as_state_message`) with a comment saying it is used for PerPartitionCursor and it is NOT used for other classes that are not cursors to emit states?

airbyte_cdk/sources/streams/concurrent/cursor.py
```diff
@@ -240,6 +240,15 @@ def observe(self, record: Record) -> None:
     def _extract_cursor_value(self, record: Record) -> Any:
         return self._connector_state_converter.parse_value(self._cursor_field.extract_value(record))
 
+    def close_partition_without_emit(self, partition: Partition) -> None:
```
maxi297, line 243: It feels like if we add a public method here, it should be also in the public interface. Based on that, I'm wondering if it would be easier to have the `ConcurrentCursorFactory` set the `message_repository` as a `NoopMessageRepository` to prevent the cursors from emitting anything. This solution would work with any type of cursors as well (let's say we want to support custom cursors soon)

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,270 @@
...
+            if cursor_state:
+                states.append(
+                    {
+                        "partition": self._to_dict(partition_tuple),
+                        "cursor": copy.deepcopy(cursor_state),
+                    }
+                )
+        state: dict[str, Any] = {"states": states}
+        return state
+
```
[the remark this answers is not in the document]
maxi297: Is there a reason we're doing a deepcopy here? If so, I think we should document it

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -178,7 +178,7 @@ def _request_headers(
             stream_slice,
             next_page_token,
             self._paginator.get_request_headers,
-            self.stream_slicer.get_request_headers,
+            self.request_option_provider.get_request_headers,
```
maxi297, line 181: @brianjlai was there a reason we were still using the `stream_slicer` here instead of the `request_option_provider`?

maxi297 (commented):
I'm very happy with the progress on this! I've added a couple of comments more in terms of code maintenance and structure but the functional part seemed fine. I want to check the tests eventually (tomorrow hopefully) but I can still leave a couple of comments here

airbyte_cdk/sources/streams/concurrent/cursor.py
tolik0: Fixed

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Fixed

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,273 @@
...
+)
+from airbyte_cdk.sources.streams.concurrent.cursor import Cursor, CursorField
+from airbyte_cdk.sources.streams.concurrent.partitions.partition import Partition
+from airbyte_cdk.sources.types import Record, StreamSlice, StreamState
+
+logger = logging.getLogger("airbyte")
+
+
+class ConcurrentCursorFactory:
+    def __init__(self, create_function: Callable[..., Cursor]):
```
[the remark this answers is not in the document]
tolik0: These are keyword arguments, so we can’t directly enforce types in the Callable signature.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Why do FinalStateCursor need this method?

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -304,6 +307,74 @@ def _group_streams(
...
+                    == DatetimeBasedCursorModel.__name__
+                    and self._stream_supports_concurrent_partition_processing(
+                        declarative_stream=declarative_stream
+                    )
+                    and hasattr(declarative_stream.retriever, "stream_slicer")
+                    and isinstance(
+                        declarative_stream.retriever.stream_slicer, PerPartitionWithGlobalCursor
+                    )
+                ):
+                    stream_state = state_manager.get_stream_state(
```
maxi297, line 322: The inside of the `elif` clause seems very similar to the one for `if self._is_datetime_incremental_without_partition_routing(declarative_stream, incremental_sync_component_definition)`. Should we merge them?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,316 @@
...
+)
+from airbyte_cdk.sources.streams.concurrent.cursor import ConcurrentCursor, Cursor, CursorField
+from airbyte_cdk.sources.streams.concurrent.partitions.partition import Partition
+from airbyte_cdk.sources.types import Record, StreamSlice, StreamState
+
+logger = logging.getLogger("airbyte")
+
+
+class ConcurrentCursorFactory:
+    def __init__(self, create_function: Callable[..., ConcurrentCursor]):
```
maxi297, line 31: out of scope (sharing so that we are on the same page): I fear the `...` is a bad smell. The stream state parameter makes sense to me because it should make sense to all the cursors. I think it goes back to what we mentioned which is that right now, the per-partition is directly coupled with datetime and when we will have other types of cursor we will need to revisit that.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
maxi297: Good point: this indeed seems weird. My question should probably have been: why do we expose that? I only see this used internally.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,316 @@
...
+        Set the initial state for the cursors.
+
+        This method initializes the state for each partition cursor using the provided stream state.
+        If a partition state is provided in the stream state, it will update the corresponding partition cursor with this state.
+
+        Additionally, it sets the parent state for partition routers that are based on parent streams. If a partition router
+        does not have parent streams, this step will be skipped due to the default PartitionRouter implementation.
+
+        Args:
+            stream_state (StreamState): The state of the streams to be set. The format of the stream state should be:
```
maxi297, line 228: Can we add the global state/lookback window parts to the doc as well?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,316 @@
...
+        )
+        self._message_repository.emit_message(state_message)
+
+    def stream_slices(self) -> Iterable[StreamSlice]:
+        slices = self._partition_router.stream_slices()
+        self._timer.start()
+        for partition in slices:
+            yield from self.generate_slices_from_partition(partition)
+
+    def generate_slices_from_partition(self, partition: StreamSlice) -> Iterable[StreamSlice]:
```
maxi297, line 177: nit: can we prefix with `_` to make it clear it is private?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,316 @@
...
+            self.state,
+        )
+        state_message = self._connector_state_manager.create_state_message(
+            self._stream_name, self._stream_namespace
+        )
+        self._message_repository.emit_message(state_message)
+
+    def stream_slices(self) -> Iterable[StreamSlice]:
+        slices = self._partition_router.stream_slices()
+        self._timer.start()
```
maxi297, line 173: nit: It's probably fine because we control the CDK and we don't expect people to call `stream_slices` but it would be weird if `stream_slices` was called more than once. Should we at least add a comment mentioning that the assumption is that it's once called once per sync?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -0,0 +1,316 @@
...
+                states.append(
+                    {
+                        "partition": self._to_dict(partition_tuple),
+                        "cursor": copy.deepcopy(cursor.state),
+                    }
+                )
+        state: dict[str, Any] = {"states": states}
+
+        if self._global_cursor:
+            state["state"] = self._global_cursor
```
maxi297, line 116: Would it make sense to have this as `global_state` instead of just `state`? Is this a backward compatibility thing? I'm wondering because just `state` is not so explicit given that there are `per_partition_state`, `global_state`, `parent_state`, etc...

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -0,0 +1,2215 @@
...
+
+def _run_read(
+    manifest: Mapping[str, Any],
+    config: Mapping[str, Any],
+    stream_name: str,
+    state: Optional[Union[List[AirbyteStateMessage], MutableMapping[str, Any]]] = None,
+) -> List[AirbyteMessage]:
+    catalog = ConfiguredAirbyteCatalog(
+        streams=[
+            ConfiguredAirbyteStream(
```
maxi297, line 262: nit: Using `ConfiguredAirbyteStreamBuilder` will make this more maintainable as if the format changes, we have a layer of protection for this not to break

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -0,0 +1,2215 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
```
maxi297, line 1: out-of-scope (just a comment to see if we can do differently next time): I like how exhaustive the tests are. I also feel they might be a bit hard to maintain for someone that does not have prior knowledge of this code. I'm wondering if some of those things might have helps:
* Re-using read and message parsing logic by using the EntrypointWrapper (it feels like all the orjson parsing does not need to be visible from the test's perspective)
* Using variables to improve naming and make dependencies explicit. For example, it feels easier to relate to other parts of the test if I see this:
```
                        "comments": [
                            {"id": 9, "post_id": 1, "updated_at": "2023-01-01T00:00:00Z"},
                            {"id": 10, "post_id": 1, "updated_at": _MOST_RECENT_DATE_OF_PARTITION_X},
                            {"id": 11, "post_id": 1, "updated_at": "2024-01-24T00:00:00Z"},
                        ],
```
I can assume that the dates for id 9 and id 11 are not so important and I can understand why the date for id 10 is important.

I'm wondering if there is anything you would add to this list. Again, I'm not asking to change the tests but I feel it is crucial to eventually improve them because we have a good idea of what's going on here but for someone new to cursors, they will freak out.

maxi297 (commented):
I don't think there is anything blocking but I added a couple of comments. I'll re-review whenever you want and approve once most of these are resolved

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: The arguments are explicitly stated in the `create` method:
```
def create(
    self, stream_state: Mapping[str, Any], runtime_lookback_window: Any
) -> ConcurrentCursor:
```
Additionally, this class is designed specifically to work with `ConcurrentCursor`. We can revisit and refactor the implementation if and when support for other cursor types is added.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Yes, this follows the same format used in declarative streams. Would you like me to create an issue to refactor it later for better clarity?

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
tolik0: I agree, these tests are quite complex, and something could easily be missed during updates. I’ll add documentation to explain what is being filtered and why the arguments are structured as they are. Additionally, I’ll incorporate variables to simplify the tests and improve readability.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
maxi297: Ok, that makes sense. I think for now, we could just extract this in a constant called `_GLOBAL_STATE_KEY` with a comment as to why it is only `"state"` and it would be more than enough. Thanks for the context!

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Fixed

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
tolik0: Fixed

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: We are using `cursor_field` outside of this class:
```
                    concurrent_streams.append(
                        DefaultStream(
                            partition_generator=partition_generator,
                            name=declarative_stream.name,
                            json_schema=declarative_stream.get_json_schema(),
                            availability_strategy=AlwaysAvailableAvailabilityStrategy(),
                            primary_key=get_primary_key_from_stream(declarative_stream.primary_key),
                            cursor_field=perpartition_cursor.cursor_field.cursor_field_key,
                            logger=self.logger,
                            cursor=perpartition_cursor,
                        )
                    )
```
https://github.com/airbytehq/airbyte-python-cdk/blob/cf5107f3d75db401dc1512a32be485558783684d/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L255

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Fixed

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: Updated

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
tolik0: Added helper method

tolik0:
Regression tests for Jira successful: [GitHub Action](https://github.com/airbytehq/airbyte/actions/runs/12826900382/job/35767780432)

Failure: `test_all_records_are_the_same_with_state`

Issues:

1. `project_avatars`: `projectId` is added by a transformation, so `id` does not uniquely identify the record.
2. `screen_tab_fields`: Similar to `project_avatars`, `screenId` and `tabId` are added by a transformation, so `id` does not uniquely identify the record.

maxi297 (approved):
When you say `It no longer uses resumable full refresh state, which resulted in an increased number of records.` for source-chargebee contact stream, it is when executed with a state value as an input, correct? 

If the above is true, I'm all good with this change. Thanks a lot for the due diligence on this one.

PR #130, opened 2024-12-05
fix(concurrent-cdk): Move the grouping of concurrent and synchronous streams into the read and discover commands instead of when initializing the source

brianjlai:
## Problem

We've seen a number of issues crop up due to the fact that in low-code sources, we instantiate the streams within the source's `__init__()` method. It also created issues for config migrations because we would instantiate a source's streams using an unmigrated config which might fail validations earlier than originally expected

## Solution

This PR moves the logic of grouping and creating streams back into the `read()` and `discover()` where errors will properly be surfaced where the used to be prior to moving to the concurrent CDK.

todo :

- [x] unit tests passing
- [x] retest locally on asana or other manifest-only source with an invalid config which would not properly surface check failed message

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of concurrent and synchronous streams in data processing.
	- Simplified state management in the `ConcurrentDeclarativeSource` class.

- **Bug Fixes**
	- Improved test coverage for JSON and JSONL decoders with larger datasets.

- **Tests**
	- Updated tests for `ConcurrentDeclarativeSource` to align with new stream grouping logic.
	- Increased line count in test responses to better simulate larger datasets.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -86,22 +86,9 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
             component_factory=component_factory,
         )
 
+        # todo: We could remove state from initialization. Now that streams are grouped during the read(), a source
+        #  no longer needs to store the original incoming state. But maybe there's an edge case?
         self._state = state
-
-        self._concurrent_streams: Optional[List[AbstractStream]]
-        self._synchronous_streams: Optional[List[Stream]]
-
-        # If the connector command was SPEC, there is no incoming config, and we cannot instantiate streams because
-        # they might depend on it. Ideally we want to have a static method on this class to get the spec without
-        # any other arguments, but the existing entrypoint.py isn't designed to support this. Just noting this
-        # for our future improvements to the CDK.
-        if config:
-            self._concurrent_streams, self._synchronous_streams = self._group_streams(
-                config=config or {}
-            )
-        else:
-            self._concurrent_streams = None
-            self._synchronous_streams = None
 
         concurrency_level_from_manifest = self._source_config.get("concurrency_level")
         if concurrency_level_from_manifest:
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -86,23 +86,10 @@ def __init__(
             component_factory=component_factory,
         )
 
+        # todo: We could remove state from initialization. Now that streams are grouped during the read(), a source
+        #  no longer needs to store the original incoming state. But maybe there's an edge case?
```
brianjlai, line 90: added this comment as a todo, i don't think we use state anywhere, but in essence of fewer changes to fix the bug.

unit_tests/sources/declarative/decoders/test_json_decoder.py
```diff
@@ -54,7 +54,7 @@ def test_jsonl_decoder(requests_mock, response_body, expected_json):
 def large_event_response_fixture():
     data = {"email": "email1@example.com"}
     jsonl_string = f"{json.dumps(data)}\n"
-    lines_in_response = 2  # ≈ 58 MB of response
+    lines_in_response = 2_000_000  # ≈ 58 MB of response
```
brianjlai, line 57: i might've accidentally changed this by accident in a past review

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -866,7 +867,21 @@ def _mock_party_members_skills_requests(http_mocker: HttpMocker) -> None:
...
+def mocked_init(self, is_sequential_state: bool = True):
+    """
+    This method is used to patch the existing __init__() function and always set is_sequential_state to
+    false. This is required because we want to test the concurrent state format. And because streams are
+    created under the hood of the read/discover/check command, we have no way of setting the field without
+    patching __init__()
+    """
+    self._is_sequential_state = False
+
+
```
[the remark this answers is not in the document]
brianjlai: sir this is a ~wendy's~ test already

unit_tests/sources/declarative/decoders/test_json_decoder.py
pnilan: Ha I remember that PR, too.

pnilan (approved):
Looks good!

Tested it with `source-sendgrid`. Resolved the issue.

brianjlai:
https://github.com/airbytehq/airbyte/actions/runs/12188006089/job/34000075439

regression test counts were successful and expected diffs verified.

PR #135, opened 2024-12-05
chore: fix typing issues

aaronsteers:
## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of stream configurations for improved clarity and error reduction.
	- Improved flexibility in incremental sync checks with nullable types.

- **Bug Fixes**
	- Refined error handling for concurrency levels and stream state management.

- **Documentation**
	- Updated method signatures for better understanding and usability.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -206,7 +206,7 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
             # so we need to treat them as synchronous
             if (
                 isinstance(declarative_stream, DeclarativeStream)
-                and name_to_stream_mapping[declarative_stream.name].get("retriever")["type"]
+                and name_to_stream_mapping[declarative_stream.name]["retriever"]["type"]
                 == "SimpleRetriever"
             ):
                 incremental_sync_component_definition = name_to_stream_mapping[
@@ -215,7 +215,7 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
 
                 partition_router_component_definition = (
                     name_to_stream_mapping[declarative_stream.name]
-                    .get("retriever")
+                    .get("retriever", {})
                     .get("partition_router")
                 )
                 is_without_partition_router_or_cursor = not bool(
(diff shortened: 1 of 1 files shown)


airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -206,7 +206,7 @@ def _group_streams(
             # so we need to treat them as synchronous
             if (
                 isinstance(declarative_stream, DeclarativeStream)
-                and name_to_stream_mapping[declarative_stream.name].get("retriever")["type"]
+                and name_to_stream_mapping[declarative_stream.name]["retriever"]["type"]
```
aaronsteers, line 209: Note: Previous code was unsafe because if "retriever" is not present during "get()", then "None" would not be subscriptable. This way, return type is "Any" which doesn't upset the typer - but of course the safer way is to check types before we infer that they are present - or just noqa it with the assumption we are making on the structure.

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -215,7 +215,7 @@ def _group_streams(
 
                 partition_router_component_definition = (
                     name_to_stream_mapping[declarative_stream.name]
-                    .get("retriever")
+                    .get("retriever", {})
                     .get("partition_router")
```
aaronsteers, line 219: Another case where .get() will return "None" if not present, which doesn't have "partition_router". My fix, allows it to fail gracefully. Alternatively, we could use the same method as above:

```suggestion
                    name_to_stream_mapping[declarative_stream.name]["retriever"]
                    .get("partition_router")
```

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -320,10 +320,11 @@ def _group_streams(
     def _is_datetime_incremental_without_partition_routing(
         self,
         declarative_stream: DeclarativeStream,
-        incremental_sync_component_definition: Mapping[str, Any],
+        incremental_sync_component_definition: Mapping[str, Any] | None,
     ) -> bool:
         return (
-            bool(incremental_sync_component_definition)
+            incremental_sync_component_definition is not None
+            and bool(incremental_sync_component_definition)
```
aaronsteers, line 327: This was being passed a possibly "None" value and it seemed simple enough to allow it to accept None, then just treat None explicitly falsey for the test condition.

@maxi297 - Does this look okay or should we just assert or noqa that it is not None in the caller?

maxi297 (approved):
LGTM!

PR #147, opened 2024-12-06
chore: make SDM image rootless

alafanechere:
## Summary by CodeRabbit

- **New Features**
  - Upgraded the base image to enhance performance and security.
  - Implemented a command to improve directory ownership for better security.

- **Bug Fixes**
  - Adjusted user permissions to ensure the application runs with least privilege.


alafanechere:
@ChristoGrab published a pre-release of the image and successfully tested `source-pokeapi` built from it.

Dockerfile
```diff
@@ -23,6 +23,10 @@ RUN mkdir -p source_declarative_manifest \
 # Remove unnecessary build files
 RUN rm -rf dist/ pyproject.toml poetry.lock README.md
 
+# Set ownership of /airbyte to the airbyte user
```
aaronsteers, line 26: @alafanechere - Is this a correct point, and if so, wdyt of including here for clarity?

```suggestion
# Set ownership of /airbyte to the (non-root) airbyte user
```

aaronsteers (approved):
Looks good! And grateful to you both (@alafanechere and @ChristoGrab) for the extra step of manually pre-testing in Cloud.

One inline comment for clarification and I think this is ready to go! 🚀

aaronsteers:
@alafanechere and @ChristoGrab - Will one of you be able to monitor after deploying? Specifically:

1. Check that Cloud (prod) has received the new version of the image?
2. Check that after new version is deployed, that execution is still running? 

(This would be helpful to add to our `RELEASES.md` file - I'm not 100% sure how to do it myself and I understand it differs slightly from that standard process of deploying/checking connectors post-release.)

aaronsteers:
Also - after this merges, we can release in the normal way - by publishing the draft release. It should automatically rebuild and republish with this Dockerfile definition. 🚀

aaronsteers (changes requested):
Sounds like manual testing is not 100% green as of now.

aaronsteers:
@alafanechere - I don't remember us overriding how the image is built during `--use-local-cdk`. I believe we only inject the python package into an existing SDM image (maybe "latest"?). So, in that case, I don't know if we can rely on CI to catch issues in the Dockerfile generation - or more specifically, with how our test connectors would interact with the different base image.

Can you confirm if that is your understanding as well?

alafanechere:
> @alafanechere - I don't remember us overriding how the image is built during `--use-local-cdk`. I believe we only inject the python package into an existing SDM image (maybe "latest"?). So, in that case, I don't know if we can rely on CI to catch issues in the Dockerfile generation - or more specifically, with how our test connectors would interact with the different base image.
> 
> Can you confirm if that is your understanding as well?

@aaronsteers your understanding is correct. The automated connector testing on CDK change in this repo does not use the "dev" image. Which image is used by a connector is defined in its metadata.yaml which we're not editing at test time via `--use-local-cdk`.
I manually tested `source-pokeapi` built from a pre-release of SDM from this branch.
I confirm the tests are passing and a basic Airbyte Cloud sync works with it.

aaronsteers (approved):
Code looks good. 👍

PR #149, opened 2024-12-06
feat(low-code cdk): add config component resolver

lazebnyi:
## What
This PR add config component resolver to dynamic stream
Fixed: https://github.com/airbytehq/airbyte-internal-issues/issues/10787?issue=airbytehq%7Cairbyte-internal-issues%7C10789

## How
Described schema, models and data classes for  `ConfigComponentsResolver` and`StreamConfig`. Defined behavior for components resolving using source config.  Added validation to ensure dynamic stream names are unique.

## Release Notes

- **New Features**
	- Introduced `StreamConfig` and `ConfigComponentsResolver` components for enhanced stream configuration management.
	- Updated `DynamicDeclarativeStream` to support new component resolution options.
	- Enhanced flexibility in component mapping with updated field path definitions.
	- Expanded component type mappings to include new configurations.

- **Bug Fixes**
	- Improved handling of unique dynamic stream names to prevent duplicates.

- **Tests**
	- Added unit tests for `ConfigComponentsResolver` to validate dynamic stream behavior.
	- Enhanced tests for `HttpComponentsResolver` with additional mock data and configuration usage.


airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1160,14 +1168,16 @@ class WaitUntilTimeFromHeader(BaseModel):
...
             ["data"],
             ["data", "records"],
-            ["data", "{{ parameters.name }}"],
+            ["data", 1, "name"],
+            ["data", "{{ components_values.name }}"],
             ["data", "*", "record"],
+            ["*", "**", "name"],
         ],
         title="Field Path",
     )
```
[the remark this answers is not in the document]
lazebnyi: this is autogenerated file

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1189,6 +1199,24 @@
...
+class StreamConfig(BaseModel):
+    type: Literal["StreamConfig"]
+    configs_pointer: List[str] = Field(
+        ...,
+        description="A list of potentially nested fields indicating the full path in source config file where streams configs located.",
+        examples=[["data"], ["data", "streams"], ["data", "{{ parameters.name }}"]],
+        title="Configs Pointer",
+    )
+    parameters: Optional[Dict[str, Any]] = Field(None, alias="$parameters")
+
```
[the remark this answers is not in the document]
lazebnyi: this is autogenerated file

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2974,6 +2976,52 @@ definitions:
       - type
       - retriever
       - components_mapping
+  StreamConfig:
```
maxi297, line 2979: The naming is mildly confusing to me. It does not seem to be a config per se but more like something to figure out where in the source config we can find the stream config. Is there a reason why we created a component that only has one field? I'm wondering if we could just move the `configs_pointer` under `ConfigComponentsResolver`

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2974,6 +2976,52 @@ definitions:
...
+        examples:
+          - ["data"]
+          - ["data", "streams"]
+          - ["data", "{{ parameters.name }}"]
+      $parameters:
+        type: object
+        additionalProperties: true
+  ConfigComponentsResolver:
+    type: object
+    description: (This component is experimental. Use at your own risk.) Component resolve and populates stream templates with components fetched from the source config.
```
maxi297, line 3007: nit: `(This component is experimental. Use at your own risk.) Resolves and populates stream templates with components fetched from the source config.`

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -350,7 +351,12 @@ def _dynamic_stream_configs(
                 if "type" not in dynamic_stream:
                     dynamic_stream["type"] = "DeclarativeStream"
 
-                dynamic_stream_configs.append(dynamic_stream)
+                # Ensure that each stream is created with a unique name
```
maxi297, line 354: Why could we have duplicates here? It feels like we would need more context if we keep those lines. I see from the test that an API could return twice the same information. Is this a case we've seen with a source? With my limited understanding of the problem, this is more like a config error than something we should handle by silently removing some streams.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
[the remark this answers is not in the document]
maxi297: Go home coderabbitai, you are drunk!

maxi297 (commented):
A couple of questions to make sure I have the full context but generally, it looks good to me!

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
lazebnyi: We can move the `configs_pointer` under `ConfigComponentsResolver`, but if any transformations or other specific details are needed for the config, we would have to add them at the `ConfigComponentsResolver` level, which isn't ideal. Having a separate entity, such as a `StreamConfig`, would make it easier to extend if needed.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
lazebnyi: updated

airbyte_cdk/sources/declarative/manifest_declarative_source.py
lazebnyi: I agree. This can help identify those issues as soon as possible. 
Updated

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -350,6 +353,19 @@ def _dynamic_stream_configs(
...
+                name = dynamic_stream.get("name")
+
+                if name in seen_dynamic_streams:
+                    error_message = f"Dynamic streams list contains a duplicate name: {name}. Please check your configuration."
+
+                    raise AirbyteTracedException(
+                        message=error_message,
+                        internal_message=error_message,
+                        failure_type=FailureType.config_error,
```
maxi297, line 365: I'm wondering if we need to do the distinction between dynamic streams from config and dynamic streams from API. If this error comes for the config resolver, than it is a config error. If this error comes from the API, it's probably a system error. WDYT?

lazebnyi: It depends on how we think about the manifest. Is this a configuration for the source as well? If so, this could be treated as a configuration error because the problem lies in the endpoint defined in the manifest. If not, it should depend on the type of resolver.

I think it's okay to categorize it as a configuration error because, in my view, the description of the manifest involves configuring the source.

maxi297: I see it the other way. A config issue is an issue that the Airbyte user can fix by updating the config. An Airbyte user won't easily have access to update the description of the manifest so in this case, it should be flagged as a system error. Does that make sense?

lazebnyi: Ok. I have added condition for failure_type and unit test for it.

maxi297 (approved):
LGTM! Thanks for addressing those concerns

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -350,6 +353,24 @@ def _dynamic_stream_configs(
                 if "type" not in dynamic_stream:
                     dynamic_stream["type"] = "DeclarativeStream"
 
+                # Ensure that each stream is created with a unique name
+                name = dynamic_stream.get("name")
+
+                if name in seen_dynamic_streams:
+                    error_message = f"Dynamic streams list contains a duplicate name: {name}. Please contact Airbyte Support."
+                    failure_type = FailureType.system_error
```
aldogonzalez8, line 361: nit: Should this be a system error? As we are getting streams from an endpoint, this, in theory, would be an issue with such an endpoint. Unless we handled wrongly the info from such an endpoint, maybe?

lazebnyi: This is already `system_error` 😅

aldogonzalez8: Oh yes, that was my point, is not system error an indicator for "our" system? Im just trying to distinguish if this should be also some config error.
Sorry for the previous misleading comment.

aldogonzalez8 (approved):
Approved, changes make sense to me.

PR #152, opened 2024-12-06
fix: %s formatted state values were being converted to str in state blob

pnilan:
## What
- [This PR](https://github.com/airbytehq/airbyte/pull/48389) updated the default stream state converter to always be the `CustomFormatConcurrentStreamStateConverter`, which relies on the `DatetimeParser` for parsing and formatting dates. Inadvertently the returned `output_format` of cursor values for %s timestamps was changed from `int` to `str`.
- This PR updates the `DatetimeParser` to return int, float, or str for its `format` method.
- This also updated the `DatetimeBasedCursor` to convert the formatted datetime to a string to maintain compatibility
- Update the DateTimeParser to output the same type as the input state value.
    - %s -> int
    - %s_as_float -> float
    - %ms -> int
    - else -> str

## Release Notes

- **New Features**
	- Enhanced datetime handling with methods now returning string, integer, or float representations for timestamps.
	- Introduced a new method to dynamically adjust the lookback window based on runtime conditions.

- **Bug Fixes**
	- Updated test cases to align with the new output types for datetime formatting, ensuring accurate validation.

- **Tests**
	- Improved test coverage for datetime parsing to accommodate both string and integer inputs.

### DIFF
--- a/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
+++ b/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
@@ -37,12 +37,12 @@ class DatetimeParser:
             return parsed_datetime.replace(tzinfo=datetime.timezone.utc)
         return parsed_datetime
 
-    def format(self, dt: datetime.datetime, format: str) -> str:
+    def format(self, dt: datetime.datetime, format: str) -> Union[str, int]:
         # strftime("%s") is unreliable because it ignores the time zone information and assumes the time zone of the system it's running on
         # It's safer to use the timestamp() method than the %s directive
         # See https://stackoverflow.com/a/4974930
         if format == "%s":
-            return str(int(dt.timestamp()))
+            return int(dt.timestamp())
         if format == "%s_as_float":
             return str(float(dt.timestamp()))
         if format == "%ms":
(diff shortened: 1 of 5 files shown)


maxi297:
I think I've seen this as well but didn't see any negative impact from this being a string in the state blob. Is this an issue because we can't revert?

pnilan:
@maxi297 Main issue is failing CATs but we could also just update CATs or update abnormal state files  -- `test_state_with_abnormally_large_values` compares the returned state cursor value type against the type in the schema.

Practically speaking I don't think there is an issue with converting to and storing state as a string.

maxi297:
Ok! And is the datetime based cursor supporting the new format to ensure we can revert if needed?

pnilan:
@maxi297 Sorry, could you clarify what you're asking?

The intent of this PR was to maintain existing behavior. One concern I see is that given the SDM image has been updated with the change to default to `CustomFormatConcurrentStreamStateConverter`, it's possible that existing manifest-only connectors would have their state converted to string.

maxi297:
What is the downside of having state as strings? This is the part I'm not getting. The connectors are not breaking right? So the only downside I can see is the we would not be able to revert to a version that is using the DatetimeBasedCursor instead of the ConcurrentCursor. Can you confirm that this is the problem we are trying to solve right now?

pnilan:
Right, I agree there's isn't a downside of storing as strings -- my main concern was failing CATs, but I saw you had a PR a while ago where you disabled the test in question for `source-zendesk-support`, so if that is the norm then I can close this PR w/o merging.

Issue #164, opened 2024-12-10
feat(JsonParser) - Create new `JsonParser` component


pnilan:
## Problem
- Because the `ZipfileDecoder` will decompress data from the `requests.Response` object, it needs a component/class to parse the decompressed data that is not passed to it as a `requests.Response` object (like the `JsonDecoder` class).

## Proposed Solution
- Implement a `Parser` interface that has one primary method: `parse`
- Implement a `JsonParser` class

natikgadzhi:
@pnilan How does this change with @artem1205's pull requests? What needs to be rewroked here?

PR #166, opened 2024-12-11
feat: Add `JsonParser` component to declarative framework

pnilan:
## What
- Adds new `JsonParser` component which inherits from the `Parser` interface.
- Resolves https://github.com/airbytehq/airbyte-python-cdk/issues/164

## Recommended Reviewing Order:
1. declarative_component_schema.yaml
2. composite_raw_decoder.py
3. model_to_component_factory.py
4. test_composite_decoder.py

## Release Notes

- **New Features**
  - Added JSON parsing capabilities to the declarative components framework.
  - Introduced `JsonParser` for handling JSON data with flexible encoding support.

- **Improvements**
  - Enhanced data handling with new parsing methods, including support for various data types and error handling.
  - Improved readability and consistency of documentation.

- **Testing**
  - Added comprehensive unit tests for new JSON parsing functionality.
  - Expanded test coverage for error handling in JSON parsing.


pnilan:
@coderabbitat

First off, you should ignore all `__init__.py` files when providing a recommended reviewing order. Additionally, for the context of this repo, you can always ignore including the `declarative_component_schema.py` file as this is auto-generated based on the `declarative_component_schema.yaml` file. Additionally, when providing a recommended reading order, it's fine to just include the filename and not the full path (unless there are multiple files with similar names).

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -42,6 +42,30 @@ def parse(
...
+    """
+    Parser strategy for converting JSON-structure str, bytes, or bytearray data into MutableMapping[str, Any].
+    """
+
+    encoding: str = "utf-8"
+
+    def parse(self, data: BufferedIOBase) -> Generator[MutableMapping[str, Any], None, None]:
+        raw_data = data.read()
+        try:
+            body_json = json.loads(raw_data.decode(self.encoding))
```
maxi297, line 56: Should we use orjson instead of json? If we do, we should make sure it is not affected by bugs like https://github.com/airbytehq/airbyte-python-cdk/pull/210

maxi297, line 56: Should we use orjson instead of json? If we do, we should make sure it is not affected by bugs like https://github.com/airbytehq/airbyte-python-cdk/pull/210

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -42,6 +42,30 @@ def parse(
...
+        except json.JSONDecodeError:
+            logger.warning(f"Data cannot be parsed into json: {data=}")
+            yield {}
```
maxi297, line 59: Shouldn't we raise a system_error in that case? Wouldn't it means that there is an unexpected behavior and we might lose records if we just skip?

maxi297 (commented):
My understanding is that from there, we will start parsing the json twice if there is pagination defined: once during dpath extraction and once during pagination. 

I don't think we need to fix now that but I'd like to:
* Make sure we use an efficient JSON parser (orjson)
* Poke our master performance on this ticket to make sure he knows @artem1205 

I think the longer term solution would be to have a `response` object so that we can keep the result of the parsing in-memory but it will not be easy to do since [our interfaces depends on the requests.Response](https://github.com/airbytehq/airbyte-python-cdk/blob/0e7802a4c485355d8d5abaa5f23b48b682dc4eb7/airbyte_cdk/sources/declarative/extractors/dpath_extractor.py#L71).

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
pnilan: Yes, I agree -- yielding an empty dictionary is existing behavior in the `JsonDecoder` and we should probably update that, too, in a separate PR.

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
pnilan: I'm inclined to maintain using the `json` library for now. See below for justification.

I played around with this for a bit and noticed that very large numbers (int128) are automatically converted to scientific notation float types. `json` library does not do this. No exception is raised when deserializing (unlike when serializing int128). I don't think we can assume we'll never receive int128 values. See screenshot for quick comparison.

Screenshot 2025-01-14 at 7 54 31 AM

artem1205: hmm, good point, let's leave it with json, just to be sure we do not break anything.

maxi297: My hypothesis is that this only happens for JSON that represents an integer. We always expect object, arrays or maybe in some weird case strings that represent text and not numbers. 

So:
* Move to orjson
* Catch exception for orjson with a clear comment that we don't know of any errors but want to be overly cautious given it's HTTP responses than could be anything and that we don't control
* Fallack on json
* Raise system_error else

unit_tests/sources/declarative/decoders/test_composite_decoder.py
```diff
@@ -117,3 +118,30 @@ def test_composite_raw_decoder_jsonline_parser(requests_mock, encoding: str):
...
+        "test_with_buffered_io_base_data_containing_int128",
+    ],
+)
+def test_json_parser_with_valid_data(data):
+    encodings = ["utf-8", "utf", "iso-8859-1"]
+
+    for encoding in encodings:
+        raw_data = json.dumps(data).encode(encoding)
+        for i, actual in enumerate(
+            JsonParser(encoding=encoding).parse(BufferedReader(BytesIO(raw_data)))
```
artem1205, line 146: nit: could you please wrap it in requests_mock and use CompositeRawDecoder.decode interface?

will approve 👍  after this

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -42,6 +45,31 @@ def parse(
...
+
+    def parse(self, data: BufferedIOBase) -> Generator[MutableMapping[str, Any], None, None]:
+        raw_data = data.read()
+        try:
+            body_json = orjson.loads(raw_data.decode(self.encoding))
+        except Exception:
+            try:
+                body_json = json.loads(raw_data.decode(self.encoding))
+            except Exception as exc:
+                raise AirbyteTracedException(
```
artem1205, line 60: LGTM overall, but I guess we could get rid of nested try/except block and use something like this:

```suggestion
    body_json = self._parse_orjson(raw_data) or self._parse_json(raw_data)
    if body_json is None:
            raise AirbyteTracedException(
            ....


def _parse_orjson(self, raw_data: bytes):
    try:
        return orjson.loads(raw_data.decode(self.encoding))
    except orjson.JSONDecodeError:
         logger.error(...)
         return None

def _parse_json(self, raw_data: bytes):
    try:
        return json.loads(raw_data.decode(self.encoding))
    except json.JSONDecodeError:
        logger.warning(...)
        return None
```

artem1205 (approved):
LGTM overall, approving to unblock. please take a look at my latest comment

docs/RELEASES.md
```diff
@@ -9,9 +9,9 @@ A few seconds after any PR is merged to `main` , a release draft will be created
 3. Optionally tweak the text in the release notes - for instance to call out contributors, to make a specific change more intuitive for readers to understand, or to move updates into a different category than they were assigned by default. (Note: You can also do this retroactively after publishing the release.)
 4. Publish the release by pressing the “Publish release” button.
 
-*Note:*
```
pnilan, line 12: Auto-updated via `poetry run poe fix-all`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2555,3 +2565,25 @@ def create_config_components_resolver(
             components_mapping=components_mapping,
             parameters=model.parameters or {},
         )
+
+    _UNSUPPORTED_DECODER_ERROR = (
```
natikgadzhi, line 2569: Having this is fine if it's used in several spots.

natikgadzhi (approved):
A couple nits, perhaps nuke the `allowAdditionalFields`? Otherwise looks good, shipit.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -72,6 +72,8 @@
     CsvParser,
     GzipParser,
     JsonLineParser,
+    JsonParser,
+    Parser,
```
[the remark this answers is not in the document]
pnilan: Used in _is_supported_parser_for_pagination method

Issue #168, opened 2024-12-12
[Concurrent Low-Code] Allow low-code streams using AsyncRetriever to be run within the Concurrent CDK


brianjlai:
## Context

After allowing certain types of low-code streams to be processed within the concurrent framework, we ran into an issue where stream that used the `AsyncRetriever` component would result in errors during processing. One such example is

```
{
  "type": "TRACE",
  "trace": {
    "type": "ERROR",
    "emitted_at": 1733873724427,
    "error": {
      "message": "Invalid state within AsyncJobRetriever. Please contact Airbyte Support",
      "internal_message": "AsyncPartitionRepository is expected to be accessed only after `stream_slices`",
      "stack_trace": "Traceback (most recent call last):\n  File \"/Users/brian.lai/dev/airbyte-python-cdk/airbyte_cdk/sources/streams/concurrent/partition_reader.py\", line 40, in process_partition\n    for record in partition.read():\n  File \"/Users/brian.lai/dev/airbyte-python-cdk/airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py\", line 59, in read\n    for stream_data in self._retriever.read_records(self._json_schema, self._stream_slice):\n  File \"/Users/brian.lai/dev/airbyte-python-cdk/airbyte_cdk/sources/declarative/retrievers/async_retriever.py\", line 119, in read_records\n    records: Iterable[Mapping[str, Any]] = self._job_orchestrator.fetch_records(partition)\n  File \"/Users/brian.lai/dev/airbyte-python-cdk/airbyte_cdk/sources/declarative/retrievers/async_retriever.py\", line 60, in _job_orchestrator\n    raise AirbyteTracedException(\nairbyte_cdk.utils.traced_exception.AirbyteTracedException: AsyncPartitionRepository is expected to be accessed only after `stream_slices`\n",
      "failure_type": "system_error",
      "stream_descriptor": {
        "name": "contacts"
      }
    }
  }
}
```

As a temporary fix, we changed the `concurrent_declarative_source.py` to only run streams using the `SimpleRetriever` concurrently.

Slack thread with more context:
https://airbytehq-team.slack.com/archives/C063B9A434H/p1733264441227669?thread_ts=1733257895.550789&cid=C063B9A434H

## Problem / Solution

The root of the issue is two fold

### Issue 1:

[...]

PR #170, opened 2024-12-13
fix(Low-Code Concurrent CDK): Refactor the low-code AsyncRetriever to use an underlying StreamSlicer

brianjlai:
## What is the issue

Something uncovered while using the latest version of the Concurrent CDK on `source-sendgrid` was that because it relied on the assumption all Retrievers had an underlying `stream_slicer` defined. This however was not the way the `AsyncRetriever` was implemented. 

## Implementation Details

To fix this, I've refactored the `AsyncRetriever` to follow a more established pattern exhibited by our existing `Retrievers`. Now the retriever will always have an underlying `stream_slicer` that can be invoked to generate the partitions within the concurrent framework. This allows us to continue to use the `StreamSlicerPartitionGenerator`.

This change was also designed to be non-breaking because we are not changing the developer facing interface. Instead we use the existing `AsyncRetriever` fields to construct the `AsyncJobPartitionGenerator` which is effectively an implementation detail.

Note:
This will not completely give us the ability to run streams using the `AsyncRetriever` in the concurrent framework. This just fixed the first issue identified in https://github.com/airbytehq/airbyte-python-cdk/issues/168.  I will work on a follow up PR that addresses the second issue. But I wanted to separate the PRs so I can release this refactor which should be a no-op for existing connectors like `source-sendgrid` and we should see no change in behavior. And because of that, I have not ungated the async report streams  `concurrent_declarative_source.py` since we need to address part 2.

todo: add unit tests to AsyncJobPartitionRouter

## Summary by CodeRabbit

- **New Features**
  - Introduced `AsyncJobPartitionRouter` for improved management of asynchronous job handling.
  - Added `AsyncRetriever` component for managing asynchronous job operations.

- **Bug Fixes**
  - Streamlined logic in `AsyncRetriever` to enhance performance and reduce complexity.

- **Tests**
  - Added tests for the new `AsyncRetriever` and `AsyncJobPartitionRouter` components to ensure proper functionality and integration.
  - Created unit tests for `AsyncJobPartitionRouter` to validate its behavior with different partitioning scenarios.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -308,6 +308,9 @@ from airbyte_cdk.sources.declarative.partition_routers import (
     SinglePartitionRouter,
     SubstreamPartitionRouter,
 )
+from airbyte_cdk.sources.declarative.partition_routers.async_job_partition_router import (
+    AsyncJobPartitionRouter,
+)
 from airbyte_cdk.sources.declarative.partition_routers.substream_partition_router import (
     ParentStreamConfig,
 )
@@ -2146,18 +2149,24 @@ class ModelToComponentFactory:
             urls_extractor=urls_extractor,
         )
 
-        return AsyncRetriever(
+        async_job_partition_router = AsyncJobPartitionRouter(
             job_orchestrator_factory=lambda stream_slices: AsyncJobOrchestrator(
                 job_repository,
                 stream_slices,
-                JobTracker(
-                    1
-                ),  # FIXME eventually make the number of concurrent jobs in the API configurable. Until then, we limit to 1
+                JobTracker(1),
+                # FIXME eventually make the number of concurrent jobs in the API configurable. Until then, we limit to 1
                 self._message_repository,
-                has_bulk_parent=False,  # FIXME work would need to be done here in order to detect if a stream as a parent stream that is bulk
+                has_bulk_parent=False,
+                # FIXME work would need to be done here in order to detect if a stream as a parent stream that is bulk
             ),
-            record_selector=record_selector,
             stream_slicer=stream_slicer,
+            config=config,
+            parameters=model.parameters or {},
+        )
+
+        return AsyncRetriever(
+            record_selector=record_selector,
+            stream_slicer=async_job_partition_router,
             config=config,
             parameters=model.parameters or {},
         )
--- a/airbyte_cdk/sources/declarative/partition_routers/async_job_partition_router.py
+++ b/airbyte_cdk/sources/declarative/partition_routers/async_job_partition_router.py
@@ -48,6 +48,13 @@ class AsyncJobPartitionRouter(StreamSlicer):
             )
 
     def fetch_records(self, partition: AsyncPartition) -> Iterable[Mapping[str, Any]]:
+        """
+        This method of fetching records extends beyond what a PartitionRouter/StreamSlicer should
+        be responsible for. However, this was added in because the JobOrchestrator is required to
+        retrieve records. And without defining fetch_records() on this class, we're stuck with either
+        passing the JobOrchestrator to the AsyncRetriever or storing it on multiple classes.
+        """
+
         if not self._job_orchestrator:
             raise AirbyteTracedException(
                 message="Invalid state within AsyncJobRetriever. Please contact Airbyte Support",
--- a/airbyte_cdk/sources/declarative/partition_routers/async_job_partition_router.py
+++ b/airbyte_cdk/sources/declarative/partition_routers/async_job_partition_router.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
+
+from dataclasses import InitVar, dataclass, field
+from typing import Any, Callable, Iterable, Mapping, Optional
+
+from airbyte_cdk.models import FailureType
+from airbyte_cdk.sources.declarative.async_job.job_orchestrator import (
+    AsyncJobOrchestrator,
+    AsyncPartition,
+)
+from airbyte_cdk.sources.declarative.partition_routers.single_partition_router import (
+    SinglePartitionRouter,
+)
+from airbyte_cdk.sources.streams.concurrent.partitions.stream_slicer import StreamSlicer
+from airbyte_cdk.sources.types import Config, StreamSlice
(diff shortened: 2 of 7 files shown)


airbyte_cdk/sources/declarative/partition_routers/async_job_partition_router.py
```diff
@@ -0,0 +1,58 @@
...
+
+from airbyte_cdk.models import FailureType
+from airbyte_cdk.sources.declarative.async_job.job_orchestrator import (
+    AsyncJobOrchestrator,
+    AsyncPartition,
+)
+from airbyte_cdk.sources.declarative.partition_routers.single_partition_router import (
+    SinglePartitionRouter,
+)
+from airbyte_cdk.sources.streams.concurrent.partitions.stream_slicer import StreamSlicer
```
brianjlai, line 14: This was intentional to use the concurrent `StreamSlicer` interface. There is probably an argument for consistency to use the declarative `StreamSlicer`, but that would require we implement unused methods for providing request parameters and we expect the concurrent interface to be the long term interface.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2146,22 +2149,28 @@ def create_async_retriever(
...
-        return AsyncRetriever(
+        async_job_partition_router = AsyncJobPartitionRouter(
             job_orchestrator_factory=lambda stream_slices: AsyncJobOrchestrator(
                 job_repository,
                 stream_slices,
-                JobTracker(
-                    1
-                ),  # FIXME eventually make the number of concurrent jobs in the API configurable. Until then, we limit to 1
+                JobTracker(1),
+                # FIXME eventually make the number of concurrent jobs in the API configurable. Until then, we limit to 1
```
maxi297, line 2157: [outside of the scope of this PR but important] This is very interesting! I don't remember this very well but it seems that we can only have one concurrent job on the API side at the same time until we tackle this `FIXME`. 

A couple of notes/questions:
* [source-salesforce](https://github.com/airbytehq/airbyte/blob/2370c070f42109a7c9c055629b468835909f97d9/airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py#L77) is using 5 concurrent jobs and it seems to be working well so we have a case where the limit is not 1 that has been tested in prod
* It feels like the instantiation of this `JobTracker` is wrong. All instances of `AsyncJobOrchestrator` should probably share the same instance as the limit on the API side is usually not per stream. It is done properly in source-salesforce and probably is not an issue source-sendgrid as there is only one stream with a job
* @artem1205 do you know how many concurrent job we can have for the Amazon Ads API and for the Amazon Seller Partner API? If there benefit in making it configurable now?

airbyte_cdk/sources/declarative/retrievers/async_retriever.py
```diff
@@ -116,7 +93,7 @@ def read_records(
     ) -> Iterable[StreamData]:
         stream_state: StreamState = self._get_stream_state()
         partition: AsyncPartition = self._validate_and_get_stream_slice_partition(stream_slice)
-        records: Iterable[Mapping[str, Any]] = self._job_orchestrator.fetch_records(partition)
+        records: Iterable[Mapping[str, Any]] = self.stream_slicer.fetch_records(partition)
```
maxi297, line 96: This line was surprising to me. I don't know why but the fact that we can retrieve records from a stream_slicer hurts my brain. I'm not strongly oppose to it though and maybe this is just a naming thing? I think I understand the challenge around the instantiation of the JobOrchestrator though and I'm out of naming for this one haha

maxi297 (commented):
I think I'm fine with those changes but I'd like to see the tests to confirm

airbyte_cdk/sources/declarative/retrievers/async_retriever.py
brianjlai: yeah so i really went back and forth on this one a lot yesterday. I agree that it feels a bit wrong that the new stream slicer is also providing the ability to fetch records since it's not in the interface and effectively this only works because the `AsyncRetriever` explicitly types `AsyncJobPartitionRouter` instead of the generic `StreamSlicer` interface.

The original way I had this was that I exposed a `AsyncJobPartitionRouter.job_orchestrator` property method, and kept old method:
```
def _job_orchestrator(self) -> AsyncJobOrchestrator:
    if not self.stream_slicer.job_orchestrator:
        raise AirbyteTracedException(
            message="Invalid state within AsyncJobRetriever. Please contact Airbyte Support",
            internal_message="AsyncPartitionRepository is expected to be accessed only after `stream_slices`",
            failure_type=FailureType.system_error,
        )

    return self.stream_slicer.job_orchestrator
```

So either way was left with a bit of an unfortunate side effect of relying on methods that only exist on `AsyncJobPartitionRouter` and not `StreamSlicer`. I'm not opposed to switching it back to this if you feel strongly about not intermixing `fetch_records()` onto the `AsyncJobPartitionRouter`

brianjlai:
> I think I'm fine with those changes but I'd like to see the tests to confirm

Yup definitely. I'm adding a couple more unit tests today and i tested this locally on Sendgrid, but I plan to regression test this against live traffic on `source-sendgrid` to validate that this is a drop in replacement end to end.

brianjlai:
live test results: https://github.com/airbytehq/airbyte/actions/runs/12360671390

results analysis:

- Expected record count for `contacts` (the only async reports stream) matches
- the mismatches on the config where `is_resumable` is going from True -> False is expected since this is moving more full refresh streams to concurrent which doesn't have RFR
- there is an added change to the manifest because `6.7.2` made a change where the ResponseToFileExtractor was no longer assigned at default. This might have fixed a bug, but Sendgrid relied on this and since its an isolated use case, it makes more sense to just fix the manifest once if we're made the CDK work properly. line of code here: https://github.com/airbytehq/airbyte-python-cdk/pull/50/files#diff-1d9bc4ca384e5c00867da05e9db9d919aba908c59bb73aab83883b9ed20def05R2072

maxi297 (approved):
LGTM! Thanks @brianjlai

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: i'll write up a ticket for this since it's something we'll want configurable at least at the manifest level (or customer config level but that we can decide later).

PR #175, opened 2024-12-16
feat(HttpComponentsResolver): added stream slices to HttpComponentsResolver

darynaishchenko:
## Summary by CodeRabbit

- **New Features**
	- Introduced `HttpComponentsResolver` and `ConfigComponentsResolver` for enhanced component resolution.
	- Added properties (`stream_slice` and `stream_template_config`) to `ComponentMappingDefinition` for improved mapping flexibility.

- **Bug Fixes**
	- Improved handling of stream slices in the `resolve_components` method of `HttpComponentsResolver`.

- **Tests**
	- Enhanced tests for `HttpComponentsResolver` to cover scenarios involving stream slices and parent-child relationships.
	- Added new test functions to validate resolver behavior with dynamic streams and parent streams.


airbyte_cdk/sources/declarative/resolvers/http_components_resolver.py
```diff
@@ -88,19 +88,25 @@ def resolve_components(
...
 
-            for resolved_component in self._resolved_components:
-                valid_types = (
-                    (resolved_component.value_type,) if resolved_component.value_type else None
-                )
-                value = resolved_component.value.eval(
-                    self.config, valid_types=valid_types, **kwargs
-                )
+        for stream_slice in stream_slices:
+            for components_values in self.retriever.read_records({}, stream_slice):
```
aldogonzalez8, line 94: Can you add name of the parameters here?

aldogonzalez8:
These changes make sense to me, but can we have a test with manifest configuration including the slicer?

airbyte_cdk/sources/declarative/resolvers/http_components_resolver.py
```diff
@@ -88,19 +88,25 @@ def resolve_components(
         """
         kwargs = {"stream_template_config": stream_template_config}
 
-        for components_values in self.retriever.read_records({}):
-            updated_config = deepcopy(stream_template_config)
-            kwargs["components_values"] = components_values  # type: ignore[assignment] # component_values will always be of type Mapping[str, Any]
+        stream_slices = self.retriever.stream_slices() if self.retriever.stream_slicer else [{}]
```
lazebnyi, line 91: Do you know of any cases where `stream_slicer` is not present in `retriever`?

https://github.com/airbytehq/airbyte-python-cdk/blob/adef1e8fb38fb792825c6a4c08d8e81b0fab3851/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py#L82

I also think that using `[None]` is safer.

unit_tests/sources/declarative/resolvers/test_http_components_resolver.py
```diff
@@ -221,6 +221,7 @@ def test_http_components_resolver(
 ):
     mock_retriever = MagicMock()
     mock_retriever.read_records.return_value = retriever_data
+    mock_retriever.stream_slices.return_value = [{}]
```
lazebnyi, line 224: Could we test with few slices?

lazebnyi (changes requested):
Overall, LGTM. 

However, we need to add another test case and resolve some comments.

airbyte_cdk/sources/declarative/resolvers/http_components_resolver.py
darynaishchenko: I don't know a case when stream_slicer not present in the retriever, also  default stream slicer that you are referring to uses single partition router [here](https://github.com/airbytehq/airbyte-python-cdk/blob/adef1e8fb38fb792825c6a4c08d8e81b0fab3851/airbyte_cdk/sources/declarative/partition_routers/single_partition_router.py#L51). So I think we can go without the condition at all.
 I'll update the code.

airbyte_cdk/sources/declarative/resolvers/http_components_resolver.py
darynaishchenko: updated

unit_tests/sources/declarative/resolvers/test_http_components_resolver.py
darynaishchenko: added more tests

lazebnyi (approved):
lgtm

PR #176, opened 2024-12-16
feat(low-code cdk): add transformation to dynamic schema loader

lazebnyi:
## What

During dynamic schema generation, we need to transform keys or add fields to enhance the schema's compatibility and functionality.

## How 

A schema transformations component has been integrated into the dynamic schema loader. This component is responsible for managing a list of transformations, enabling us to easily modify schema keys or insert additional fields as needed.

## Summary by CodeRabbit

- **New Features**
	- Introduced `schema_transformations` to enhance schema manipulation capabilities, allowing for various transformations such as adding/removing fields and key formatting.
	- Added support for the `FlattenFields` transformation type in stream configuration.

- **Bug Fixes**
	- Updated test cases to reflect schema changes, ensuring accuracy in expected outputs.

- **Tests**
	- Added new test case for validating the `FlattenFields` transformation and adjusted existing tests for updated schema definitions.


unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -2200,21 +2213,25 @@ def test_add_fields_no_value_type(self):
...
+                            ),
+                            value_type=None,
+                            parameters={},
+                        )
+                    ],
+                    parameters={},
+                )
+            ]
+            * 2
+        )  # Entity will be present twice as we resolve parameters and add the component to both the record
```
maxi297, line 2233: That seems odd, right? Why would the transformation be there twice? If seems especially dangerous if we eventually have transformations that are not idempotent

aldogonzalez8: Is this what is happening?

- Stream parameters have transformations (T1).
- We resolve parameters and pass transformations to stream and record selector(?). Then Stream has T1 and Record selector receives T1.
- Inside of create_record_selector() we check if current model RecordSelector has transformations and we add it, but, we already received the transformation from stream creation logic.
- We end with record selector transformations [T1, T1] 

Im not sure if I'm following correct logic but I agree it is odd. I would say we should check the transformation is not already in the array or something else that keep us away from duplication.

lazebnyi: Implementation updated with schema transformation

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2514,6 +2514,16 @@ definitions:
...
+      transformations:
+        title: Transformations
+        description: A list of transformations to be applied to each output record.
+        type: array
+        items:
+          anyOf:
+            - "$ref": "#/definitions/AddFields"
+            - "$ref": "#/definitions/CustomTransformation"
+            - "$ref": "#/definitions/RemoveFields"
+            - "$ref": "#/definitions/KeysToLower"
```
aldogonzalez8: Can this list be moved to another definition? We would need to maintain fewer locations as DeclarativeStream and DynamicSchemaLoader now have transformations.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1767,6 +1767,17 @@ definitions:
...
+        title: Schema Transformations
+        description: A list of transformations to be applied to the schema.
+        type: array
+        items:
+          anyOf:
+            - "$ref": "#/definitions/AddFields"
+            - "$ref": "#/definitions/CustomTransformation"
+            - "$ref": "#/definitions/RemoveFields"
+            - "$ref": "#/definitions/KeysToLower"
+            - "$ref": "#/definitions/KeysToSnakeCase"
```
aldogonzalez8, line 1780: Can this list be moved to another definition? We would need to maintain fewer locations as DeclarativeStream and DynamicSchemaLoader now have transformations.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1831,6 +1831,21 @@ class DynamicSchemaLoader(BaseModel):
...
         title="Retriever",
     )
+    schema_transformations: Optional[
+        List[
+            Union[
+                AddFields,
+                CustomTransformation,
+                RemoveFields,
+                KeysToLower,
+                KeysToSnakeCase,
```
aldogonzalez8, line 1841: > Can this list be moved to another definition? We would need to maintain fewer locations as DeclarativeStream and DynamicSchemaLoader now have transformations.

Actually, this is the right place for the comment, right? As yaml is automatically generated.

aldogonzalez8: For example, we guess we want the "flatten fields" transformation to be added here, so we just do it once on a common list.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1831,6 +1831,21 @@ class DynamicSchemaLoader(BaseModel):
         description="Component used to coordinate how records are extracted across stream slices and request pages.",
         title="Retriever",
     )
+    schema_transformations: Optional[
```
aldogonzalez8, line 1834: nit: isn't it redundant to name it schema_transformations, as we are already in the schema area? Other fields are not named schema_title or schema_retriever

aldogonzalez8:
I liked the idea of having it in the record selector as we would resolve this in other structures. However, if we want transformations at the dynamic schema_loader level, should we also file an issue to have transformations for components_resolver?

cc @maxi297

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
lazebnyi: This component is used to transform a schema. If we refer to it simply as a 'transformation', it could be misleading. It might suggest that transformations are used to alter retrieved records, which is not the case. The actual process involves first retrieving the record and, if necessary (such as when a schema API returns more than one schema), filtering them. Only after extracting the schema from the retrieved records do we proceed to transform it.

lazebnyi:
@aldogonzalez8 Do you have cases when we need transform records in `components_resolver`?

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
aldogonzalez8: I'm ok leaving it as schema_transformations, but not sure it would be misleading as the DynamicSchemaLoader context would provide sufficient context IMO but not a reason to hold this pr for this.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
lazebnyi: You are correct; this is an autogenerated file. Therefore, all efforts to unify the interface should be handled in the YAML schema file, which could lead to more complex logic or misleading as not all components from list of refs can be support by parent component . As such, it's best to leave this as it is and bring up this issue during our next brainstorming session.

lazebnyi:
Could you help me undsersted why do you need this transformation at the `components_resolver` level? I'm asking because the resolver is typically used to fetch records, then iterates through them, and dynamically create the stream and  populates the stream template with values from these records.

unit_tests/sources/declarative/schema/test_dynamic_schema_loader.py
```diff
@@ -64,6 +64,21 @@
...
                     "paginator": {"type": "NoPagination"},
                 },
+                "schema_transformations": [
+                    {
+                        "type": "AddFields",
+                        "fields": [
+                            {
+                                "type": "AddedFieldDefinition",
+                                "path": ["StaticField"],
+                                "value": "{{ {'type': ['null', 'string']} }}",
```
maxi297, line 74: nit: It does not seem a good UX for the dev to have to set up the value as an object as opposed to:
```
                    {
                        "type": <a type>,
                        "fields": [
                            {
                                "path": StaticField,
                                "value": string,
                            }
                        ],
                    },
```

Why I flagged this as nit is it feels like we could always improve the UX be adding a new `AddFieldsSchema` transformation that would allow to do that, right? I just want to validate that adding this would work and be backward compatible so we're not closing any door in the future

maxi297 (approved):
LGTM! :shipit:

maxi297:
Regarding @aldogonzalez8 's comment on the transformation for component resolver, it will be needed for source-airtable as well in order to put parent stream information on the records so that we can get those values available when we resolve. @darynaishchenko

aldogonzalez8:
> If I understand correctly, with this transformation, you extract certain fields to the properties level and then reference them from the `components_mapping`. But what is the benefit of doing this? Is it possible to collect values directly in the `components_mapping` and skip the transformation stage?
> 
> cc @aldogonzalez8

Yes, I could transform data on the extractor that receives such data, but I liked how this organically worked with transformations. This can happen in many ways, I guess. Still I feel that in the future in any place where we retrieve records/data it would be nice the chance to have transformations for such records/data.

lazebnyi:
@maxi297 For airtable, I don't think we need it at all, as we will have access to parent fields in the components_mapping once we add stream_slicing to the http resolver. - https://github.com/airbytehq/airbyte-python-cdk/blob/ec9e480b98760b3f91e5c8e30a656c96515c537a/airbyte_cdk/sources/declarative/resolvers/http_components_resolver.py#L97

aldogonzalez8 (approved):
APPROVED

lazebnyi:
@maxi297 @aldogonzalez8 @darynaishchenko For now, the migration to 'manifest only' (migration without any custom components) for source google sheets or source airtable is not hindered by the absence of transformations in the components resolver. I prefer to leave the components resolver without transformation, as once the transformation is moved to the record selector, we will achieve this for free.

unit_tests/sources/declarative/schema/test_dynamic_schema_loader.py
lazebnyi: Yes, we can define a more schema transformation-oriented interface in the AddFieldsSchema transformation. So I think we ok here.

PR #179, opened 2024-12-17
feat: add Composite Raw Decoder

artem1205:
# What

Resolving Source Amazon Seller Partner CDK migration

add composite decoder to apply `decompressors` | `decoders` consequently 

>[!NOTE]
> memory test usage moved to `unit_tests/sources/declarative/decoders/test_decoders_memory_usage.py`. `JsonlDecoder` (as well as others _streamable_ decoders)  should be deleted/replaced with `CompositeRawDecoder`+ `parser` 

## Summary by CodeRabbit

- **New Features**
	- Introduced `CompositeRawDecoder`, `GzipParser`, `JsonLineParser`, and `CsvParser` for enhanced data decoding capabilities.
	- Updated `SimpleRetriever` to support the new decoding options.
- **Bug Fixes**
	- Improved handling of Gzip-compressed data formats.
- **Tests**
	- Added unit tests for `CompositeRawDecoder` and its associated parsers, validating decoding functionalities.
	- Introduced memory usage tests for JSON line decoders.
- **Chores**
	- Minor modifications to workflow file for version tag handling.


airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,127 @@
...
+    def read(self, __n: int) -> bytes: ...
+
+
+@dataclass
+class Parser(ABC):
+    inner_parser: Optional["Parser"] = None
+
+    @abstractmethod
+    def parse(
+        self, data: BufferedDataInput | Iterable, *args, **kwargs
```
maxi297, line 28: I'm a bit confused by this interface. Does that mean that all the users with have to check the `data` parameter as such:
```
    def parse(
        self, data: BufferedDataInput | Iterable, *args, **kwargs
    ) -> Generator[MutableMapping[str, Any], None, None]:
        if isinstance(data, BufferedDataInput):
            < do X >
        elif isinstance(data, Iterable):
            < do Y >
        else:
            raise ValueError(f"Unexpected type {type(data)}")
```

I also don't see the usage of `BufferedDataInput` in the sample parsers you provided below, maybe I missed it. Can you show me how this would be used?

maxi297 (commented):
I added a couple of questions to improve my understanding

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    at least read to _n bytes method should be supported to align with BufferedReader and avoid OOM.
+    """
+
+    def read(self, __n: int) -> bytes: ...
+
+
+@dataclass
+class Parser(ABC):
+    inner_parser: Optional["Parser"] = None
```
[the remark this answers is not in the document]
artem1205: > Should this be removed from the interface and the parser implementations will decide whatever they use internally?

this was added to interface to use declarative approach and wrap parser-in-parser-in-parser as many times as it is needed, e.g.:
```yaml

- decoder:
   type: CompositeRawDecoder
   parser: 
     type: GzipParser
     inner_parser:
       type: parser1
         inner_parser:
           type: parser2
...
```

In my understanding parent parsers does not know about inner_parser implementation and all they do is just pass transformed object down through the pipeline.

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
artem1205: > Does that mean that all the users with have to check the data parameter as such:

Usually not, cause I expect user to choose _valid_ parser sequence.

> I also don't see the usage of BufferedDataInput in the sample parsers you provided below, maybe I missed it. Can you show me how this would be used?

We start processing from passing `response.raw` in 
```python3
self.parser.parse(data=response.raw)
```
which in fact is readable (buffered) object.

any decompressor, like gzip or zip returns decompressor object , which in turn is also readable (buffered).

maxi297: Ok! I really like the idea of having a buffered reader as a part of the interface. Maybe it should even be a by more meaning that the file-based [CsvParser is relying on more methods than just "read"](https://github.com/airbytehq/airbyte-python-cdk/blob/57e1b524e2c55b2c1e8b80da38c0cfda3c6a385a/airbyte_cdk/sources/file_based/file_types/csv_parser.py#L151). 

So I think my question now is: can we remove `Iterable` from the interface?

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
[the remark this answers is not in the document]
maxi297: My understanding is that it'll be part of the implementations but not the interface. So for example in `declarative_component_schema.yaml`, we will have:

```
  DefaultDecoder:
    title: Decoder with a parser
    type: object
    required:
      - type
      - parser
    properties:
      type:
        type: string
        enum: [DefaultDecoder]
      parser:
        anyOf:
          - "$ref": "#/definitions/CsvParser"
          - "$ref": "#/definitions/JsonlParser"
          <...>

  CsvParser:
    title: CSV Parser
    type: object
    required:
      - type
    properties:
      type:
        type: string
        enum: [CsvParser]
      <probably some CSV parsing options that we have for file based>
  GzipParser:
    title: GZIP Parser
    type: object
    required:
      - type
      - underlying_parser
    properties:
      type:
        type: string
        enum: [GzipParser]  
      underlying_parser:
        anyOf:
          - "$ref": "#/definitions/CsvParser"
          - "$ref": "#/definitions/JsonlParser"
          <...>
```

Each of these parser components will have their instantiation method in the model_to_component_factory and if they require an inner parser (like the GzipParser), we will instante that parser at that point. So the fact that a parser uses a parser underneath does not need to be exposed as part of the Python interface.

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
artem1205: I guess yes, refactored interface to `BufferedIOBase`

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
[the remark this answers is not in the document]
artem1205: agreed, refactored

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+from airbyte_cdk.sources.declarative.decoders.decoder import Decoder
+
+logger = logging.getLogger("airbyte")
+
+
+@dataclass
+class Parser(ABC):
+    @abstractmethod
+    def parse(
+        self, data: BufferedIOBase, *args, **kwargs
```
maxi297, line 22: Just to make sure I understand: Why do we have `*args, **kwargs` in the interface?

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+                yield json.loads(line.decode(self.encoding))
+            except json.JSONDecodeError:
+                logger.warning(f"Cannot decode/parse line {line} as JSON")
+                # Handle invalid JSON lines gracefully (e.g., log and skip)
+                pass
+
+
+@dataclass
+class CsvParser(Parser):
+    # TODO: add more parameters: see read_csv for more details, e.g.: quotechar, headers,
```
maxi297, line 62: nit: I think the `TODO` should be `migrate implementation to re-use file-base classes`

maxi297 (commented):
That's cool! I think this is a good improvement

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
artem1205: no reason (or real use-case) for now, can be deleted.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1701,6 +1722,12 @@ def create_jsonl_decoder(
     ) -> JsonlDecoder:
         return JsonlDecoder(parameters={})
 
+    @staticmethod
+    def create_jsonline_parser(
```
brianjlai, line 1726: nit should be `create_json_line_parser`

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,102 @@
...
+    delimiter: Optional[str] = ","
+
+    def parse(
+        self,
+        data: BufferedIOBase,
+    ) -> Generator[MutableMapping[str, Any], None, None]:
+        """
+        Parse CSV data from decompressed bytes.
+        """
+        reader = pd.read_csv(  # type: ignore
```
brianjlai, line 74: was there a reason why we're opting for `pandas`? I think this discussion was brought up while we were implementing the file based cdk, and ultimately we decided to use the `csv` library for reading from CSVs. The interfaces are similar but different in a few ways.

The main reason I ask is that if we plan to migrate this implementation to use file based, it would probably make sense to use the same library. How much lift do you estimate it would be to use `csv` instead. And for example on how its used, you can see `csv_parser.py`. If the lift is reasonable I think we should use it, but if the effort is high, I would be okay leaving this as is

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2766,9 +2766,59 @@ definitions:
...
+        type: string
+        enum: [ GzipParser ]
+      inner_parser:
+        anyOf:
+          - "$ref": "#/definitions/JsonLineParser"
+          - "$ref": "#/definitions/CsvParser"
+  JsonLineParser:
+    type: object
+    properties:
+      encoding:
```
brianjlai, line 2805: I think this definition is missing in the definition, the requirement that every component be explicitly typed:
```
    required:
      - type
    properties:
      type:
        type: string
        enum: [JsonLineParser]
```

brianjlai (commented):
a few comments and small change requests. overall good though

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
artem1205: fixed

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
```diff
@@ -0,0 +1,102 @@
...
+            data, sep=self.delimiter, iterator=True, dtype=object, encoding=self.encoding
+        )
+        for chunk in reader:
+            chunk = chunk.replace({nan: None}).to_dict(orient="records")
+            for row in chunk:
+                yield row
+
+
+@dataclass
+class CompositeRawDecoder(Decoder):
```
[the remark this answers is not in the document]
artem1205: renamed

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2766,9 +2766,59 @@ definitions:
...
+  CompositeRawDecoder:
+    description: "(This is experimental, use at your own risk)"
+    type: object
+    required:
+      - type
+      - parser
+    properties:
+      type:
+        type: string
+        enum: [ CompositeRawDecoder ]
```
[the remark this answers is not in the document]
artem1205: formatted with prettier `poe format-fix`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
artem1205: renamed

airbyte_cdk/sources/declarative/decoders/composite_decoder.py
artem1205: > was there a reason why we're opting for pandas?

Copypasted from our commonly used approach (see `ResponseToFileExtractor`)

> it would probably make sense to use the same library.

Agreed, refactored to `csv.DictReader`

pnilan (approved):
Looks good.

brianjlai (approved):
thanks for making the changes! 🚢

PR #182, opened 2024-12-19
fix: (DeclarativeOAuthFlow) - allow `DeclarativeOauth2Authenticator` to use `access_token` directly when no `token_refresh_endpoint` or `refresh_token` values are provided.

bazarnov:
## What
Resolving:
- https://github.com/airbytehq/airbyte-internal-issues/issues/11176

## How
- make `token_refresh_endpoint` to be `Optional`
- add a new optional field to the model `access_token_value`
- make the `access_token_value` to be set in `get_auth_header()`
- covered the change with the unit_test `test_get_auth_header_without_refresh_token_and_without_refresh_token_endpoint`

## Summary by CodeRabbit

- **New Features**
	- `token_refresh_endpoint` is now an optional parameter in the OAuth authentication process, allowing for greater flexibility.
	- Introduced `access_token_value` to specify an access token directly, bypassing the token refresh process.
	- Added a new optional `session_token` field for alternative authentication methods.

- **Bug Fixes**
	- Enhanced error handling in API key authentication to prevent simultaneous settings of `inject_into` and `header`.

- **Tests**
	- Added a new test for the behavior of the authentication header when an `access_token_value` is provided without a `token_refresh_endpoint` or refresh token.


unit_tests/sources/declarative/auth/test_oauth.py
```diff
@@ -129,6 +129,25 @@ def test_refresh_without_refresh_token(self):
...
+    def test_get_auth_header_without_refresh_token_and_without_refresh_token_endpoint(self):
+        """
+        Coverred the case when the `access_token_value` is supplied,
+        without `token_refresh_endpoint` or `refresh_token` provided.
+
+        In this case, it's expected to have the `access_token_value` provided to return the permanent `auth header`,
+        contains the authentication.
+        """
+        oauth = DeclarativeOauth2Authenticator(
+            access_token_value="my_access_token_value",
```
bnchrch, line 141: ❓ Why do `access_token_value="my_access_token_value",` and `client_secret="{{ config['client_secret'] }}",` differ in their semantics?

I assumed access_token (like refresh token, client id, and client secret) would be retrieved from the config.

e.g. 
```
access_token_value ="{{ config['access_token'] }}",
```

bazarnov: Will correct this one shortly.

bnchrch (approved):
TY!

PR #183, opened 2024-12-20
feat(low-code): added keys replace transformation

darynaishchenko:
Added `KeysReplaceTransformations` to be able to replace symbols in record keys.

### DIFF
--- a/airbyte_cdk/sources/declarative/transformations/keys_replace_transformation.py
+++ b/airbyte_cdk/sources/declarative/transformations/keys_replace_transformation.py
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/transformations/keys_replace_transformation.py
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
```
aldogonzalez8, line 2: nit: you worked on this for a long time.
jk I have left a few in my PRs, too.

aldogonzalez8 (approved):
APPROVED

aldogonzalez8:
Changes makes sense to me, just left annoying nit comment,

maxi297 (approved):
The code is good, the tests are clean! Can we add a bit more context on the PR description to mention which connector this is intended for? This will be valuable context if we need to debug something related to this feature

PR #187, opened 2024-12-23
feat: add DateTimeNormalizerFields transformation

artem1205:
# What

Resolving Source Amazon Seller Partner CDK migration

add new type for transformation: `DateTimeNormalizer`

Reason:

 standart `datetime` library cannot parse correctly some of the timezone formats, e.g. `2022-07-11 01:34:18 PDT`


artem1205:
Closing as non-relevant as for now.

This transformations should be applied during normalization, cause we need schema to distinguish date and date-time fields

We should start from https://github.com/airbytehq/airbyte-internal-issues/issues/8861 to refactor general interface and then proceed with specific schema normalization functions.

PR #192, opened 2024-12-30
fix: `Async Retriever` change url path for download retriever

artem1205:
# What

-  pass `url` as `extra_field` to ignore it in state manager
- add `transformations` to download retriever

> [!CAUTION]
> changing url path in `stream_slice` for download retriever is technically a **_breaking change_**, but I don't want to bump major version since `AsyncRetriever` is an `ExperimentalClass`

# Reason

see https://github.com/airbytehq/airbyte-python-cdk/pull/192#discussion_r1901702963

## Release Notes

- **New Features**
	- Enhanced `StreamSlice` class with boolean evaluation support
	- Improved record selector with customizable transformations

- **Improvements**
	- Updated async HTTP job repository to provide more flexible stream slice construction
	- Refined URL retrieval mechanism in job processing tests

The changes introduce more dynamic and flexible data processing capabilities within the Airbyte CDK, allowing for more nuanced record transformations and stream handling.

### DIFF
--- a/airbyte_cdk/sources/declarative/requesters/http_job_repository.py
+++ b/airbyte_cdk/sources/declarative/requesters/http_job_repository.py
@@ -189,7 +189,8 @@ class AsyncHttpJobRepository(AsyncJobRepository):
         for url in self.urls_extractor.extract_records(
             self._polling_job_response_by_id[job.api_job_id()]
         ):
-            stream_slice: StreamSlice = StreamSlice(partition={"url": url}, cursor_slice={})
+            stream_slice = job.job_parameters()
+            stream_slice.extra_fields.update({"url": url})
             for message in self.download_retriever.read_records({}, stream_slice):
                 if isinstance(message, Record):
                     yield message.data
--- a/airbyte_cdk/sources/declarative/requesters/http_job_repository.py
+++ b/airbyte_cdk/sources/declarative/requesters/http_job_repository.py
@@ -189,8 +189,12 @@ class AsyncHttpJobRepository(AsyncJobRepository):
         for url in self.urls_extractor.extract_records(
             self._polling_job_response_by_id[job.api_job_id()]
         ):
-            stream_slice = job.job_parameters()
-            stream_slice.extra_fields.update({"url": url})
+            job_slice = job.job_parameters()
+            stream_slice = StreamSlice(
+                partition=job_slice.partition,
+                cursor_slice=job_slice.cursor_slice,
+                extra_fields={**job_slice.extra_fields, "url": url},
+            )
             for message in self.download_retriever.read_records({}, stream_slice):
                 if isinstance(message, Record):
                     yield message.data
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/requesters/http_job_repository.py
```diff
@@ -189,7 +189,8 @@ def fetch_records(self, job: AsyncJob) -> Iterable[Mapping[str, Any]]:
         for url in self.urls_extractor.extract_records(
             self._polling_job_response_by_id[job.api_job_id()]
         ):
-            stream_slice: StreamSlice = StreamSlice(partition={"url": url}, cursor_slice={})
+            stream_slice = job.job_parameters()
+            stream_slice.extra_fields.update({"url": url})
```
brianjlai, line 193: I have mixed feelings about having to make the stream_slice's `extra_fields` mutable. Even though in this case it might be fine, I like that in our code the contract is that a given slice's underlying mappings won't change. I think we should try to keep the interface as is where `extra_fields` is just a Mapping

One way we could do this is by constructing a new slice by copying things into a new slice
```
job_slice = job.job_parameters()
StreamSlice(
    partition=job_slice.partition,
    cursor_slice=job_slice.cursor_slice,
    extra_fields={**job_slice.extra_fields, "url": url},
)
```

And the other question, why was the URL being in the partition creating issues for the state_manager, I'm not fully against it being an extra field, but I feel like the partition is the natural place for a usability standpoint since that is what we're partitioning on. So I'd want to understand why we need to move it

artem1205: > And the other question, why was the URL being in the partition creating issues for the state_manager, I'm not fully against it being an extra field, but I feel like the partition is the natural place for a usability standpoint since that is what we're partitioning on. So I'd want to understand why we need to move it

Long story short: `url` field in partition changes the way we store (and serialize) `StreamSlice` keys.

1. in amazon Ads we use incremental sync, so we have the following slice that should be passed to the `download_retriever` ( I redefined it at line 198 for testing purposes)
image
2. during the read operation, we try to `observe` (L153), but the `self._cursor_per_partition` has only initial partition `{"parent_slice":{},"profileId":1}`
image
3. and therefore cannot be updated (KeyError) by `self._to_partition_key(stream_slice.partition)`
image

That is why I decided to move url field to extra_fields => to exclude it from serialization.

brianjlai: Ah okay thank you for the clear explanation. It's a shame that we're left with using the `extra_fields` as a means of access to avoid it being included. Not something that I am completely opposed to so from a code perspective things look good to me so I'll approve

But because @maxi297 implemented most of the original Async job work, I think we should just get his eyes on the final design as well. Might be worth waiting the extra day for him when he is back.

brianjlai (approved):

PR #194, opened 2024-12-31
feat: add `CustomSchemaNormalization`

artem1205:
# What

- add `CustomSchemaNormalization` to declarative schema

## Summary by CodeRabbit

- **New Features**
	- Introduced a custom schema normalization strategy for record selection.
	- Enhanced flexibility in handling schema normalization types.

- **Improvements**
	- Updated the `RecordSelector` to support custom normalization approaches.
	- Expanded schema transformation capabilities with more dynamic configuration options.

- **Technical Updates**
	- Refactored schema normalization handling in the component factory.
	- Added an abstract base class for type transformation logic.
	- Made `TypeTransformer` available for external use in the module interface.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2555,21 +2555,48 @@ definitions:
           - "$ref": "#/definitions/CustomRecordFilter"
           - "$ref": "#/definitions/RecordFilter"
       schema_normalization:
-        "$ref": "#/definitions/SchemaNormalization"
-        default: None
+        anyOf:
+         - "$ref": "#/definitions/SchemaNormalization"
+         - "$ref": "#/definitions/CustomSchemaNormalization"
+#        default: None
       $parameters:
         type: object
         additionalProperties: true
   SchemaNormalization:
     title: Schema Normalization
     description: Responsible for normalization according to the schema.
-    type: string
-    enum:
-      - None
-      - Default
-    examples:
-      - None
-      - Default
+    type: object
+    required:
+      - type
+    properties:
+      type:
+        type: string
+        enum: [ SchemaNormalization ]
+      transform_config:
+        type: string
+        enum:
+          - None
+          - Default
+  CustomSchemaNormalization:
+    title: Schema Normalization
+    description: Responsible for normalization according to the schema.
+    type: object
+    required:
+      - type
+    properties:
+      type:
+        type: string
+        enum: [ CustomSchemaNormalization ]
+      class_name:
+        title: Class Name
+        description: Fully-qualified name of the class that will be implementing the custom decoding. Has to be a sub class of Decoder. The format is `source_<name>.<package>.<class_name>`.
+        type: string
+        additionalProperties: true
+        examples:
+          - "source_amazon_ads.components.LedgerDetailedViewReportsTypeTransformer"
+      $parameters:
+        type: object
+        additionalProperties: true
   RemoveFields:
     title: Remove Fields
     description: A transformation which removes fields from a record. The fields removed are designated using FieldPointers. During transformation, if a field or any of its parents does not exist in the record, no error is thrown.
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2558,25 +2558,20 @@ definitions:
         anyOf:
          - "$ref": "#/definitions/SchemaNormalization"
          - "$ref": "#/definitions/CustomSchemaNormalization"
-#        default: None
+        default: None
       $parameters:
         type: object
         additionalProperties: true
   SchemaNormalization:
     title: Schema Normalization
     description: Responsible for normalization according to the schema.
-    type: object
-    required:
-      - type
-    properties:
-      type:
-        type: string
-        enum: [ SchemaNormalization ]
-      transform_config:
-        type: string
-        enum:
-          - None
-          - Default
+    type: string
+    enum:
+      - None
+      - Default
+    examples:
+      - None
+      - Default
   CustomSchemaNormalization:
     title: Schema Normalization
     description: Responsible for normalization according to the schema.
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -669,7 +669,7 @@ definitions:
         additionalProperties: true
   CustomSchemaNormalization:
     title: Custom Schema Normalization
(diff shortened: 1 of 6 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1999,8 +2009,10 @@ def create_record_selector(
                 else None,
                 **client_side_incremental_sync,
             )
-        schema_normalization = TypeTransformer(
-            SCHEMA_TRANSFORMER_TYPE_MAPPING[model.schema_normalization]
+        schema_normalization = (
+            TypeTransformer(SCHEMA_TRANSFORMER_TYPE_MAPPING[model.schema_normalization])
+            if isinstance(model.schema_normalization, SchemaNormalizationModel)
+            else self._create_component_from_model(model.schema_normalization, config=config)  # type: ignore[arg-type] # custom normalization model expected here
```
brianjlai, line 2015: Presumably this `CustomNormalization` component would need to already adhere or inherit from the `TypeTransformer` interface/class?

And that's why we don't need to invoke the `TypeTransformer` constructor?

brianjlai (commented):
a few nits and one question, otherwise seems pretty straightforward!

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2570,6 +2572,28 @@ definitions:
     examples:
       - None
       - Default
+  CustomSchemaNormalization:
+    title: Custom Schema Normalization
+    description: Use this to implement custom normalization according to the schema.
```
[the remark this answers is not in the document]
artem1205: updated

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1513,7 +1529,9 @@ class RecordSelector(BaseModel):
         description="Responsible for filtering records to be emitted by the Source.",
         title="Record Filter",
     )
-    schema_normalization: Optional[SchemaNormalization] = SchemaNormalization.None_
+    schema_normalization: Optional[Union[SchemaNormalization, CustomSchemaNormalization]] = (
+        SchemaNormalization.None_
```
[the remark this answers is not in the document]
artem1205: agreed, refactored.
we have toggle button in builder (`true` parsed as "Default"), it should still work after changes

image

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
artem1205: yes, most of the time we use `customtypetransformer` by _registertration_ some custom function, but generally speaking we need only `transform` interface, I can add abstract class for this

https://github.com/airbytehq/airbyte-python-cdk/blob/3984559c436fd7db2f73ac1f999518de355d282b/airbyte_cdk/sources/declarative/extractors/record_selector.py#L110-L116

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2570,6 +2572,28 @@ definitions:
     examples:
       - None
       - Default
+  CustomSchemaNormalization:
```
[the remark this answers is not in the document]
artem1205: 👍 reordered

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: Yeah I think we need to create an interface in the declarative package for this. I don't think we want users to be re-implementing `TypeTransformer` for a couple reasons:

- It looks like it has a lot of method we don't actually care about
- We want to avoid the pattern of customers re-implementing concrete classes since it couples our internal implementation to customer usage w/ unexpected breaking changes

So we should just create a new TypeTransformer interface in low-code (i'll leave naming to you), and then I think `RecordSelector.schema_normalization` will need to be changed to a union of these two classes interfaces. Once we have these added will 👍 the PR. thanks for making the other adjustments!

artem1205: added [AbstractTypeTransformer](https://github.com/airbytehq/airbyte-python-cdk/pull/194/commits/f787b6d126b709c8c6432106440c8a07447de482)

artem1205: hey @brianjlai , could you please approve the PR?

airbyte_cdk/sources/declarative/extractors/type_transformer.py
```diff
@@ -0,0 +1,55 @@
...
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+#
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Dict, Mapping
+
+
+@dataclass
+class AbstractTypeTransformer(ABC):
```
brianjlai, line 11: Would we be able to rename this to `TypeTransformer`?

And in `record_selector.py`, we alias it to `DeclarativeTypeTransformer`, I think it would be a tiny developer experience improvement to not have this look like our other custom classes which aren't prefixed  by Abstract and the idea of Python abstract class interfaces.

```
from airbyte_cdk.sources.declarative.extractors.type_transformer import TypeTransformer as DeclarativeTypeTransformer
```

I guess the drawback is a user implementing the wrong interface, but how does this sound to you?

brianjlai (approved):
sorry about the late approval and review. just a few small things but nothing to block on

airbyte_cdk/sources/declarative/extractors/record_selector.py
```diff
@@ -38,7 +34,7 @@ class RecordSelector(HttpSelector):
     extractor: RecordExtractor
     config: Config
     parameters: InitVar[Mapping[str, Any]]
-    schema_normalization: TypeTransformer
+    schema_normalization: TypeTransformer | AbstractTypeTransformer
```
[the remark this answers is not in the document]
artem1205: refactored,
fyi, it is just another way to implement `Union` : https://peps.python.org/pep-0604

airbyte_cdk/sources/declarative/extractors/type_transformer.py
artem1205: I'm ok with this since we use direct import inside our CDK and expose only `TypeTransformer` from utils.

airbyte_cdk/sources/declarative/extractors/record_selector.py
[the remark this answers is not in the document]
brianjlai: yep, roughly the same thing i looked at and tested, but just more for consistency as opposed to functionally working

PR #198, opened 2025-01-06
fix(airbyte-cdk): unable to create custom retriever

aldogonzalez8:
## Summary by CodeRabbit

- **New Features**
  - Enhanced method signatures in the ModelToComponentFactory to provide more flexible parameter handling.
  - Added support for custom retriever components in the declarative stream framework.

- **Improvements**
  - Made the decoder parameter optional with a default value in `create_http_requester`.
  - Made the transformations parameter optional in `create_record_selector`.

## What?

I was encountering errors like the following when creating a Custom Retriever:

```
ValueError: Error creating component 'RecordSelector' with parent custom component source_google_sheets.components.SheetsDataRetriever: Please provide RecordSelector.$parameters.transformations
````
From what I can see,  decoder can be optional [here](https://github.com/airbytehq/airbyte-python-cdk/blob/f8cb6598fffbb9b2d9c270c4f5cc974bd9012e01/airbyte_cdk/sources/declarative/models/declarative_component_schema.py#L1954) can is default to JsonDecoder [here](https://github.com/airbytehq/airbyte-python-cdk/blob/f8cb6598fffbb9b2d9c270c4f5cc974bd9012e01/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L2069-L2073).

For transformations also we set it as optional [here](https://github.com/airbytehq/airbyte-python-cdk/blob/f8cb6598fffbb9b2d9c270c4f5cc974bd9012e01/airbyte_cdk/sources/declarative/models/declarative_component_schema.py#L1695-L1705).

### CODE REVIEW

File: airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py [hunk 1]
```diff
@@ -1976,7 +1981,7 @@ def create_record_selector(
         config: Config,
         *,
         name: str,
-        transformations: List[RecordTransformation],
+        transformations: Optional[List[RecordTransformation]] = None,
```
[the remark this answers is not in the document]
aldogonzalez8: > Since we run on 3.10, the latter would now work fine, right?

Yes, that's reasonable.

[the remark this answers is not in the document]
natikgadzhi: I expect that at some point Ruff linters will do this for us in bulk

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1560,7 +1560,12 @@ def create_exponential_backoff_strategy(
         )
 
     def create_http_requester(
-        self, model: HttpRequesterModel, decoder: Decoder, config: Config, *, name: str
+        self,
+        model: HttpRequesterModel,
+        config: Config,
+        decoder: Decoder = JsonDecoder(parameters={}),
```
natikgadzhi, line 1566: Help me with my shitty Python knowledge. The signature change does two things here: 
1. Add default value for `decoder` — this is additive and will fix situations where decoder was not passed in
2. The order of named arguments changed. Is this important, or did we always pass them named and it's compatible?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1976,9 +1981,9 @@ def create_record_selector(
         config: Config,
         *,
         name: str,
-        transformations: List[RecordTransformation],
-        decoder: Optional[Decoder] = None,
-        client_side_incremental_sync: Optional[Dict[str, Any]] = None,
+        transformations: List[RecordTransformation] | None = None,
+        decoder: Decoder | None = None,
```
natikgadzhi, line 1985: Intentional that it defaults to None here, but JSONDecoder upstairs?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2008,7 +2013,7 @@ def create_record_selector(
             name=name,
             config=config,
             record_filter=record_filter,
-            transformations=transformations,
+            transformations=transformations or [],
```
natikgadzhi, line 2016: Non-blocking: Is it intentional that transformations default to `None` on argument list of the method, but to `[]` here? Should we just say `transormations: List[RecordTransformation] = []` in the args list, or is it parsed to `None` explicitly?

natikgadzhi (approved):
If all other tests are green, I want to unblock this! Asked a bunch of non-blocking questions, but they are, again, non-blocking.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: Lint was very annoyed about using a mutable object as the default, and I thought it was safer to pass [] as the default factory does [here](https://github.com/airbytehq/airbyte-python-cdk/blob/f8cb6598fffbb9b2d9c270c4f5cc974bd9012e01/airbyte_cdk/sources/declarative/extractors/record_selector.py#L45) for RecordSelector.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: Interesting point. Default To None seems to work fine in a quick test, so maybe JsonDecoder(parameters={}) is unnecessary. Let me dig in a little bit.

aldogonzalez8: Oh yes, now I remember. The HttpRequester expects a decoder.

`airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py:1619: error: Argument "decoder" to "HttpRequester" has incompatible type "Decoder | None"; expected "Decoder"  [arg-type]`

But we could create one if None for consistency with the arguments I other methods.

aldogonzalez8: In a second thought, it would be a little bit redundant, so I think I will put it back as it was so:

> Intentional that it defaults to None here, but JSONDecoder upstairs?

Yes, The HttpRequester expects a decoder and mypy yells for that.

maxi297 (approved):
[source-linkedinads has a custom retriever](https://github.com/airbytehq/airbyte/blob/56f6dc38ac2288b8ddf75eba8e840bdd68b4b7d7/airbyte-integrations/connectors/source-linkedin-ads/source_linkedin_ads/components.py#L212). Was it broken? Is there anything to do with this one?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: > Add default value for decoder — this is additive and will fix situations where decoder was not passed in

Yes

> The order of named arguments changed. Is this important, or did we always pass them named and it's compatible?

Usually (and from what I could find in CDK and airbyte base code), this method is only called by `_create_component_from_model`, which passes the parameters as keyword args.

image

image

image

aldogonzalez8:
@maxi297 Also doing a dummy experiment if I try to create a CustomRetriever with an HttpRequester in linkedin-ads I see the error I observed when working with google-sheets

image

```
            except TypeError as error:
                missing_parameters = self._extract_missing_parameters(error)
                if missing_parameters:
                    raise ValueError(
                        f"Error creating component '{type_name}' with parent custom component {model.class_name}: Please provide "
                        + ", ".join((f"{type_name}.$parameters.{parameter}" for parameter in missing_parameters))
                    )
>               raise TypeError(f"Error creating component '{type_name}' with parent custom component {model.class_name}: {error}")
E               TypeError: Error creating component 'HttpRequester' with parent custom component source_linkedin_ads.components.LinkedInAdsCustomRetriever: ModelToComponentFactory.create_http_requester() missing 1 required positional argument: 'decoder'
```

PR #202, opened 2025-01-06
fix(low-code cdk pagination): Fix the offset strategy so that it resets back to 0 when a stream is an incremental data feed

brianjlai:
Fixes https://github.com/airbytehq/oncall/issues/7210

## What

We were observing this error for low-code connectors using the offset pagination strategy:

```
  File "/home/airbyte/.pyenv/versions/3.10.15/lib/python3.10/site-packages/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py", line 87, in reset
    raise ValueError(
ValueError: Reset value None for OffsetIncrement pagination strategy was not an integer
```

## How

After investigating this a bit, I realized that this was only happening when we used an incremental stream where `is_data_feed` was enabled. 

And that is because the `StopConditionPaginationStrategyDecorator` that gets instantiated for data feeds is not implemented correctly for reseting back to 0. We accept optional values, but the problem is that the decorator can still accidentally pass in `None` when the surrounding `reset()` passes None. Instead we need to call `reset()` with no incoming parameter. I also fixed this in the `OffsetPaginationStrategy` as well just to cover all our bases.

## Summary by CodeRabbit

- **New Features**
	- Enhanced pagination strategies with more flexible reset behaviors
	- Added support for stop page conditions in data feed pagination

- **Tests**
	- Introduced new test for data feed paginator with stop page condition
	- Simplified offset increment reset test logic

- **Bug Fixes**
	- Improved handling of reset values in pagination strategies
	- Added support for `None` reset values in offset increment and stop condition strategies

### DIFF
--- a/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
+++ b/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
@@ -83,7 +83,9 @@ class OffsetIncrement(PaginationStrategy):
             return self._offset
 
     def reset(self, reset_value: Optional[Any] = 0) -> None:
-        if not isinstance(reset_value, int):
+        if reset_value is None:
+            self._offset = 0
+        elif not isinstance(reset_value, int):
             raise ValueError(
                 f"Reset value {reset_value} for OffsetIncrement pagination strategy was not an integer"
             )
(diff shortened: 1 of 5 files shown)


natikgadzhi (approved):
Clean work. Thank you for the fix @brianjlai

PR #206, opened 2025-01-08
feat(low-code): added flatten_lists option to FlattenFields transformation

darynaishchenko:
#### Problem
`FlattenFields` transformation updates array fields by default
 from `{"array": [1, 2, 3]}` 
  to `{"array.1": 1, "array.2": 2, "array.3": 3}`
  (for more examples see `unit_tests/sources/declarative/transformations/test_flatten_fields.py`)
  
  Some connectors (e.g `airtable`) don't need such transformation of arrays, we just need to keep it as is.
  
  #### Solution
  Added `flatten_lists` to `FlattenFields`(default is True) to be able to set up needed behavior of the transformation. 

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional `flatten_lists` configuration to control list flattening behavior during data transformations.
	- Default behavior remains unchanged (lists will be flattened).
	- Users can now choose to preserve original list structures if needed.

- **Tests**
	- Enhanced test cases to verify the new `flatten_lists` configuration option, including scenarios for both `True` and `False` settings.

### DIFF
--- a/airbyte_cdk/sources/declarative/transformations/flatten_fields.py
+++ b/airbyte_cdk/sources/declarative/transformations/flatten_fields.py
@@ -11,6 +11,8 @@ from airbyte_cdk.sources.types import Config, StreamSlice, StreamState
 
 @dataclass
 class FlattenFields(RecordTransformation):
+    flatten_lists: bool = True
+
     def transform(
         self,
         record: Dict[str, Any],
@@ -39,7 +41,7 @@ class FlattenFields(RecordTransformation):
                     )
                     stack.append((value, new_key))
 
-            elif isinstance(current_record, list):
+            elif isinstance(current_record, list) and self.flatten_lists:
                 for i, item in enumerate(current_record):
                     force_with_parent_name = True
                     stack.append((item, f"{parent_key}.{i}"))
--- a/unit_tests/sources/declarative/transformations/test_flatten_fields.py
+++ b/unit_tests/sources/declarative/transformations/test_flatten_fields.py
@@ -8,14 +8,27 @@ from airbyte_cdk.sources.declarative.transformations.flatten_fields import (
     FlattenFields,
 )
 
+_FLATTEN_LISTS = True
+_DO_NOT_FLATTEN_LISTS = False
+
 
 @pytest.mark.parametrize(
     "flatten_lists, input_record, expected_output",
     [
-        (True, {"FirstName": "John", "LastName": "Doe"}, {"FirstName": "John", "LastName": "Doe"}),
-        (True, {"123Number": 123, "456Another123": 456}, {"123Number": 123, "456Another123": 456}),
-        (
-            True,
+        pytest.param(
+            _FLATTEN_LISTS,
+            {"FirstName": "John", "LastName": "Doe"},
+            {"FirstName": "John", "LastName": "Doe"},
+            id="flatten simple record with string values",
+        ),
+        pytest.param(
+            _FLATTEN_LISTS,
+            {"123Number": 123, "456Another123": 456},
+            {"123Number": 123, "456Another123": 456},
+            id="flatten simple record with int values",
+        ),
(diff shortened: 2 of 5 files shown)


unit_tests/sources/declarative/transformations/test_flatten_fields.py
```diff
@@ -10,11 +10,12 @@
 
 
 @pytest.mark.parametrize(
-    "input_record, expected_output",
+    "flatten_lists, input_record, expected_output",
     [
-        ({"FirstName": "John", "LastName": "Doe"}, {"FirstName": "John", "LastName": "Doe"}),
-        ({"123Number": 123, "456Another123": 456}, {"123Number": 123, "456Another123": 456}),
+        (True, {"FirstName": "John", "LastName": "Doe"}, {"FirstName": "John", "LastName": "Doe"}),
```
maxi297, line 15: I'm not really attached to this but I'm wondering what is your opinion on this. It is becoming harder and harder when we have a lot of cases to relate to which parameter is what. Do you think it would be overkill to have `_FLATTEN_LISTS = True` and `_DO_NOT_FLATTEN_LISTS = False` and use them here for extra clarity? Or who it be better to have ids for each test cases? Or should we do both?

airbyte_cdk/sources/declarative/transformations/flatten_fields.py
```diff
@@ -11,6 +11,8 @@
 
 @dataclass
 class FlattenFields(RecordTransformation):
+    flatten_lists: bool = True
```
maxi297, line 14: Should we make this a private fields by prefixing this by `_`?

maxi297 (approved):
I really like the exhaustive test cases. I'll approve even regardless of some nit comments

unit_tests/sources/declarative/transformations/test_flatten_fields.py
darynaishchenko: good point, added ids and constants for flatten_fields

airbyte_cdk/sources/declarative/transformations/flatten_fields.py
darynaishchenko: for me it can be public as users should have access to this field to configure the behavior of the transformation. I don't think transforming arrays as described in the pr description is a common case, I assume most of connectors will need to keep lists as is.

maxi297: I'm not attached a lot to changing this so we can leave it as-is. I just want to explain why I think this is not public with this hypothetical situation: We have a user that complains about a bug saying that they set `flatten_lists == False` and therefore they expect sets to be flattened but not lists. This seems logical as the `flatten_lists` parameter says that it flatten lists but says nothing about sets. Given this, we decide to keep the same behavior but just be explicit that it flattens any collection (not just lists) by renaming the variable `flatten_collections`. This is now a breaking change even though the behavior has not changed. This seems to indicate that we are leaking how the class work because we had to do a breaking change even though nothing has changed.

Now, there is one precision to that which is that in python, any argument can be reference by kwargs. For example:
```
def toto(a):
  print(a)
  
toto("to be printed")  # works fine
toto(a="to be printed")  # works fine
```
This is a bit annoying in Python because the name of your parameters are part of the API by default. [PEP 570](https://peps.python.org/pep-0570/) tried to fix that (the [Benefits of Positional-Only Parameters](https://peps.python.org/pep-0570/#benefits-of-positional-only-parameters) sections mentions the benefits in terms of backward compatibility). But this means that in our case, changing `flatten_lists` to `flatten_collections` in the `__init__` (or equivalent for dataclasses) would still be breaking but this is the "pleasures" of working with Python and in our case, it would still limit the spread of the breaking change. 

TLDR: If we keep the same behavior, we should be able to do many changes without this being a breaking change

PR #208, opened 2025-01-08
ci: use binary of airbyte-ci instead of dev version that requires python10

aaronsteers:
## Summary by CodeRabbit

- **Chores**
  - Updated GitHub Actions workflow configuration for Airbyte CI tools installation
  - Modified command for installing CI tools from development to binary method
AirbyteLogFormatter class


aaronsteers:
It looks like this is working now. Once I see the tests are passed, I'll run `/format-fix` to undo my dummy change and fix the lint/format failure. Then this will be ready to go.

Issue #197, opened 2025-01-04
Shopify acceptance tests fail and block the PRs


rpopov:
Running the CAT tests locally on:
* Oracle Linux 9.5
* Oracle Linux 8.7
* Fedora Workstation 41

**Observations**
* Running the CDK acceptance tests locally with:
```
cd airbyte/airbyte-integrations/bases/connector-acceptance-test
poetry run pytest -p connector_acceptance_test.plugin --acceptance-test-config=../../connectors/source-shopify --timeout=300
```
fails with the container (Docker image: **registry.dagger.io/engine   v0.15.1**) errors:
```
# expect more open files due to per-client SQLite databases
# many systems default to 1024 which is far too low
ulimit -n 1048576 || echo "cannot increase open FDs with ulimit, ignoring"
... ... ...
buildkitd: failed to create engine: failed to create network providers: CNI setup error: plugin type="bridge" failed (add): failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 3: iptables v1.8.10 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
iptables v1.8.10 (legacy): can't initialize iptables table `filter': Table does not exist (do you need to insmod?)
Perhaps iptables or your kernel needs to be upgraded.
```
* Namely:
  * The internally executed `/usr/sbin/iptables -t nat -S --wait` command fails with **exit code 3** which does not allow the container to initialize and terminates it.
  * `man iptables` states: "Errors which indicate an incompatibility between kernel and user space cause an exit code of 3".
  * `ulimit -n 1048576` used to fail when used podman instead of docker.

* The same failure seems to happen in the pipelines of the Pull requests. Examples:
  * https://github.com/airbytehq/airbyte-python-cdk/pull/195
  * https://github.com/airbytehq/airbyte-python-cdk/pull/196
  * https://github.com/airbytehq/airbyte-python-cdk/pull/193 
* The logs report the same:
```
...
=========================== short test summary info ============================
SKIPPED [1] ../app/connector_acceptance_test/plugin.py:65: Skipping TestConnectorAttributes.test_streams_define_primary_key: not found in the config.
SKIPPED [1] ../app/connector_acceptance_test/plugin.py:65: Skipping TestConnectorDocumentation.test_prerequisites_content: not found in the config.
SKIPPED [1] ../app/connector_acceptance_test/tests/test_core.py:142: Config is not provided
[...]

PR #214, opened 2025-01-13
In each record to filter and transform, publish a local service field holding the original object the record is extracted from

rpopov:
Implement https://github.com/airbytehq/airbyte/issues/50395 discussed in https://github.com/airbytehq/airbyte/discussions/49971

At the record extraction step, in each record, add the service field $root holding a reference to:
* the root response object when parsing JSON format
* the original record, when parsing JSONL format 
that each record to process is extracted from.
**More service fields could be added in the future.**
The service fields are available in the record's filtering and transform steps.  

Avoid:
* reusing the maps/dictionaries produced, thus avoid building cyclic structures
* transforming the service fields in the Flatten transformation.

Explicitly clean the service field(s) after the transform step, thus making them:
* local for the filter and transform steps
* not visible to the next mapping and store steps (as they should be)
* not visible in the tests beyond the test_record_selector (as they should be)
This allows the record transformation logic to define its "local variables" to reuse
some interim calculations.

The contract of body parsing seems irregular in representing the cases of bad JSON, no JSON and empty JSON. 
It cannot be unified as that irregularity is already used.    

Update the development environment setup documentation
* to organize and present the setup steps explicitly
* to avoid misunderstandings and wasted efforts.

Update CONTRIBUTING.md to 
* collect and organize the knowledge on running the test locally;
* state the actual testing steps;
* clarify and make explicit the procedures and steps.

The unit, integration, and acceptance tests in this exact version succeed under Fedora 41, while
one of them fails under Oracle Linux 8.7. not related to the contents of this PR.
The integration tests of the CDK fail due to missing `secrets/config.json` file for the Shopify source.
See https://github.com/airbytehq/airbyte-python-cdk/issues/197


rpopov:
Still https://github.com/airbytehq/airbyte-python-cdk/issues/197 is a problem. It blocks the build.

PR #220, opened 2025-01-15
fix: (DeclarativeOAuthFlow) - add the ability to override the `refresh_token_request_body` key names, updated `oauth_connector_input_specification` tooltips

bazarnov:
## What
During the QA sessions for `DeclarativeOAuthFlow` in Builder, the following issues were discovered:
- need a way to override the keys inside the `refresh_token_request_body` to provide the custom names as keys to add more flexibility and cover more custom OAuth flows, which have non-default `request_body` to be provided.
- made the `oauth_connector_input_specification.extract_output` field `Optional`, since it's going to be deprecated soon.
- fix the `oauth_connector_input_specification` (DeclarativeOAuthFlow) tooltips to reflect the updated `Jinja2` variables declaration style. Resolving: https://github.com/airbytehq/airbyte-internal-issues/issues/11337

## Summary by CodeRabbit

- **New Features**
	- Enhanced OAuth2 authentication configuration
	- Added support for custom key names in OAuth token requests
	- Improved flexibility for specifying client ID, client secret, refresh token, and grant type names

- **Improvements**
	- Updated OAuth authenticator to allow more dynamic parameter naming
	- Expanded configuration options for OAuth token refresh process

### DIFF
--- a/airbyte_cdk/sources/declarative/auth/oauth.py
+++ b/airbyte_cdk/sources/declarative/auth/oauth.py
@@ -56,8 +56,12 @@ class DeclarativeOauth2Authenticator(AbstractOauth2Authenticator, DeclarativeAut
     token_expiry_is_time_of_expiration: bool = False
     access_token_name: Union[InterpolatedString, str] = "access_token"
     access_token_value: Optional[Union[InterpolatedString, str]] = None
+    client_id_name: Union[InterpolatedString, str] = "client_id"
+    client_secret_name: Union[InterpolatedString, str] = "client_secret"
     expires_in_name: Union[InterpolatedString, str] = "expires_in"
+    refresh_token_name: Union[InterpolatedString, str] = "refresh_token"
     refresh_request_body: Optional[Mapping[str, Any]] = None
+    grant_type_name: Union[InterpolatedString, str] = "grant_type"
     grant_type: Union[InterpolatedString, str] = "refresh_token"
     message_repository: MessageRepository = NoopMessageRepository()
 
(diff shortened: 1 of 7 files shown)


maxi297 (approved):
:shipit:

airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -122,18 +136,27 @@ def get_token_refresh_endpoint(self) -> Optional[str]:
...
+    def get_client_secret_name(self) -> str:
+        return self._client_secret_name.eval(self.config)  # type: ignore # eval returns a string in this context
+
     def get_client_secret(self) -> str:
         client_secret: str = self._client_secret.eval(self.config)
         if not client_secret:
             raise ValueError("OAuthAuthenticator was unable to evaluate client_secret parameter")
         return client_secret
 
+    def get_refresh_token_name(self) -> str:
```
bnchrch, line 157: 💅 Should these be `_key` instead?
```suggestion
    def get_refresh_token_key(self) -> str:
```

bnchrch (commented):
one question

airbyte_cdk/sources/declarative/auth/oauth.py
bazarnov: The naming was taken out from the existing methods, like these ones:
https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/auth/oauth.py#L143-L147

Should we change them all in this case? I'd have it as `_name` still to have consistency, but I don't mind having `_key` if this is critical.

bnchrch: Fair enough! didnt see that was already the convention in the cdk.

bnchrch (approved):
LGTM! TY!

a-man-is-sitting-in-a-chair-and-talking-to-a-woman

PR #221, opened 2025-01-15
security fix: sanitize transformation error message to hide record values

ChristoGrab:
## What

A Builder user alerted us to a Record transformation error that was exposing sensitive email values in their test reads. This PR updates the error message constructor to sanitize values, only displaying the structure/type of the field causing the failure.

### DIFF
--- a/airbyte_cdk/sources/utils/transform.py
+++ b/airbyte_cdk/sources/utils/transform.py
@@ -225,16 +225,17 @@ class TypeTransformer:
             logger.warning(self.get_error_message(e))
 
     def get_error_message(self, e: ValidationError) -> str:
-
-        def _get_type_structure(input_data: Any) -> Any:
+        def get_type_structure(input_data: Any) -> Any:
             if isinstance(input_data, dict):
-                return {key: _get_type_structure(value) for key, value in input_data.items()}
+                return {
+                    key: get_type_structure(field_value) for key, field_value in input_data.items()
+                }
             elif isinstance(input_data, list):
-                return [_get_type_structure(value) for value in input_data] if input_data else "array"
+                return [get_type_structure(item) for item in input_data] if input_data else "array"
             else:
-                return type(input_data).__name__.lower()
-            
-        key_path = "." + ".".join(map(str, e.path))
-        type_structure = _get_type_structure(e.instance)
+                return python_to_json[type(input_data)]
 
-        return f"Failed to transform value of {type_structure} to '{e.validator_value}'. Key path: '{key_path}'"
+        field_path = ".".join(map(str, e.path))
+        type_structure = get_type_structure(e.instance)
+
+        return f"Failed to transform value of type '{type_structure}' to '{e.validator_value}' at path: '{field_path}'"
--- a/airbyte_cdk/sources/utils/transform.py
+++ b/airbyte_cdk/sources/utils/transform.py
@@ -225,17 +225,20 @@ class TypeTransformer:
             logger.warning(self.get_error_message(e))
 
     def get_error_message(self, e: ValidationError) -> str:
-        def get_type_structure(input_data: Any) -> Any:
+        def _get_type_structure(input_data: Any) -> Any:
             if isinstance(input_data, dict):
-                return {
-                    key: get_type_structure(field_value) for key, field_value in input_data.items()
-                }
+                structure = {key: _get_type_structure(field_value) for key, field_value in input_data.items()}
+                return f"object with structure {structure}"
             elif isinstance(input_data, list):
-                return [get_type_structure(item) for item in input_data] if input_data else "array"
+                if not input_data:
+                    # Handle empty array
+                    return "array"
+                items = [_get_type_structure(item) for item in input_data]
+                return f"array with structure {items}"
             else:
                 return python_to_json[type(input_data)]
 
         field_path = ".".join(map(str, e.path))
-        type_structure = get_type_structure(e.instance)
-
-        return f"Failed to transform value of type '{type_structure}' to '{e.validator_value}' at path: '{field_path}'"
+        type_structure = _get_type_structure(e.instance)
+        
+        return f"Failed to transform value from type '{type_structure}' to type '{e.validator_value}' at path: '{field_path}'"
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,6 +225,16 @@ def transform(
...
-        key_path = "." + ".".join(map(str, e.path))
-        return f"Failed to transform value {repr(e.instance)} of type '{instance_json_type}' to '{e.validator_value}', key path: '{key_path}'"
+        def _get_type_structure(input_data: Any) -> Any:
+            if isinstance(input_data, dict):
+                structure = {
+                    key: _get_type_structure(field_value) for key, field_value in input_data.items()
+                }
+                return f"object with structure {structure}"
+            else:
+                return python_to_json[type(input_data)]
```
bnchrch, line 235: ❓ What happens if `input_data` is `None`?

airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,6 +225,16 @@ def transform(
             logger.warning(self.get_error_message(e))
 
     def get_error_message(self, e: ValidationError) -> str:
-        instance_json_type = python_to_json[type(e.instance)]
-        key_path = "." + ".".join(map(str, e.path))
-        return f"Failed to transform value {repr(e.instance)} of type '{instance_json_type}' to '{e.validator_value}', key path: '{key_path}'"
+        def _get_type_structure(input_data: Any) -> Any:
+            if isinstance(input_data, dict):
+                structure = {
+                    key: _get_type_structure(field_value) for key, field_value in input_data.items()
```
bnchrch, line 231: ❓Were doing recursive calls. What happens if theres 10000 levels? 
❗ We need a max recursion stop. e.g. dont let it go further than 3 levels, and if you do just stub it with "object" or something like that

airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,6 +225,16 @@ def transform(
             logger.warning(self.get_error_message(e))
 
     def get_error_message(self, e: ValidationError) -> str:
-        instance_json_type = python_to_json[type(e.instance)]
-        key_path = "." + ".".join(map(str, e.path))
-        return f"Failed to transform value {repr(e.instance)} of type '{instance_json_type}' to '{e.validator_value}', key path: '{key_path}'"
+        def _get_type_structure(input_data: Any) -> Any:
```
bnchrch, line 228: 💅 Im rarely a fan of nested functions. To much closure wrangling and mutation risk. Is there a reason we did this here? if not lets move it to its own level

airbyte_cdk/sources/utils/transform.py
ChristoGrab: Good callout, added explicit handling for `null` values and a unit test for it

airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,16 +226,41 @@ def transform(
...
+        # Handle null values
+        if input_data is None:
+            return "null"
+
+        # Avoid recursing too deep
+        if current_depth >= MAX_NESTING_DEPTH:
+            return (
+                "object"
+                if isinstance(input_data, dict)
+                else "array"
```
bnchrch, line 250: ❓ Why do we handle the array case explicitly if the recursion depth is hit? but not elsewhere?

airbyte_cdk/sources/utils/transform.py
ChristoGrab: Heard. Converted `_get_type_structure` from a nested function to a top-level instance method

airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,6 +226,41 @@ def transform(
...
+
+        if isinstance(input_data, dict):
+            structure = {
+                key: self._get_type_structure(field_value, current_depth + 1)
+                for key, field_value in input_data.items()
+            }
+
+            if current_depth == 0:
+                return f"object with structure {structure}"
+            return structure
```
bnchrch, line 263: ❓ is the "object with structure" message nessesary?  without it the code is cleaner

```suggestion
        if isinstance(input_data, dict):
            return {
                key: self._get_type_structure(field_value, current_depth + 1)
                for key, field_value in input_data.items()
            }
```

airbyte_cdk/sources/utils/transform.py
ChristoGrab: Makes sense! Added a constant for the max depth. Also helped me catch that I was wrapping each recursive object layer in the "object with structure" prefix substring, so we now only do that at the top depth_level

ChristoGrab: Whoops, nvm, just saw the new comment haha

ChristoGrab: Committed the suggestion to clarify the error message and updated the tests

airbyte_cdk/sources/utils/transform.py
ChristoGrab: This was my bad - I had originally included explicit handling of arrays in the nested method, but removed it and should not have put it back in after de-nesting the sturcture method.

My understanding is that when an array is encountered and the schema defines it (with "type": "array" and an "items" property), the normalizer automatically iterates over each element and applies the necessary conversion. We only need to transform the individual elements, and the array itself doesn't require a separate conversion step. Essentially, there’s no scenario where the array itself would fail to be converted, since the conversion logic is applied to its elements. Does that track?

airbyte_cdk/sources/utils/transform.py
```diff
@@ -225,6 +226,37 @@ def transform(
...
+            )
+
+        if isinstance(input_data, dict):
+            return {
+                key: self._get_type_structure(field_value, current_depth + 1)
+                for key, field_value in input_data.items()
+            }
+
+        else:
+            return python_to_json[type(input_data)]
```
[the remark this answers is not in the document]
ChristoGrab: @coderabbitai I don't think we need to do so here, as the data being transformed comes from API responses or source systems, so we expect to only deal with standard JSON-serializable types. The `python_to_json` mapping should already cover all the standard types that we expect to encounter.

If a custom type did make it to the transformer, it's probably best to fail fast on a KeyError anyway given our expectations.

bnchrch (approved):
Well sanitized
a-man-is-spraying-something-in-a-kitchen-with-the-words-box-video-downloader-behind-him

PR #224, opened 2025-01-16
feat(low-code): added condition to TypesMap of DynamicSchemaLoader

darynaishchenko:
### What
Dynamic schema loader: there can be a case when type of the field doesn't depend on "type" field value only. Example: `source airtable`.
field_1
```
{
"type": "formula"
"result": {"type": "number"}
}
```
field_2
```
{
"type": "formula"
"result": {"type": "singleLineText"}
}
```
### How
To properly identify field type in cases like above `condition` option was added to `TypesMap` object. Now providing TypesMap object condition we can create 2 separate cases for string values and number values.
See `unit_tests/sources/declarative/schema/test_dynamic_schema_loader.py` for more examples.

## Summary by CodeRabbit

- **New Features**
	- Introduced a `condition` property to support more flexible type mapping and record filtering in declarative components.
	- Enhanced schema loader to support conditional type validation and replacement.

- **Improvements**
	- Expanded type mapping capabilities to allow string conditions.
	- Improved dynamic schema generation with context-aware type validation.

- **Tests**
	- Added a new test case to validate the dynamic schema loader with type conditions.


lazebnyi (approved):
lgtm

PR #225, opened 2025-01-17
fix: handle backoff_strategies in CompositeErrorHandler

ChristoGrab:
## What

Although users can define custom backoff logic in the manifest, the CompositeErrorHandler component does not currently access the backoff_strategies, meaning it always defaults to the default exponential behavior. By accessing the `backoff_strategies` in the child `ErrorHandler` components, we can handle user-specified strategies.

Resolves: https://github.com/airbytehq/airbyte/issues/42928

## Cautionary Note

My understanding of the ErrorHandler flow is that it is not currently configured to respect multiple backoff strategies for a single stream, and will always use the first non-`None` value as the strategy for both:

 1. matches in `response_filters` 
 2. default retryable errors.
 
 For example, using:
 
 ```
 type: CompositeErrorHandler
error_handlers:
  - type: DefaultErrorHandler
    max_retries: 2
    backoff_strategies:
      - type: ConstantBackoffStrategy
        backoff_time_in_seconds: 10
    response_filters:
      - type: HttpResponseFilter
        action: RETRY
        http_codes:
          - 404
  - type: DefaultErrorHandler
    response_filters:
      - type: HttpResponseFilter
        action: RETRY
        http_codes:
          - 500
    max_retries: 3
    backoff_strategies:
      - type: ExponentialBackoffStrategy
        factor: 0.5
 ```
 
 We would expect a `500` level response to utilize the Exponential strategy. However, the output seen is in fact:

``` 
Retrying. Sleeping for 10.0 seconds (rate_limiting.py:114)
Retrying. Sleeping for 10.0 seconds (rate_limiting.py:114)
```

This is a function of the [errorhandler logic flow and is a known behavior](https://github.com/airbytehq/airbyte-python-cdk/blob/b5ed82c0bd3dcce7dee4d4291291e1960a1b7b21/airbyte_cdk/sources/declarative/requesters/error_handlers/default_error_handler.py#L35), distinct from the bug this PR tries to address.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling with improved backoff strategy management in the Composite Error Handler.
	- Added ability to aggregate and prioritize backoff strategies across multiple error handlers.

- **Tests**
	- Added comprehensive test coverage for backoff strategy implementation.
	- Verified correct behavior of backoff strategies in HTTP request retry scenarios.


unit_tests/sources/declarative/requesters/error_handlers/test_composite_error_handler.py
```diff
@@ -272,3 +275,40 @@ def test_max_time_is_max_of_underlying_handlers(test_name, max_times, expected_m
...
+            [ConstantBackoffStrategy(5, {}, {}), ConstantBackoffStrategy(10, {}, {})],
+        ),
+        (
+            "test_some_handlers_without_strategies",
+            [[ConstantBackoffStrategy(5, {}, {})], None, [ConstantBackoffStrategy(10, {}, {})]],
+            [ConstantBackoffStrategy(5, {}, {}), ConstantBackoffStrategy(10, {}, {})],
+        ),
+    ],
+)
+def test_composite_error_handler_backoff_strategies(
```
maxi297, line 301: This test ensures that the right strategies are returned which is good. It think there is still a test missing to make sure that when we bootstrap from a manifest, the backoff strategies are actually passed/used by the HttpClient. WDYT?

maxi297 (commented):
Can we link the GitHub issue this change is related to?

unit_tests/sources/declarative/requesters/error_handlers/test_composite_error_handler.py
ChristoGrab: Yup, that makes sense to me as a closer indicator of integration with the declared manifest, will add!

airbyte_cdk/sources/declarative/requesters/error_handlers/composite_error_handler.py
```diff
@@ -77,3 +78,19 @@ def interpret_response(
...
+
+    @property
+    def backoff_strategies(self) -> Optional[List[BackoffStrategy]]:
+        """
+        Combines backoff strategies from all child error handlers into a single flattened list.
+        Note: The first non-None strategy in this list becomes the default strategy for ALL retryable errors,
+        including both user-defined response filters and errors that fall back to DEFAULT_ERROR_MAPPING.
+        The list structure is currently not used to map different strategies to different error conditions.
+
+        Returns None if no handlers have strategies defined, which will result in the default backoff strategy being used.
```
maxi297, line 90: This comments only makes sense in the context of the HttpRequester [here](https://github.com/airbytehq/airbyte-python-cdk/blob/83dba91eb8c463780c316b7d9f6d70c601065bc6/airbyte_cdk/sources/declarative/requesters/http_requester.py#L85-L88). Can we clarify that?

maxi297 (approved):
LGTM! Can we create an issue to improve on backoff strategies and fix the issue when there are multiple error handlers in the composite error handler?

ChristoGrab:
Created the issue to improve the handling of backoff strategies, linking here for additional visibility: https://github.com/airbytehq/airbyte-internal-issues/issues/11449

airbyte_cdk/sources/declarative/requesters/error_handlers/composite_error_handler.py
ChristoGrab: Yup, updated to be more explicit about the behavior's origin 👍

PR #226, opened 2025-01-17
feat(AsyncRetriever): Allow for streams using AsyncRetriever and DatetimeBasedCursor to perform checkpointing

brianjlai:
## What

Add the ability for streams that define an AsyncRetriever and a DatetimeBasedCursor to perform periodic checkpointing when the date time window of a stream slice is synced successfully.

This is needed to unblock `source-amazon-seller-partner` which supports incremental async jobs.

## How

It's worth noting that this PR goes a bit in the opposite direction where we want to deprecate the existing `DatetimeBasedCursor` because now its more closely integrated into the `AsyncRetriever` stream slicing mechanism. However, due to the short time frame to deliver Amazon Seller Partner, trying to inject the `ConcurrentCursor` into the low-code `AsyncJobPartitionRouter` felt like the more difficult path.

The tradeoff in the short term here is that our Async + Incremental streams will run synchronously (worth noting that this was already the existing behavior based on how we construct the `concurrent_declarative_source.py`) So we incur a little bit of tech debt in exchange for simplicity to implement

## Summary by CodeRabbit

- **New Features**
	- Enhanced async retriever functionality with improved cursor management
	- Added support for more flexible stream state handling in async data processing
	- Introduced new cursor-related methods for better stream slice management

- **Improvements**
	- Updated stream processing logic to handle different retriever types
	- Improved state tracking and record processing in async retrievers

- **Technical Enhancements**
	- Added new methods for cursor and record tracking
	- Expanded support for datetime-based cursors in partition routing

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_stream.py
+++ b/airbyte_cdk/sources/declarative/declarative_stream.py
@@ -1,6 +1,7 @@
 #
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
+
 import logging
 from dataclasses import InitVar, dataclass, field
 from typing import Any, Iterable, List, Mapping, MutableMapping, Optional, Union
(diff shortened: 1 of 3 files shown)


brianjlai:
closing in favor of https://github.com/airbytehq/airbyte-python-cdk/pull/228

PR #227, opened 2025-01-17
feat(low-code): add DpathFlattenFields

darynaishchenko:
### What
source-airtable need record transformations that moves everything inside "fields" field to the top level of records and keeps others top level fields.
We already have FlattenFields transformation, but it transforms also objects inside "fields", records in airtable should not follow this format. Everything inside "fields" should be kept as is.
### How.
Added `DpathFlattenFields`, which works with provided path and updates record with object found in the path, doesn't change fields inside the object.  

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `DpathFlattenFields` transformation for flattening nested fields in records.
	- Supports dynamic field path selection and optional deletion of original values.
	- Enhances data manipulation capabilities for declarative sources.

- **Tests**
	- Added comprehensive unit tests for the `DpathFlattenFields` transformation.
	- Verified functionality across various scenarios, including handling of wildcards and non-existing fields.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -1316,6 +1316,7 @@ definitions:
             - "$ref": "#/definitions/KeysToLower"
             - "$ref": "#/definitions/KeysToSnakeCase"
             - "$ref": "#/definitions/FlattenFields"
+            - "$ref": "#/definitions/DpathFlattenFields"
             - "$ref": "#/definitions/KeysReplace"
       state_migrations:
         title: State Migrations
@@ -1865,6 +1866,7 @@ definitions:
             - "$ref": "#/definitions/KeysToLower"
             - "$ref": "#/definitions/KeysToSnakeCase"
             - "$ref": "#/definitions/FlattenFields"
+            - "$ref": "#/definitions/DpathFlattenFields"
             - "$ref": "#/definitions/KeysReplace"
       schema_type_identifier:
         "$ref": "#/definitions/SchemaTypeIdentifier"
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1316,6 +1316,7 @@ definitions:
             - "$ref": "#/definitions/KeysToLower"
             - "$ref": "#/definitions/KeysToSnakeCase"
             - "$ref": "#/definitions/FlattenFields"
+            - "$ref": "#/definitions/DpathFlattenFields"
```
natikgadzhi, line 1319: It would be an interesting exercise to add a bunch of these transformations to Builder UI — it currently only supports AddField and RemoveField I think

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1969,6 +1971,33 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
+  DpathFlattenFields:
+    title: Dpath Flatten Fields
+    description: A transformation that flatten field values to the to top of the record.
```
natikgadzhi, line 1976: I _think_ these are the descriptions that show up on the docs site in YAML reference. It could be a good idea to add a larger full example block for how to define this.

natikgadzhi: This is an aspirational and non-blocking comment, do not spend time here yet ;) There are too many components without great docs, no need to polish this one yet

natikgadzhi (approved):
This looks good!

PR #228, opened 2025-01-17
TMP development to show how things could work with concurrent cursor

maxi297:


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -487,6 +488,7 @@ def __init__(
         self._message_repository = message_repository or InMemoryMessageRepository(
             self._evaluate_log_level(emit_connector_builder_messages)
         )
+        self._state_manager = state_manager
```
maxi297, line 491: My guess is that this makes sense here given that we want to instantiate cursors with the state eventually to avoid a `set_initial_state` method call

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1476,6 +1476,17 @@ def _merge_stream_slicers(
                     stream_cursor=cursor_component,
                 )
         elif model.incremental_sync:
+            if model.retriever.type == "AsyncRetriever":
+                if model.incremental_sync.type != "DatetimeBasedCursor":
+                    # TODO explain why it isn't supported
```
maxi297, line 1481: TODO: I'm not exactly sure why other types of cursors wouldn't be supported but I was only doing this for source-amazon-ads so I wanted to be more restrictive than not.

Note that Global/PerPartition cursors were not updated which we will need for source-amazon-ads

airbyte_cdk/sources/types.py
```diff
@@ -151,7 +153,7 @@ def __json_serializable__(self) -> Any:
         return self._stream_slice
 
     def __hash__(self) -> int:
-        return hash(orjson.dumps(self._stream_slice, option=orjson.OPT_SORT_KEYS))
+        return SliceHasher.hash("dummy_name", self._stream_slice)
```
maxi297, line 156: This had to be updated to support `AsyncPartition`. SliceHasher considers `__json_serializable__` but orjson does not. I figure we should have the same slicing logic everywhere and if we want to update this logic to orjson, just do it once in `SliceHasher`

TODO: I'm not sure why we require the name of the stream for the slice hasher.

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -322,6 +322,7 @@
                     "http_method": "GET",
                 },
             },
+            "incremental_sync": {"$ref": "#/definitions/incremental_cursor"},
```
maxi297, line 325: To ensure that async_retriever with incremental syncs are also concurrent

airbyte_cdk/sources/types.py
maxi297: I'm very confused about this one. The only place I see this failing is when we organize the most recent cursor values by partition [here](https://github.com/airbytehq/airbyte-python-cdk/blob/83dba91eb8c463780c316b7d9f6d70c601065bc6/airbyte_cdk/sources/streams/concurrent/cursor.py#L236). The SliceHasher was added as part of [this change](https://github.com/airbytehq/airbyte-python-cdk/pull/39/files#diff-16a209c104b2a4ed6dfdf4ee7a0bf1d7b90207b4ca0f1e7a78afbb71cb3e834eR15-R30). I see the only variable I see that requires the slice to be hashed (apart from the cursor) is [this one](https://github.com/airbytehq/airbyte-python-cdk/blob/83dba91eb8c463780c316b7d9f6d70c601065bc6/airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py#L53) but it is already organized by stream_name so I don't think it is that useful. We can probably remove the name from the slice hasher. Would you see any other cases @brianjlai ?

maxi297: I pushed some changes with a default value for stream_name during the hashing and I think we would be fine

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1476,6 +1476,20 @@ def _merge_stream_slicers(
...
                 )
         elif model.incremental_sync:
+            if model.retriever.type == "AsyncRetriever":
+                if model.incremental_sync.type != "DatetimeBasedCursor":
+                    # We are currently in a transition to the Concurrent CDK and AsyncRetriever can only work with the support or unordered slices (for example, when we trigger reports for January and February, the report in February can be completed first). Once we have support for custom concurrent cursor or have a new implementation available in the CDK, we can enable more cursors here.
+                    raise ValueError("AsyncRetriever with cursor other than DatetimeBasedCursor is not supported yet")
+                if model.retriever.partition_router:
+                    # Note that this development is also done in parallel to the per partition development which once merged we could support here by calling `create_concurrent_cursor_from_perpartition_cursor`
+                    raise ValueError("Per partition state is not supported yet for AsyncRetriever")
+                return self.create_concurrent_cursor_from_datetime_based_cursor(
```
brianjlai, line 1486: This is one part of the transition to `ConcurrentCursor` that feels a bit jank because we're passing back a `ConcurrentCursor` to persist to the various low-code components and that is no longer technically the same `StreamSlicer` (low-code) interface that was to be expected. So at least in the short term our typing is going to be incorrect.

I think I can live with this since we do ultimately want to drop these low-code interfaces, but I'll add a comment here and we'll need to override mypy if we agree that this is tolerable

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -15,7 +15,6 @@
     Dict,
     List,
     Mapping,
-    MutableMapping,
```
brianjlai, line 18: unused import

brianjlai (approved):
i added the various fixes noted which should have this in a good state to merge. Approving since the fixes I pushed were what I would have commented to address

aaronsteers:
@maxi297 - FYI, I'm updating from `main` to resolve an issue in the `Test Connectors` workflow: `ModuleNotFoundError: No module named 'pkg_resources'`

PR #234, opened 2025-01-20
feat(concurrent cursor): attempt at clamping datetime

maxi297:
## What

source-amazon-seller-partner has reports that require specific start and end time ([example](https://developer-docs.amazon.com/sp-api/docs/report-type-values-analytics#amazon-search-terms-report)). The constraints are:
* Day: based on [this](https://github.com/airbytehq/airbyte/blob/1f40c82e100b85573fc4b29280f2dcf89bae1842/airbyte-integrations/connectors/source-amazon-seller-partner/source_amazon_seller_partner/streams.py#L825-L828), the lower boundary would have to truncate the hours, minutes, seconds and microseconds
```
pendulum.now("utc").start_of("day")
DateTime(2025, 1, 21, 0, 0, 0, tzinfo=Timezone('UTC'))
pendulum.now("utc").end_of("day")
DateTime(2025, 1, 21, 23, 59, 59, 999999, tzinfo=Timezone('UTC'))
```
* Week: Weekday should be Sunday but in the current implementation, we made it configurable
* Month: Slices will start on the first day of the month

## How
Adding clamping in ConcurrentCursor

## Can this change be reverted?
Yes

## Summary by CodeRabbit

- **New Features**
	- Added advanced clamping strategies for date-based cursors.
	- Introduced time-based cursor control with day, week, and month granularity.
	- Enhanced concurrent stream slicing with flexible time interval management.

- **Improvements**
	- Expanded cursor configuration options for more precise data retrieval.
	- Implemented robust time-based data synchronization strategies.

- **Technical Updates**
	- Updated cursor and stream handling mechanisms.
	- Added new protocols and strategies for cursor value management.
	- Introduced comprehensive unit tests for clamping strategies and cursor behavior.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1043,6 +1053,38 @@ def create_concurrent_cursor_from_datetime_based_cursor(
...
 
+        clamping_strategy = NoClamping()
+        if datetime_based_cursor_model.clamping:
+            match datetime_based_cursor_model.clamping.target:
+                case Clamping.target.DAY:
+                    clamping_strategy = DayClampingStrategy()
+                    end_date_provider = ClampingEndProvider(
+                        DayClampingStrategy(is_ceiling=False),
+                        end_date_provider,
+                        granularity=timedelta(seconds=1),
```
maxi297, line 1064: Note that all granularity in [this implementation](https://github.com/airbytehq/airbyte/blob/1f40c82e100b85573fc4b29280f2dcf89bae1842/airbyte-integrations/connectors/source-amazon-seller-partner/source_amazon_seller_partner/streams.py#L823-L853) seems to have been `1 microseconds`. 

However, I think we should update this to take `datetime_based_cursor_model.granularity`. WDYT?

airbyte_cdk/sources/streams/concurrent/clamping.py
```diff
@@ -0,0 +1,95 @@
...
+            day=1,
+            hour=0,
+            minute=0,
+            second=0,
+            microsecond=0,
+        )
+
+
+class Weekday(Enum):
+    MONDAY = 0
```
maxi297, line 70: Should I add a comment to make it explicit that the int values match the datetime library ones?

brianjlai (approved):
approving. i responded to some of your questions and the next commit I push has what you have mentioned addressed

brianjlai:
@maxi297 

I pushed a commit that does the following:

- allow for interpolation of the config to get the clamping value: This has the drawback of not being able to use enums since the string could be anything. I did also notice we do this for JWT, but interpolation on JWT doesn't work :/
- use the `cursor_granularity` for the clamping end provider
- Add tests to `ModelToComponentFactory` to verify that we create the right clamping strategy when parsing the model
- There were some formatting and things like an duplicated test that coderabbit caught. fixed those too

If you want to give this one more look, otherwise I'll merge this in the morning so that we can rebase this into the concurrent cursor for async retriever branch and I'll continue retesting with amazon seller partner on that CDK version

airbyte_cdk/sources/streams/concurrent/clamping.py
brianjlai: see above docstring added

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: I say it's using microseconds because this is what the implementation of pendulum `x.end_of("day")` does

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1053,18 +1054,29 @@ def create_concurrent_cursor_from_datetime_based_cursor(
...
+            )
+            evaluated_target = target.eval(config=config)
+            match evaluated_target:
+                case "DAY":
                     clamping_strategy = DayClampingStrategy()
                     end_date_provider = ClampingEndProvider(
                         DayClampingStrategy(is_ceiling=False),
-                        end_date_provider,
-                        granularity=timedelta(seconds=1),
+                        end_date_provider,  # type: ignore  # Having issues w/ inspection for GapType and CursorValueType as shown in existing tests. Confirmed functionality is working in practice
```
maxi297, line 1072: Yeah, `CursorType` and `GapType` should probably have been templated of a `Generic` `ConcurrentCursor`. I'm not sure this would have solved the pain to work with MyPy because we still need a way to describe that the relationship between `CursorType` and `GapType`. It's kind of a mess...

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1074,15 +1086,19 @@ def create_concurrent_cursor_from_datetime_based_cursor(
                     clamping_strategy = WeekClampingStrategy(weekday)
                     end_date_provider = ClampingEndProvider(
                         WeekClampingStrategy(weekday, is_ceiling=False),
-                        end_date_provider,
-                        granularity=timedelta(days=1),
+                        end_date_provider,  # type: ignore  # Having issues w/ inspection for GapType and CursorValueType as shown in existing tests. Confirmed functionality is working in practice
+                        granularity=cursor_granularity or timedelta(days=1),
```
maxi297, line 1090: We don't have a case here but I would assume it should be only `cursor_granularity` here. The downside I see with fallbacking to `timedelta(days=1)` is that if we have a clamping WeekClampingStrategy(Weekday.Monday) and we are on Monday, the sync will be missing the last day if `cursor_granularity` is not defined

airbyte_cdk/sources/streams/concurrent/cursor_types.py
```diff
@@ -0,0 +1,32 @@
...
+    def __ge__(self: "CursorValueType", other: "CursorValueType") -> bool:
+        pass
+
+    @abstractmethod
+    def __add__(self: "CursorValueType", other: GapType) -> "CursorValueType":
+        pass
+
+    @abstractmethod
+    def __sub__(self: "CursorValueType", other: GapType) -> "CursorValueType":
+        pass
```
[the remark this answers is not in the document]
brianjlai: already like this

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: yeah, we've run into this a few times where whatever type we pass in even though it works, the signature doesn't match. We probably should revisit, but for now i think we'll just have to punt on this

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -8,7 +8,6 @@
 import importlib
 import inspect
 import re
-import sys
```
brianjlai, line 11: this was an unused import. getting rid of

PR #236, opened 2025-01-21
feat(low-code): add profile assertion flow to oauth authenticator component

lazebnyi:
## What 
To authenticate in some connectors like Google Sheets, we need to pass two-step authentication. Before requesting an access token, the user needs to generate a JWT to authenticate the access token request with it. This is described in RFC 7523 - https://datatracker.ietf.org/doc/html/rfc7523.

## How 
We have extended the current OAuth implementation with new fields (e.g. `profile_assertion`, `use_profile_assertion`) and methods to use OAuth JWT Assertion Profiles.

## Summary by CodeRabbit

- **New Features**
	- Enhanced OAuth2 authentication with optional profile assertion support.
	- Added flexibility for client credentials configuration.
	- Introduced new methods for handling JWT-based token refresh.

- **Improvements**
	- Made `client_id` and `client_secret` optional.
	- Added support for conditional grant type configuration.
	- Improved error handling for authentication scenarios.

- **Tests**
	- Added new test coverage for profile assertion functionality.
	- Updated existing tests to enhance validation of refresh token scenarios.
	- Modified mock request behavior to reflect updated refresh endpoint URL.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -1058,8 +1058,6 @@ definitions:
     type: object
     required:
       - type
-      - client_id
-      - client_secret
     properties:
       type:
         type: string
@@ -1254,6 +1252,15 @@ definitions:
             default: []
             examples:
               - ["invalid_grant", "invalid_permissions"]
+      profile_assertion:
+        title: Profile Assertion
+        description: The authenticator being used to authenticate the client authenticator.
+        "$ref": "#/definitions/JwtAuthenticator"
+      use_profile_assertion:
+        title: Use Profile Assertion
+        description: Enable using profile assertion as a flow for OAuth authorization.
+        type: boolean
+        default: false
       $parameters:
         type: object
         additionalProperties: true
--- a/airbyte_cdk/sources/declarative/auth/oauth.py
+++ b/airbyte_cdk/sources/declarative/auth/oauth.py
@@ -118,6 +118,11 @@ class DeclarativeOauth2Authenticator(AbstractOauth2Authenticator, DeclarativeAut
             if self.token_expiry_date
             else pendulum.now().subtract(days=1)  # type: ignore # substract does not have type hints
         )
+        self.use_profile_assertion = (
+            InterpolatedBoolean(self.use_profile_assertion, parameters=parameters)
+            if isinstance(self.use_profile_assertion, str)
+            else self.use_profile_assertion
+        )
         self.assertion_name = "assertion"
 
         if self.access_token_value is not None:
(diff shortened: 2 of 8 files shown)


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1254,6 +1252,15 @@ definitions:
             default: []
             examples:
               - ["invalid_grant", "invalid_permissions"]
+      profile_assertion:
+        title: Profile Assertion
+        description: The authenticator being used to authenticate the client authenticator.
+        "$ref": "#/definitions/JwtAuthenticator"
+      use_profile_assertion:
```
maxi297, line 1259: Why do we need both `profile_assertion` and `use_profile_assertion`. Can't we assume that if `profile_assertion` is not empty, than we use profile assertion? Else, can we clarify the description to explain what happens when:
* `profile_assertion` and `use_profile_assertion==True`
* `profile_assertion` and `use_profile_assertion==False or None`
* not `profile_assertion` and `use_profile_assertion==True`
* (this one seems obvious but still...) not `profile_assertion` and `use_profile_assertion==False or None`

maxi297 (approved):
Just a small question regarding the declarative interface. The rest makes sense to me if the tests with source-google-sheets are good

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
lazebnyi: The use_profile_assertion was implemented to resolve uncertainty when `profile_assertion`, `client_id` and `client_secret` are provided in the manifest. This issue could be resolved by the default behavior and by defining it in the documentation. However, I decided that `use_profile_assertion` makes it more obvious which flow will be used for authentication. 

* `profile_assertion` and `use_profile_assertion==True`
authentication will be use `profile_assertion` flow to authenticate stream, raise exception when `profile_assertion` is not provided and allowing that `client_id` and `client_secret` could not be provided 

* `profile_assertion` and `use_profile_assertion==False or None`
authentication will be use basic behavior and use `client_id` and `client_secret` flow to authenticate stream and raise exception when any of those field is not provided

* not `profile_assertion` and `use_profile_assertion==True`
an error will be raised with the following message: `OAuthAuthenticator configuration error: 'profile_assertion' is required when using the profile assertion flow.`

* (this one seems obvious but still...) not `profile_assertion` and `use_profile_assertion==False or None`
authentication will use the basic behavior and the client_id and client_secret flow to authenticate the stream. It will raise an exception when either of these fields is not provided, allowing for cases where profile_assertion may not be provided.

aldogonzalez8 (approved):
These changes make sense to me. I would like a manifest test, but I don't want to block you on that as we already have a required implementation for this.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
aldogonzalez8: It probably is fine having "two flows," the error handling seems correct to me, but I also see this as confusing in the future. Should we give this responsibility to a new class? <-not super string on this but is another way.

Also we could benefit from updating documentation for DeclarativeOauth2Authenticator.

PR #241, opened 2025-01-22
fix: don't mypy unit_tests

natikgadzhi:
Bad, temporary fix to avoid the editor becoming a full clown fiesta.

Is there a path to have a relaxed set of rules for unit_tests and actually enforce them? It feels bad to drop mypy from editors in tests completely. 

## Summary by CodeRabbit

- **Configuration**
	- Updated type checking configuration to exclude unit test directory from mypy type checking.


aaronsteers:
> Is there a path to have a relaxed set of rules for unit_tests and actually enforce them? It feels bad to drop mypy from editors in tests completely.

Yes, I think we can have relaxed rules in specific subdirectories. I know this is possible with Ruff and 92% confident we can do it in mypy also.

aaronsteers (approved):
LGTM

PR #260, opened 2025-01-23
feat(file-based): sync file acl permissions and identities

aldogonzalez8:
## What
We want to introduce logic for syncing access control lists and other metadata and Identities for connectors, specifically in this implementation for File Bases connectors. 

If you are wondering what we mean by identities and ACLs here are a couple of documents that should clarify the general idea:
- [Permissions and Identity Streams](https://docs.google.com/document/d/1fL6iIvf7WcmCL2Ho24lH1krbPwQYA5tLlzBN7EZDtgk/edit?pli=1&tab=t.0#heading=h.1iugmixxbryj)
- [Access Control Lists (ACL) in Airbyte](https://docs.google.com/document/d/1Qr9SpoB8sTbdjd1A-XNNi8mvalI_UHCIB__un5cz8y4/edit?tab=t.0#heading=h.ys7dj6lh5ycr)

Relates to [[source-google-drive] Add ACL: permissions colmns and identities stream](https://github.com/airbytehq/airbyte-internal-issues/issues/11469)

Reltaes to [[source-microsoft-sharepoint] Implement Identities and Permissions](https://github.com/airbytehq/airbyte-internal-issues/issues/11610)

## How
We will leverage DefaultFileBased stream and stream reader to reuse most of the logic for scrapping files and let connectors implement the logic from the domain they handle. This initial iteration scopes introducing ACLs and Identities for [source-google-drive in this PR](https://github.com/airbytehq/airbyte/pull/52099), but in the future, we expect to add more connectors.

In the UI we will add a new Transfer Mode to Replicate Permissions ACL

Note: This image is from source-google-drive, which introduces a domain optional field; this is not part of base config mode as the domain could not be a concept for other connectors.

image

## Review guide

1. `airbyte_cdk/sources/file_based/file_based_source.py`:  logic to introduce new PermissionsFileBasedStream and Identities streams. We depend if use_permissoins_transfer mode is selected, if not we fallback to DefaultFileBased Stream
2. `airbyte_cdk/sources/file_based/stream/permissions_file_based_stream.py`: This new stream inherits from DefaultFileBasedStream as we reuse most of the logic to find Files, but we override for providing a schema and read records.
3. `airbyte_cdk/sources/file_based/stream/identities_stream.py`: Simple FileIdentities stream to read Identities, again we override, from new Identities stream at global CDK lever as in the future we expect other non-file based connectors will sync Identities.
4. `airbyte_cdk/sources/specs/transfer_modes.py`: added a Mode for DeliverPermissions, again we expect this is reusable for other nont-file based stream in the future.
5. `airbyte_cdk/sources/file_based/config/validate_config_transfer_modes.py`: just moved logic to determine transfer modes here as we were duplicating code for Source and StreamReader.
6. `airbyte_cdk/sources/file_based/file_based_stream_reader.py`: new abstract methods for file permissions and identities so streams can use those.

## User Impact
We want to enable users for knowledge retrieval scenarios where based on the hierarchy of groups, users and folders in the source, can user A see file B?

- [ ] YES 💚
- [ ] NO ❌

## Summary by CodeRabbit

- **New Features**
  - Enhanced file transfer options now support permissions replication, enabling delivery of access permissions along with identity data.
  - Introduced an additional delivery method option to mirror source permission restrictions through identity stream inclusion.
  
- **Tests**
  - Added comprehensive tests to ensure the reliability and effectiveness of the new permissions and identity stream functionalities.


aaronsteers:
@aldogonzalez8 - I'm updating from `main` to resolve a CI issue on the "Test Connectors" workflow.

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -22,6 +22,16 @@ class Config(OneOfOptionConfig):
...
 
+    sync_acl_permissions: bool = Field(
+        title="Include ACL Permissions",
+        description="Joins Document allowlists to each stream.",
+        default=False,
+        airbyte_hidden=True,
+    )
+    domain: Optional[str] = Field(
+        title="Domain", description="The domain of the identities.", airbyte_hidden=True
+    )
```
aaronsteers, line 33: @aldogonzalez8 - Can we deliver this with net new streams and without altering existing content streams? I see we're adding identity but I'm worried about adding permissions into content stream as raised earlier. Do I understand that this PR does add permissions into the content? If we could add permissions as a separate stream, I believe this should scale better.

For instance, the real-world scenario of a new team member being added or dropped from a team, and getting added to a 2TB fileset:

1. If the permission change is detected and triggers incremental sync, it'll require resycing the entire 2TB of data.
2. If the change is not detected, then permissions will be wrong until every file gets updated and replicated over another way.

aldogonzalez8: Hey @aaronsteers , I would be happy of sync to talk about this. Just to give more clarity, we will request users to create two separate connections:

1.- File sync to move files to some destination, this moment only S3.
2.- Acl and Identities connection to whatever destination they want.

I believe we are expecting to have full refresh for Acl and Identities, actually I'm just shipping full refresh for identities stream they get for free when enabling `sync_acl_permissions `, but I can see working the files acl stream as incremental.

Now for the scenario you mention the connection syncing files, yes it will, move the 2TB of files in universal file transfer mode but for the connection syncing acl and identities it will resync the acl of the file and also the identities  stream that will have the update on the group members. 

I just created [a loom](https://www.loom.com/share/0d9a1f1c9500449fbc018bc2945d08cb) to walk through the feature.

aaronsteers: @aldogonzalez8 (cc @natikgadzhi) - My concern is that we _can't_ add permissions into the content stream(s) without creating a bad user experience. We really do need to move those into their own stream. To do so later will break any users who adopt the feature now.

Is there any way to carve out permissions into their own stream, rather than adding as new properties?

We really don't want to be in the situation where either of these are true:
1. Adding a team member forces the entire drive to force a full refresh of every file in the drive(s).
2. Permissions have to be assumed to be always wrong except if the file has just been modified.

As new properties on the existing stream, we're stuck picking between these both bad options - but if we split permissions to a separate stream we can have efficient incremental refresh _and_ also always-up-to-date permission info.

aaronsteers (dismissed):
@aldogonzalez8 (cc @natikgadzhi)

I think I misunderstood what this mode was doing. Because of the wording, I read this as we were "including" permissions along with the sync data from records. I understand now that we're not syncing the data at all, and this mode only syncs permissions.

As such, can we consider a difference in wording? Either something like "Sync Permissions Instead of Record Data" or (I think, maybe better), a third mode in our toggle?

image


Wdyt of a third mode to the above toggle? So, then it would be a choice between:

- `Replicate Records`
- `Copy Raw Files`
- `Replicate File Permissions`

In my mind at least, this makes it more clear that we won't sync records _and_ permissions, but instead we'll sync records _or_ permissions (_or_ raw files).

What do you think of this?

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -22,6 +22,16 @@ class Config(OneOfOptionConfig):
 
     delivery_type: Literal["use_records_transfer"] = Field("use_records_transfer", const=True)
 
+    sync_acl_permissions: bool = Field(
```
natikgadzhi, line 25: Non-blocking, we can move this later, but since we're doing a dev version anyway: 

How does this config option scale to other connectors? Can we put this into the connector instead of the CDK until there are at least 2 connectors who both need this? 

My concern is that this is just an arbitrary boolean, I don't think ACL is only specific to file-based connectors — any connector can have ACL streams.

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -22,6 +22,16 @@ class Config(OneOfOptionConfig):
...
+        title="Include ACL Permissions",
+        description="Joins Document allowlists to each stream.",
+        default=False,
+        airbyte_hidden=True,
+    )
+    domain: Optional[str] = Field(
```
natikgadzhi, line 31: Are we SURE that _any_ identities will have domains? Whad does domain mean if an identity is a UUID?

airbyte_cdk/sources/file_based/config/permissions.py
```diff
@@ -0,0 +1,34 @@
+#
```
natikgadzhi, line 1: I am not certain this should be specific to just file-based connectors. It's possible we want these models to be available to any connector type.

airbyte_cdk/sources/file_based/config/permissions.py
```diff
@@ -0,0 +1,34 @@
...
+
+import uuid
+from datetime import datetime
+from enum import Enum
+
+from pydantic.v1 import BaseModel
+
+
+class RemoteFileIdentityType(Enum):
```
natikgadzhi, line 12: Imagie there are `RemoteRecordIdentityType`. Is it just in the different chunk of CDK then? 

If we just want `RemoteIdentityType` instead of focusing on just files, then we could lift this to global CDK.

airbyte_cdk/sources/file_based/config/permissions.py
```diff
@@ -0,0 +1,34 @@
...
+    USER = "user"
+    GROUP = "group"
+
+
+class RemoteFileIdentity(BaseModel):
```
natikgadzhi, line 17: A one-senctence docstring would be very useful, explaining why this exists and why most fields are nulllable, and in what situations we want to use them

airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -387,6 +412,14 @@ def _use_file_transfer(parsed_config: AbstractFileBasedSpec) -> bool:
         )
         return use_file_transfer
 
+    @staticmethod
+    def _use_records_transfer(parsed_config: AbstractFileBasedSpec) -> bool:
+        use_records_transfer = (
+            hasattr(parsed_config.delivery_method, "delivery_type")
```
natikgadzhi, line 418: Nit: maybe extract these strings into a file with constants?

airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -135,6 +135,15 @@ def use_file_transfer(self) -> bool:
             return use_file_transfer
         return False
 
+    def use_records_transfer(self) -> bool:
```
natikgadzhi, line 138: Nit: Should we use the one on the source class instead, passing the config in? (as in, I genuinely don't know, educate me ;-)

airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -183,3 +202,17 @@ def _get_file_transfer_paths(self, file: RemoteFile, local_directory: str) -> Li
         makedirs(path.dirname(local_file_path), exist_ok=True)
         absolute_file_path = path.abspath(local_file_path)
         return [file_relative_path, local_file_path, absolute_file_path]
+
+    def get_file_acl_permissions(self, file: RemoteFile, logger: logging.Logger) -> Dict[str, Any]:
+        """
+        This is required for connectors that will support syncing
+        ACL Permissions from files.
+        """
+        return {}
```
natikgadzhi, line 211: Should this throw a NotImplementedError instead?

airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -183,3 +202,17 @@ def _get_file_transfer_paths(self, file: RemoteFile, local_directory: str) -> Li
...
+
+    def load_identity_groups(self, logger: logging.Logger) -> Iterable[Dict[str, Any]]:
+        """
+        This is required for connectors that will support syncing
+        identities.
+        """
+        yield {}
```
natikgadzhi, line 218: Ditto —should this throw instead?

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
aaronsteers: Resolved. I understand we're not syncing any data in the same stream as the permissions. I think we can just tweak the wording and UI a bit to make it clear _only_ permissions are syncing in this mode.

natikgadzhi (commented):
I'm just reading through this. If we're building a dev version of a connector and not in a rush to merge, I'd love to dive in and learn from this, some pieces of adding config options at CDK level but only to file-based connectors are interesting to me — I think maybe identity streams should not be only for file-based streams.

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
aldogonzalez8: No, I can let this one on the connector side, I'm looking to not having to request it.

airbyte_cdk/sources/file_based/file_based_stream_reader.py
aldogonzalez8: I will use a helper method to avoid a circular dependency.

airbyte_cdk/sources/file_based/file_based_stream_reader.py
aldogonzalez8: Yes, I'm curious about the file-based connector already bumped to the latest CDK, but I guess the weekly CDK upgrade will fail for those till we make these method noop for them(?)

airbyte_cdk/sources/file_based/file_based_stream_reader.py
aldogonzalez8: Yes, I'm curious about the file-based connector already bumped to the latest CDK, but I guess the weekly CDK upgrade will fail for those till we make these method noop for them(?)

aldogonzalez8:
> @aldogonzalez8 (cc @natikgadzhi)
> 
> I think I misunderstood what this mode was doing. Because of the wording, I read this as we were "including" permissions along with the sync data from records. I understand now that we're not syncing the data at all, and this mode only syncs permissions.
> 
> As such, can we consider a difference in wording? Either something like "Sync Permissions Instead of Record Data" or (I think, maybe better), a third mode in our toggle?
> 
> image
> Wdyt of a third mode to the above toggle? So, then it would be a choice between:
> 
> * `Replicate Records`
> * `Copy Raw Files`
> * `Replicate File Permissions`
> 
> In my mind at least, this makes it more clear that we won't sync records _and_ permissions, but instead we'll sync records _or_ permissions (_or_ raw files).
> 
> What do you think of this?

@aaronsteers This has been added; you can find the sample [here](https://cloud.airbyte.com/workspaces/3453c03a-a134-4d2f-aed0-3ec1dc83a16d/source/394f4721-8df3-4185-87d1-09b6153ce06d).

Note: I included the option "Include Identity Stream" with default True, which we can remove later, but I think there is still value in users who don't have access to the domain or similar concepts in other connectors. e.g., a Gmail user that wants ACLs but doesn't have access to public Google workspace.

image

airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
aldogonzalez8: I moved the transfer mode spec to global cdk

airbyte_cdk/sources/file_based/config/permissions.py
aldogonzalez8: I moved the transfer mode spec to global cdk

airbyte_cdk/sources/file_based/config/permissions.py
aldogonzalez8: For now I will let the source to determine the schema of the file acls and identities streams.

airbyte_cdk/sources/file_based/file_based_source.py
aldogonzalez8: Sure, makes sense.

airbyte_cdk/sources/file_based/config/validate_config_transfer_modes.py
```diff
@@ -0,0 +1,80 @@
...
+from airbyte_cdk.sources.specs.transfer_modes import DeliverPermissions
+
+DELIVERY_TYPE_KEY = "delivery_type"
+DELIVERY_TYPE_PERMISSION_TRANSFER_MODE_VALUE = "use_permissions_transfer"
+DELIVERY_TYPE_FILES_TRANSFER_MODE_VALUE = "use_file_transfer"
+PRESERVE_DIRECTORY_STRUCTURE_KEY = "preserve_directory_structure"
+INCLUDE_IDENTITIES_STREAM_KEY = "include_identities_stream"
+
+
+def use_file_transfer(parsed_config: AbstractFileBasedSpec) -> bool:
```
natikgadzhi, line 18:
```suggestion
def use_file_transfer(parsed_config: AbstractFileBasedSpec) -> bool:
"""Returns `True` if the configuration uses file transfer mode."""
```

airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -49,14 +55,21 @@
...
+)
 from airbyte_cdk.sources.file_based.stream.concurrent.adapters import FileBasedStreamFacade
 from airbyte_cdk.sources.file_based.stream.concurrent.cursor import (
     AbstractConcurrentFileBasedCursor,
     FileBasedConcurrentCursor,
     FileBasedFinalStateCursor,
 )
 from airbyte_cdk.sources.file_based.stream.cursor import AbstractFileBasedCursor
+from airbyte_cdk.sources.file_based.stream.permissions_file_based_stream import (
+    PermissionsFileBasedStream,
```
natikgadzhi, line 71: In the list of imports, this looks interesting — any reason why this stream is not at `file_based.stream` import level?

airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -49,14 +55,21 @@
     DEFAULT_SCHEMA_VALIDATION_POLICIES,
     AbstractSchemaValidationPolicy,
 )
-from airbyte_cdk.sources.file_based.stream import AbstractFileBasedStream, DefaultFileBasedStream
+from airbyte_cdk.sources.file_based.stream import (
+    AbstractFileBasedStream,
+    DefaultFileBasedStream,
+    FileIdentities,
```
natikgadzhi, line 61: Is there a reason why the naming convention is different? I.e. `DefaultFileBasedStream` but `FileIdentities`, as opposed to `FileIdentitiesStream`?

airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -183,3 +184,45 @@ def _get_file_transfer_paths(self, file: RemoteFile, local_directory: str) -> Li
         makedirs(path.dirname(local_file_path), exist_ok=True)
         absolute_file_path = path.abspath(local_file_path)
         return [file_relative_path, local_file_path, absolute_file_path]
+
+    @abstractmethod
+    def get_file_acl_permissions(self, file: RemoteFile, logger: logging.Logger) -> Dict[str, Any]:
+        """
+        This is required for connectors that will support syncing
```
natikgadzhi, line 191: It would be cool to explain what that function is supposed to return, perhaps with pseudocode example. 

I.e. "This function should return the allow list for a given file, i.e. the list of all identieis and their permission levels associated with it."

airbyte_cdk/sources/file_based/stream/identities_stream.py
```diff
@@ -0,0 +1,47 @@
...
+
+from airbyte_cdk.sources.file_based.config.file_based_stream_config import PrimaryKeyType
+from airbyte_cdk.sources.file_based.discovery_policy import AbstractDiscoveryPolicy
+from airbyte_cdk.sources.file_based.exceptions import FileBasedErrorsCollector
+from airbyte_cdk.sources.file_based.file_based_stream_reader import AbstractFileBasedStreamReader
+from airbyte_cdk.sources.streams.core import JsonSchema
+from airbyte_cdk.sources.streams.permissions.identities import Identities
+
+
+class FileIdentities(Identities):
```
natikgadzhi, line 16: I've commented above — there is a world where this stream name is consistent with the `DefaultFileBasedStream` and friends, but this is a non-blocking nit.

natikgadzhi (approved):
Approving to get this rolling — we might revisit some pieces when we work on sharepoint and that's okay! 

@aldogonzalez8 I've left some comments, but they are all non-blocking nitpicks, up to you. Feel free to merge please!

airbyte_cdk/sources/file_based/file_based_source.py
aldogonzalez8: Because Aldo is dumbe and left pycharm do the import.

airbyte_cdk/sources/file_based/config/validate_config_transfer_modes.py
aldogonzalez8: Oh I was missing that, thanks.

airbyte_cdk/sources/file_based/file_based_source.py
aldogonzalez8: Nah, let's be consistents.

airbyte_cdk/sources/file_based/stream/identities_stream.py
aldogonzalez8: I will fix consistency

airbyte_cdk/sources/file_based/file_based_source.py
aldogonzalez8: I've done this.

airbyte_cdk/sources/file_based/file_based_source.py
aldogonzalez8: I've done this.

airbyte_cdk/sources/file_based/file_based_stream_reader.py
aldogonzalez8: Just added more doc with pseudocode

airbyte_cdk/sources/file_based/stream/identities_stream.py
aldogonzalez8: I did

PR #261, opened 2025-01-23
fix(low-code): Fix legacy state migration in SubstreamPartitionRouter

tolik0:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved state handling in substream partition routing to prevent incorrect state assignments
  - Enhanced robustness of state filtering mechanism

- **Tests**
  - Added comprehensive test cases for substream partition router state management
  - Expanded test coverage for different initial state scenarios

### DIFF
--- a/airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
+++ b/airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
@@ -296,8 +296,12 @@ class SubstreamPartitionRouter(PartitionRouter):
 
         if not parent_state and incremental_dependency:
             # Attempt to retrieve child state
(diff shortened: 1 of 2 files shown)


tolik0:
I found this error in CATs for Jira: The parent state was missing, causing the substream partition router to attempt migrating the substream state for the parent stream. However, it should only be migrated if it is in the legacy format. We didn’t encounter errors related to this issue previously because the parent state was updated during the sync. In contrast, with the concurrent per-partition cursor, the parent state is updated at the end of the sync, which created this corner case where the parent state is missing.

aldogonzalez8 (approved):
APPROVED

PR #279, opened 2025-01-28
feat(low-code concurrent): Add `use_global_cursor` flag to ConcurrentPerPartitionCursor

tolik0:
## Summary by CodeRabbit

- **New Features**
	- Enhanced state management for concurrent partition cursors.
	- Added ability to switch between global and per-partition state handling.
	- Introduced mechanism to track partition cursor limits.

- **Tests**
	- Updated test cases to validate new state management scenarios.
	- Added test configurations for global and per-partition cursor states.

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -232,7 +232,10 @@ class ConcurrentPerPartitionCursor(Cursor):
                 self._over_limit += 1
                 # Try removing finished partitions first
                 for partition_key in list(self._cursor_per_partition.keys()):
-                    if partition_key in self._finished_partitions and self._semaphore_per_partition[partition_key]._value == 0:
+                    if (
+                        partition_key in self._finished_partitions
+                        and self._semaphore_per_partition[partition_key]._value == 0
+                    ):
                         oldest_partition = self._cursor_per_partition.pop(
                             partition_key
                         )  # Remove the oldest partition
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -225,14 +229,15 @@ def _ensure_partition_limit(self) -> None:
         """
         with self._lock:
             while len(self._cursor_per_partition) > self.DEFAULT_MAX_PARTITIONS_NUMBER - 1:
+                self._over_limit += 1
                 # Try removing finished partitions first
                 for partition_key in list(self._cursor_per_partition.keys()):
-                    if partition_key in self._finished_partitions:
+                    if partition_key in self._finished_partitions and self._semaphore_per_partition[partition_key]._value == 0:
```
maxi297, line 235: I am a bit conflicted about accessing the private field here. The access to `_value` is not thread-safe per se. For example, we could have the following sequence of events:
* The lock could be acquired [here](https://github.com/python/cpython/blob/main/Lib/threading.py#L532)
* We evaluate `self._semaphore_per_partition[partition_key]._value == 0` on the line above and it is true
* The semaphore value gets increased even if the lock was held

I does not seem like there is a good way to access the value of a semaphore though... So, two questions:
* Why did we added this condition? 
* Do we use it in a way where the thread safety for this access is not important?

maxi297 (commented):
One question regarding thread safety

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: The semaphore is acquired under the lock, and this check of the value is also performed under the lock, so a race condition is not possible. Additionally, we release the semaphore before adding the partition to the finished partitions. Therefore, a partition can only have a semaphore value of zero and be in the finished partitions when all cursor partitions for the partition router partitions are closed.

maxi297: The fact that all the acquiring/instantiating of the semaphore is done under the lock seems to make sense to me. I'm good with this then, thanks for the explanation!

maxi297 (approved):
:shipit:

PR #285, opened 2025-01-28
fix: (OAuthAuthenticator) - get the `access_token`, `refresh_token`, `expires_in` recursively from `response`

bazarnov:
## What
Resolving:
- https://github.com/airbytehq/airbyte-internal-issues/issues/11488

Currently the `OAuthAuthenticator` e.g `DeclarativeOauth2Authenticator` and `SingleUseRefreshTokenOauth2Authenticator` don't support the case when the `access_token` or similar field is `nested` under another entity like `data` or similar. 
We need a way to handle this dynamically for the Customer and never think about what is the `PATH` for the target `keys` to set.

## How
- added new private method `_find_and_get_value_from_response` to pull out the target key recursively, up to 5 levels of nesting (by default)
- covered and extended the `unit_tests` to reflect the change
- refactored the code and corrected unit test accordingly to simplify to logic

## User Impact

## Release Notes

- **New Features**
  - Enhanced OAuth token retrieval with support for nested JSON response structures.
  - Improved token refresh mechanism with more robust value extraction.

- **Bug Fixes**
  - Added recursive search method to handle complex authentication response formats.
  - Refined error handling during token refresh operations.

- **Improvements**
  - Increased flexibility in parsing authentication responses.
  - Enhanced error detection and handling for OAuth token retrieval.
  - Expanded test coverage for handling nested JSON responses in token retrieval.


airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -292,6 +305,59 @@ def _message_repository(self) -> Optional[MessageRepository]:
...
+        Returns:
+            Any: The value associated with the specified key if found, otherwise None.
+
+        Raises:
+            AirbyteTracedException: If the maximum recursion depth is reached without finding the key.
+        """
+        if current_depth > max_depth:
+            # this is needed to avoid an inf loop, possible with a very deep nesting observed.
+            message = f"The maximum level of recursion is reached. Couldn't find the speficied `{key_name}` in the response."
+            raise ResponseKeysMaxRecurtionReached(
```
bnchrch, line 333: ❓ do we want it to raise? What about optional keys?

bnchrch (approved):
One question but otherwise pre-approving

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
bazarnov: Ideally, yes, as the indication of the massive edge case.

PR #286, opened 2025-01-28
feat: use create_concurrent_cursor_from_perpartition_cursor

artem1205:
## What

resolving https://github.com/airbytehq/airbyte/pull/48449

## How

add concurrent_cursor_from_perpartition_cursor to asyncretriever

## User Impact

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -11,6 +11,7 @@ from datetime import timedelta
 from typing import Any, Callable, Iterable, Mapping, MutableMapping, Optional
 
 from airbyte_cdk.sources.connector_state_manager import ConnectorStateManager
+from airbyte_cdk.sources.declarative.async_job.job_orchestrator import AsyncPartition
 from airbyte_cdk.sources.declarative.incremental.global_substream_cursor import (
     Timer,
     iterate_with_last_flag_and_state,
@@ -325,7 +326,11 @@ class ConcurrentPerPartitionCursor(Cursor):
                 "Invalid state as stream slices that are emitted should refer to an existing cursor"
             )
         self._cursor_per_partition[
-            self._to_partition_key(record.associated_slice.partition)
+            self._to_partition_key(
+                record.associated_slice.stream_slice.partition
+                if isinstance(record.associated_slice, AsyncPartition)
+                else record.associated_slice.partition
+            )
         ].observe(record)
 
     def _to_partition_key(self, partition: Mapping[str, Any]) -> str:
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -19,6 +19,7 @@ from airbyte_cdk.sources.declarative.extractors import RecordSelector
 from airbyte_cdk.sources.declarative.extractors.record_filter import (
     ClientSideIncrementalRecordFilterDecorator,
 )
+from airbyte_cdk.sources.declarative.incremental import ConcurrentPerPartitionCursor
 from airbyte_cdk.sources.declarative.incremental.datetime_based_cursor import DatetimeBasedCursor
 from airbyte_cdk.sources.declarative.incremental.per_partition_with_global import (
     PerPartitionWithGlobalCursor,
@@ -231,7 +232,7 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
                     ):
                         cursor = declarative_stream.retriever.stream_slicer.stream_slicer
 
-                        if not isinstance(cursor, ConcurrentCursor):
+                        if not isinstance(cursor, ConcurrentCursor | ConcurrentPerPartitionCursor):
                             # This should never happen since we instantiate ConcurrentCursor in
                             # model_to_component_factory.py
                             raise ValueError(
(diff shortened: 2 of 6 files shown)


artem1205:
Note to self: A working https://github.com/airbytehq/airbyte-python-cdk/pull/286/commits/185fe428a80460e9d916f8602be86b18182e9d9c with `AsyncPartition` as the output of `stream_slices()`

artem1205:
Note to review: in [d6c4870](https://github.com/airbytehq/airbyte-python-cdk/pull/286/commits/d6c48705a3ee3519cb98c12737cf2a3e5deee3d8) some signatures were changed, but I guess it should not affect any connectors

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -325,7 +326,11 @@ def observe(self, record: Record) -> None:
                 "Invalid state as stream slices that are emitted should refer to an existing cursor"
             )
         self._cursor_per_partition[
-            self._to_partition_key(record.associated_slice.partition)
+            self._to_partition_key(
+                record.associated_slice.stream_slice.partition
+                if isinstance(record.associated_slice, AsyncPartition)
```
maxi297, line 331: It seems like a bad smell to have to check the instance of the stream slice. Do we actually need this? I thought that the records were created [during the record selection](https://github.com/airbytehq/airbyte-python-cdk/blob/3af96dcd713761d56190cb485f9376c65b1279c6/airbyte_cdk/sources/declarative/extractors/record_selector.py#L111) which would take the stream slice we pass [here](https://github.com/airbytehq/airbyte-python-cdk/blob/3af96dcd713761d56190cb485f9376c65b1279c6/airbyte_cdk/sources/declarative/retrievers/async_retriever.py#L95-L100). Is it just a matter of passing the right stream slice? 

The same comment applies to https://github.com/airbytehq/airbyte-python-cdk/pull/286/files#diff-6c148ab36dce37edc24313b91039c3122790e513b1cbba9bc3a4970460e37592R149

unit_tests/sources/declarative/async_job/test_integration.py
```diff
@@ -82,7 +87,7 @@ def streams(self, config: Mapping[str, Any]) -> List[Stream]:
                     config={},
                     parameters={},
                     record_selector=noop_record_selector,
-                    stream_slicer=AsyncJobPartitionRouter(
+                    stream_slicer=MockAsyncJobPartitionRouter(
```
maxi297, line 90: Why did we need to change this?

airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
```diff
@@ -482,7 +482,7 @@ def _is_breaking_exception(self, exception: Exception) -> bool:
             and exception.failure_type == FailureType.config_error
         )
 
-    def fetch_records(self, partition: AsyncPartition) -> Iterable[Mapping[str, Any]]:
+    def fetch_records(self, partition: Iterable[AsyncJob]) -> Iterable[Mapping[str, Any]]:
```
maxi297, line 485: nit: can we rename the variable?

maxi297 (commented):
My main concern is for us to respect the APIs of the object we return so that we don't have to do `isinstance`. Can we confirm that? We can have a sync to figure out a solution my comments aren't enough

airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
artem1205: renamed

unit_tests/sources/declarative/async_job/test_integration.py
maxi297: I don't see the reason why it would need us to mock the AsyncJobPartitionRouter. I think this test is broken for multiple reasons:
* It uses the old CDK flow which we don't normally allow for AsyncRetrievers
* [This condition](https://github.com/airbytehq/airbyte-python-cdk/blob/4e7f94a171311fb24e6d15affb5bb6f5dd3bd717/airbyte_cdk/sources/declarative/declarative_stream.py#L141) removes the extra_fields from the StreamSlice and therefore should probably be `if stream_slice is None or (not isinstance(stream_slice, StreamSlice) and stream_slice == {}):`

Can we at least fix the second point? I think it would be interesting to fix the first one as well but the implications are only on the state so I'm fine with adding a comment saying that as of 2025-01-30, this test uses the legacy CDK even though this is not a valid case in prod as it should use the concurrent CDK for state management reasons

unit_tests/sources/declarative/async_job/test_integration.py
```diff
@@ -36,6 +36,11 @@
 _NO_LIMIT = 10000
 
 
+class MockAsyncJobPartitionRouter(AsyncJobPartitionRouter):
+    def fetch_records(self, job: AsyncJob) -> Iterable[Mapping[str, Any]]:
```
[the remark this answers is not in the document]
artem1205: removed

airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
```diff
@@ -482,16 +482,16 @@ def _is_breaking_exception(self, exception: Exception) -> bool:
             and exception.failure_type == FailureType.config_error
         )
 
-    def fetch_records(self, partition: AsyncPartition) -> Iterable[Mapping[str, Any]]:
+    def fetch_records(self, async_job_list: Iterable[AsyncJob]) -> Iterable[Mapping[str, Any]]:
```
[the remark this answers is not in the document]
artem1205: agreed, renamed

artem1205:
> Can we at least fix the second point?

fixed, thanks for pointing out.

artem1205:
upd: `MockAsyncJobPartitionRouter` was removed as well

maxi297 (approved):
LGTM! Thanks for addressing the concerns

PR #293, opened 2025-01-29
feat(low-code): add use check availability flag to dynamic check

lazebnyi:
## What 

Check availability by attempting to fetch the first slice of data during the connection check. However, for some dynamic streams, this slice may be empty, leading to a failed connection check even if the stream has been created and we have sufficient permissions to access data (e.g., `source-google-sheets` will create a stream if there is at least one row in the sheet, but it won’t create a slice if there are no additional rows).

## How 
Add a `use_check_availability` flag to bypass the connection check if streams have been successfully created.

## Summary by CodeRabbit

- **New Features**
	- Added `use_check_availability` option to control stream availability checks.
	- Introduced more flexible configuration for dynamic stream checking.

- **Improvements**
	- Enhanced error handling and logging for stream availability checks.
	- Improved schema resolution for complex field types.

- **Bug Fixes**
	- Refined test cases to handle various scenarios for availability checks.


airbyte_cdk/sources/declarative/checks/check_dynamic_stream.py
```diff
@@ -31,21 +32,27 @@ def check_connection(
         self, source: AbstractSource, logger: logging.Logger, config: Mapping[str, Any]
     ) -> Tuple[bool, Any]:
         streams = source.streams(config=config)
+
         if len(streams) == 0:
             return False, f"No streams to connect to from source {source}"
+        if not self.use_check_availability:
```
maxi297, line 38: I would go further: is there a reason why we wouldn't want to always do this? Aren't we always at risk of having customers with an empty <insert API like google-sheets> table?

lazebnyi: We need to use `check_availability` in cases where we have both static and dynamic streams from a single source. In this scenario, `source.streams() `will always return a non-empty list.

maxi297: I'm still confused. What makes it that if we have both dynamic and static streams that the stream that we will evaluate will be within the `self.stream_count`. Should we specify them using stream name?

lazebnyi: It's a good question. In that case, I think we need to set `self.stream_count` to the static stream count + 1. However, if we need to test only one static stream, that could be problematic.

I plan to add stream name support later to make `CheckDynamicStreams` more universal. I can do this after this PR as a follow-up. WDYT?

maxi297: I don't mind that. I know the doc is very explicit about the fact that it is experimental but can we document what fields we know will change and link to a GitHub issue? After that, I'm good to approve

maxi297 (approved):
Preemptively approving

PR #297, opened 2025-01-30
fix(concurrent-partition-cursor): update cursor comparison to leverage ab_datetime

pnilan:
## What
- Related to P0: https://github.com/airbytehq/oncall/issues/7348
- Integrates `AirbyteDatetime` to parse and compare cursor values that may have different formats.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved datetime parsing and comparison logic for cursor state management to enhance precision and reliability of incremental data synchronization.

PR #298, opened 2025-01-30
fix(concurrent-partition-cursor): Fix cursor comparison error

tolik0:
## Summary by CodeRabbit

- **New Features**
	- Enhanced cursor state management for datetime-based incremental synchronization.
	- Added support for millisecond-level datetime parsing.

- **Improvements**
	- Improved modularity of cursor state extraction logic.
	- More precise handling of datetime formats in stream cursors.

- **Testing**
	- Updated test cases to reflect new datetime timestamp handling.
	- Introduced new timestamp constant for consistent testing.
	- Adjusted expected outputs for various test cases to align with new timestamp format.

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -396,10 +396,5 @@ class ConcurrentPerPartitionCursor(Cursor):
         cursor = self._cursor_per_partition[partition_key]
         return cursor
 
-    def _extract_cursor_value_from_state(self, state: StreamState) -> Any:
-        return self._connector_state_converter.parse_value(
-            state[self.cursor_field.cursor_field_key]
-        )
-
     def limit_reached(self) -> bool:
         return self._over_limit > self.DEFAULT_MAX_PARTITIONS_NUMBER
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -22,6 +22,9 @@ from airbyte_cdk.sources.streams.checkpoint.per_partition_key_serializer import 
 )
 from airbyte_cdk.sources.streams.concurrent.cursor import ConcurrentCursor, Cursor, CursorField
 from airbyte_cdk.sources.streams.concurrent.partitions.partition import Partition
+from airbyte_cdk.sources.streams.concurrent.state_converters.abstract_stream_state_converter import (
+    AbstractStreamStateConverter,
+)
 from airbyte_cdk.sources.types import Record, StreamSlice, StreamState
 
 logger = logging.getLogger("airbyte")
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -372,5 +396,10 @@ def _get_cursor(self, record: Record) -> ConcurrentCursor:
         cursor = self._cursor_per_partition[partition_key]
         return cursor
 
+    def _extract_cursor_value_from_state(self, state: StreamState) -> Any:
```
maxi297, line 399: Where is this used? Should we delete it?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -329,6 +332,27 @@ def _set_initial_state(self, stream_state: StreamState) -> None:
...
+        Initializes the global cursor state from the provided stream state.
+
+        If the cursor field key is present in the stream state, its value is parsed,
+        formatted, and stored as the global cursor. This ensures consistency in state
+        representation across partitions.
+        """
+        if self.cursor_field.cursor_field_key in stream_state:
+            global_state_value = stream_state[self.cursor_field.cursor_field_key]
+            final_format_global_state_value = self._connector_state_converter.output_format(
+                self._connector_state_converter.parse_value(global_state_value)
```
maxi297, line 346: Should we go the safest way and re-use [ab_datetime_parse](https://github.com/airbytehq/airbyte-python-cdk/pull/297/files#diff-40bb6219c9d2c70a591e9e3b2209087e1b01eb44bfe36c15ff22b43506ac69faR150-R152) as @pnilan did there?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1202,6 +1202,22 @@ def create_concurrent_cursor_from_perpartition_cursor(
...
+        datetime_format = datetime_based_cursor_model.datetime_format
+
+        cursor_granularity = (
+            parse_duration(datetime_based_cursor_model.cursor_granularity)
+            if datetime_based_cursor_model.cursor_granularity
+            else None
+        )
+
+        connector_state_converter: DateTimeStreamStateConverter
+        connector_state_converter = CustomFormatConcurrentStreamStateConverter(
```
maxi297, line 1214: nit: This is common logic with `create_concurrent_cursor_from_datetime_based_cursor`. Should we extract this in a method so that a change to one affects the other?

maxi297 (approved):
LGTM! Just a couple of comments that are not blocking

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: The `state_converter` is already used in `ConcurrentCursor` to parse the state, so I think we should apply the same logic in both classes for consistency.

Why would `ab_datetime_parse` be a safer option?

maxi297: Because it seemingly support multiple format types. That being said, I'm not attached to this if this isn't a transition issue

PR #301, opened 2025-01-30
feat(low-code): added json.loads to jwt authenticator

lazebnyi:
## What 

Some sources use JWT authentication and store all parameters in the 'config' field in JSON format (e.g., Google Sheets). This approach makes it problematic to natively use this JSON-format config without a custom implementation.

## How 

Pass `json.loads` to all interpolated fields in the JWT authenticator.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JWT authentication with improved JSON data processing
	- Added test coverage for secret key retrieval from configuration

- **Tests**
	- Introduced new test method to validate secret key configuration handling

### DIFF
--- a/airbyte_cdk/sources/declarative/auth/jwt.py
+++ b/airbyte_cdk/sources/declarative/auth/jwt.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+import json
 import base64
 from dataclasses import InitVar, dataclass
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


pnilan:
Should we use orjson instead?

lazebnyi:
@pnilan 

> Should we use orjson instead?

In that case I don't think that `orjson` will add a lot in performance, but it will add additional dependencies. So, I think `json` is preferable here for me.

maxi297 (approved):
I'm fine with this. We can make it global to all interpolation later if needed as it would not be a breaking change from what I can see. Thanks!

maxi297:
I'm not too attached on orjson on this one because it won't be called very often so the impact on performance does not seem very high

PR #303, opened 2025-01-30
chore: simplify RequestOption construction in component factory

ChristoGrab:
## What

Refactor components in the model_to_component factory that build child RequestOptions to leverage the re-usable `_create_component_from_model` method, rather than instantiating them with explicit attributes (ie, `field_name`). Hopefully this will reduce future tech debt if the RequestOption class is further modified moving forward.

Updates a couple tests to verify the behavior when using nested body_json injected values.

### DIFF
--- a/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
+++ b/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
@@ -6,11 +6,6 @@
 import datetime
 from typing import Any, Iterable, Mapping
 
-import logging
-
-logging.basicConfig(level=logging.DEBUG)
-logger = logging.getLogger(__name__)
-
 import freezegun
 import pendulum
 import pytest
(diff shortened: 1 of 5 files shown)


unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -743,7 +745,7 @@ def test_datetime_based_cursor():
         == "since_updated_at"
     )
     assert stream_slicer.end_time_option.inject_into == RequestOptionType.body_json
-    assert stream_slicer.end_time_option.field_name.eval({}) == "before_created_at"
+    assert stream_slicer.end_time_option.field_path[0].eval({}) == "before_created_at"
```
maxi297, line 748: nit: The outcome could be that `stream_slicer.end_time_option.field_path == ["before_created_at", "another dummy parameter that we don't expect"]`. Should we change this to `assert `stream_slicer.end_time_option.field_path == ["before_created_at"]`?

maxi297 (approved):
LGTM! Just a small nit

unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
ChristoGrab: Agreed, updated the test assertion

PR #305, opened 2025-01-31
fix(low-code): add wrong dynamic stream name type validation

lazebnyi:
## What 
During the dynamic component resolving process, data mapping is not limited by field types. However, during instance creation, the 'name' field is strictly parsed as a string. Therefore, when we perform stream grouping [here](https://github.com/airbytehq/airbyte-python-cdk/blob/65e6a0d12e19a84db360053d3ec793c66b203512/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L191), we encounter a KeyError because the stream name in the manifest and the stream instance name might be of different types.

## How 
Add type validation for the stream name during components resolving process.

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced validation to ensure stream names are strings, preventing potential runtime errors.

- **Tests**
	- Added a new test case to verify error handling for invalid stream name types.

### DIFF
--- a/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
@@ -482,16 +482,16 @@ class AsyncJobOrchestrator:
             and exception.failure_type == FailureType.config_error
         )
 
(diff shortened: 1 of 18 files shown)


artem1205 (approved):
LGTM!

PR #307, opened 2025-01-31
feat(low-code): improve logging on async retriever errors

maxi297:
See [this Slack thread](https://airbytehq.slack.com/archives/C027KKE4BCZ/p1738104905279239?thread_ts=1728456483.518839&cid=C027KKE4BCZ) for more information

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error logging for async job processing by adding traceback information to warnings, improving visibility into potential issues during job orchestration.

### DIFF
--- a/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
@@ -437,10 +437,10 @@ class AsyncJobOrchestrator:
                 yield from self._process_running_partitions_and_yield_completed_ones()
                 self._wait_on_status_update()
             except Exception as exception:
+                LOGGER.warning(
+                    f"Caught exception that stops the processing of the jobs: {exception}"
+                )
                 if self._is_breaking_exception(exception):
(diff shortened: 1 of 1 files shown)


maxi297:
@coderabbitai I'm good with your suggestion [here](https://github.com/airbytehq/airbyte-python-cdk/pull/307#pullrequestreview-2587988079). Can you make the change?

maxi297:
@coderabbitai yes, please commit the change

bazarnov (changes requested):
I believe the error type we try to catch is too general, we have to have a way to define this error type into a specific one at least or catch both specific and general like overall

maxi297:
Is the concern that we log too much? The user was having an issue and was not seeing the issue with the current logging solution.

bazarnov:
No, i believe we could have a specific case to catch first than catch the general Exception so we start having the errors base we might face in the future, instead of just Expection. This is basically a part of the re-factoring so no blocking you and will do this in my PR

bazarnov (approved):

PR #310, opened 2025-02-02
fix(concurrent cdk): Properly call set_initial_state() on the cursor that is initialized on the ClientSideIncrementalRecordFilterDecorator

brianjlai:
## What

While I was scoping some of the custom cursor deprecation issues, I noticed that our client side semi incremental filtering might not be working correctly on the CDK. This issue fixes how we assign state to the cursor so that it filters records correctly

## How

The issue arises because in `model_to_component_factory.py`, we actual instantiate a separate `DatetimeBasedCursor` or any cursor when we do client side filtering. See this block of code https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1547-L1569 .

Because these are two separate instances, we need to call `set_initial_state()` on both. It feels a little odd to have to do this twice, but I was unsure if we made this separate for a reason so I wanted to keep the flow as close to before as possible

Tested against `source-chargebee` + CDK unit tests

## Summary by CodeRabbit

- **Refactor**
  - Improved the initialization process for managing state in incremental processing so that streams handle state more reliably.
- **Tests**
  - Added new test coverage to ensure that streams using incremental processing correctly maintain their cursor state.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -475,10 +475,21 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
             # Also a temporary hack. In the legacy Stream implementation, as part of the read,
             # set_initial_state() is called to instantiate incoming state on the cursor. Although we no
             # longer rely on the legacy low-code cursor for concurrent checkpointing, low-code components
-            # like StopConditionPaginationStrategyDecorator and ClientSideIncrementalRecordFilterDecorator
-            # still rely on a DatetimeBasedCursor that is properly initialized with state.
+            # like StopConditionPaginationStrategyDecorator still rely on a DatetimeBasedCursor that is
+            # properly initialized with state.
             if retriever.cursor:
                 retriever.cursor.set_initial_state(stream_state=stream_state)
+
+            # Similar to above, the ClientSideIncrementalRecordFilterDecorator cursor is a separate instance
+            # from the one initialized on the SimpleRetriever, so it also must also have state initialized
+            # for semi-incremental streams using is_client_side_incremental to filter properly
+            if isinstance(retriever.record_selector, RecordSelector) and isinstance(
+                retriever.record_selector.record_filter, ClientSideIncrementalRecordFilterDecorator
+            ):
+                retriever.record_selector.record_filter._cursor.set_initial_state(
+                    stream_state=stream_state
+                )  # type: ignore  # After non-concurrent cursors are deprecated we can remove these cursor workarounds
+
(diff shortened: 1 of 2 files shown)


maxi297 (approved):
That's a good point. I'm fine with this fix for now. Soon we should have a plan to remove declarative cursors. Maybe it'll be easier when we won't use `DeclarativeStream` anymore?

Also note that [this](https://github.com/airbytehq/airbyte-python-cdk/blob/ef97304f45f3c54ac3f439dd45471aeb8272ea0c/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1561-L1569) seems weird to me as I would expect to have `DatetimeBasedCursor` too as part of the `isinstance`. Should we update this?

brianjlai:
> That's a good point. I'm fine with this fix for now. Soon we should have a plan to remove declarative cursors. Maybe it'll be easier when we won't use DeclarativeStream anymore?

I think it will get easier, but I worry there. may still be some gaps during deprecation where we find out we're using `DatetimeBasedCursor` niche places like this where the new ConcurrentCursor isn't compatible. But that's a little out of scope for us right here.

> Also note that [this](https://github.com/airbytehq/airbyte-python-cdk/blob/ef97304f45f3c54ac3f439dd45471aeb8272ea0c/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1561-L1569) seems weird to me as I would expect to have DatetimeBasedCursor too as part of the isinstance else we create a new cursor where set_initial_state won't be called. Should we update this?

My guess here is that its actually synonymous with this code within `_merge_stream_slicers()` where we end up creating a DatetimeBasedCursor here:

https://github.com/airbytehq/airbyte-python-cdk/blob/126e2339add6c8cab2b9e9e8218645e96c6936c4/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1733-L1754

So if it doesn't match those either `(PerPartitionWithGlobalCursor, GlobalSubstreamCursor)`, we end up just recreating a `DatetimeBasedCursor` (or custom cursor) on the last line. I think there's probably a good argument for just simplifying the block to:

```
if combined_slicers and not isinstance(combined_slicers, supported_slicers):
  supported_slicers = (
                  DatetimeBasedCursor,
                  GlobalSubstreamCursor,
                  PerPartitionWithGlobalCursor,
              )
  raise ValueError(
      "Unsupported Slicer is used. PerPartitionWithGlobalCursor should be used here instead"
  )
client_side_incremental_sync = {"cursor": combined_slicers}
```
I think I'll investigate this separately because its a good callout, but I want to get this current PR in to unblock the community dev work we're slating in the coming days which depends on this.

PR #314, opened 2025-02-05
feat(low-code): Add API Budget

tolik0:


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1365,6 +1371,207 @@ definitions:
...
+      next_reset_ts:
+        title: Next Reset Timestamp
+        description: The timestamp when the rate limit will reset.
+        type: string
+        format: date-time
+      period:
+        title: Period
+        description: The time interval for the rate limit window.
+        type: string
+        format: duration
```
maxi297, line 1471: What is this format? I'm not sure I've seen this elsewhere

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -40,6 +40,12 @@ properties:
     "$ref": "#/definitions/Spec"
   concurrency_level:
     "$ref": "#/definitions/ConcurrencyLevel"
+  api_budget:
+    title: API Budget
+    description: Defines how many requests can be made to the API in a given time frame. This field accepts either a generic APIBudget or an HTTP-specific configuration (HTTPAPIBudget) to be applied across all streams.
```
maxi297, line 45: Can we make it explicit on what is the scope here? For example, it seems like customer components won't be covered by the API Budget. Are there other cases that are out of scope?

maxi297 (commented):
The code looks good to me but I have some questions to make sure I have all the context. I'll re-review the code once I have this context

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1365,6 +1371,207 @@ definitions:
...
+      - matchers
+    properties:
+      type:
+        type: string
+        enum: [FixedWindowCallRatePolicy]
+      next_reset_ts:
+        title: Next Reset Timestamp
+        description: The timestamp when the rate limit will reset.
+        type: string
+        format: date-time
```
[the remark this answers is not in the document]
tolik0: I was initially considering setting the end of the day for some endpoints, but it doesn’t make sense since we would retrieve this value with the first request anyway. I'll delete this field from the schema.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: It uses Pydantic's internal parsing for duration. I'll remove the format to align it with other fields of the same type.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1365,6 +1371,207 @@ definitions:
...
+      - type
+      - next_reset_ts
+      - period
+      - call_limit
+      - matchers
+    properties:
+      type:
+        type: string
+        enum: [FixedWindowCallRatePolicy]
+      next_reset_ts:
```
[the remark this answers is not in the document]
tolik0: Do you mean next_reset_timestamp? But do we need to expose this field?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1365,6 +1371,207 @@ definitions:
...
+        type: string
+        default: "ratelimit-remaining"
+      status_codes_for_ratelimit_hit:
+        title: Status Codes for Rate Limit Hit
+        description: List of HTTP status codes that indicate a rate limit has been hit.
+        type: array
+        items:
+          type: integer
+        default: [429]
+      maximum_attempts_to_acquire:
```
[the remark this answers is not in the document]
tolik0: This represents the number of attempts to acquire before making the call. I also think we don’t need to expose this field. Should I remove it?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: Everything [that uses `http_requester` is covered](https://github.com/airbytehq/airbyte-python-cdk/pull/314/files/7be98423518c975e672629abbd4cb063048e55d2#diff-1d9bc4ca384e5c00867da05e9db9d919aba908c59bb73aab83883b9ed20def05R1913-R1975). I will update the description to make this explicit.

artem1205 (approved):
LGTM!

nit: could you please:
1. add more info to our log message so we could know which `stream`/`endpoint`/`rule` is applied/affected.
image
2. add docs/examples on how to use it?

tolik0:
@artem1205 I’ve added debug logs with request details and policies, as well as documentation.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: Updated description with the scope

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: Fixed

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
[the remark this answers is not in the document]
tolik0: Removed from the manifest

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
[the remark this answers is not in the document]
tolik0: Removed from the manifest

airbyte_cdk/sources/streams/call_rate.py
```diff
@@ -399,24 +495,17 @@ def update_from_response(self, request: Any, response: Any) -> None:
         """
 
 
+@dataclass
```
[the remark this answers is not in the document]
tolik0: Reverted this change

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
maxi297: I still see the format. Am I missing something?

airbyte_cdk/sources/streams/call_rate.py
```diff
@@ -415,6 +545,23 @@ def __init__(
...
+        try:
+            # If the request is already a PreparedRequest, it should have a URL.
+            if isinstance(request, requests.PreparedRequest):
+                endpoint = request.url
+            # If it's a requests.Request, we call prepare() to extract the URL.
+            elif isinstance(request, requests.Request):
+                prepared = request.prepare()
+                endpoint = prepared.url
+        except Exception as e:
+            logger.debug(f"Error extracting endpoint: {e}")
```
maxi297, line 560: Just for my understanding: In which case can this happen?

airbyte_cdk/sources/streams/call_rate.py
```diff
@@ -496,7 +653,7 @@ def __init__(
         self,
         ratelimit_reset_header: str = "ratelimit-reset",
         ratelimit_remaining_header: str = "ratelimit-remaining",
-        status_codes_for_ratelimit_hit: tuple[int] = (429,),
+        status_codes_for_ratelimit_hit: Union[tuple[int], list[int]] = (429,),
```
maxi297, line 656: nit: it feels like it would be easier to maintain have only one type here

airbyte_cdk/sources/streams/call_rate.py
```diff
@@ -159,6 +161,113 @@ def __call__(self, request: Any) -> bool:
...
+                return False
+
+        # Parse the URL.
+        parsed_url = parse.urlsplit(prepared_request.url)
+        # Reconstruct the base: scheme://netloc
+        request_url_base = f"{str(parsed_url.scheme)}://{str(parsed_url.netloc)}"
+        # The path (without query parameters)
+        request_path = str(parsed_url.path).rstrip("/")
+
+        # If a base URL is provided, check that it matches.
```
maxi297, line 240: This seems to be copy/paste of `HttpRequestMatcher` which makes me a bit worried that it'll be hard to maintain. Can we re-implement `HttpRequestMatcher` to use `HttpRequestRegexMatcher` under the hood?

maxi297 (commented):
Small things for maintenance reasons but the behavior makes sense to me

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: Yes, sorry, I deleted it from the Rates class but not from here. Fixed now

airbyte_cdk/sources/streams/call_rate.py
tolik0: Fixed to use one type

airbyte_cdk/sources/streams/call_rate.py
tolik0: This could happen if the request object is malformed (e.g., missing attributes) or if an unexpected object type is passed instead of a valid `requests.Request`. However, I don’t expect such an error under normal conditions.

maxi297: So the `prepare` could generate the issue, got it thanks!

airbyte_cdk/sources/streams/call_rate.py
tolik0: HttpRequestMatcher is used in Salesforce, which is why I initially split them. However, I see the maintenance concern and will try to reimplement it using HttpRequestRegexMatcher under the hood.

tolik0: Refactored

maxi297 (approved):
It's a bit of a speedrun review but I'm good with what I saw for the refactor of the HTTP matcher. Thanks for the refactor! I'm sure it'll be easier to manage.

PR #316, opened 2025-02-05
fix: (low code)run state migrations for concurrent streams

darynaishchenko:
### What
related to klaviyo campaigns stream to low code https://github.com/airbytehq/airbyte/pull/51551
Concurrent framework gets the stream_state from the state manager and not the DeclarativeStream, where state migrations are called.
### How
Added `state_migration.migrate(stream_state)` to the concurrent declarative source. Added unit test.

## Summary by CodeRabbit

- **New Features**
  - Enhanced stream state management with automated migration to ensure consistent and up-to-date processing.
  - Introduced a new class for custom state migration, providing structured handling of stream states.
  - Added support for applying state migrations during cursor creation for improved flexibility.

- **Tests**
  - Added new tests to validate the state migration process, ensuring reliable data handling and improved system stability.
  - Introduced additional tests for state migrations during cursor creation from both datetime-based and per-partition streams.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -3,7 +3,7 @@
 #
 
 import logging
-from typing import Any, Generic, Iterator, List, Mapping, Optional, Tuple, MutableMapping
+from typing import Any, Generic, Iterator, List, Mapping, MutableMapping, Optional, Tuple
 
 from airbyte_cdk.models import (
     AirbyteCatalog,
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -3,7 +3,7 @@
 #
 
 import logging
-from typing import Any, Generic, Iterator, List, Mapping, Optional, Tuple
+from typing import Any, Generic, Iterator, List, Mapping, Optional, Tuple, MutableMapping
 
 from airbyte_cdk.models import (
     AirbyteCatalog,
@@ -224,10 +224,7 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
                     stream_state = self._connector_state_manager.get_stream_state(
                         stream_name=declarative_stream.name, namespace=declarative_stream.namespace
                     )
-                    for state_migration in declarative_stream.state_migrations:
-                        if state_migration.should_migrate(stream_state):
-                            # The state variable is expected to be mutable but the migrate method returns an immutable mapping.
-                            stream_state = dict(state_migration.migrate(stream_state))
+                    stream_state = self._migrate_state(declarative_stream, stream_state)
 
                     retriever = self._get_retriever(declarative_stream, stream_state)
 
@@ -335,10 +332,7 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
                     stream_state = self._connector_state_manager.get_stream_state(
                         stream_name=declarative_stream.name, namespace=declarative_stream.namespace
                     )
-                    for state_migration in declarative_stream.state_migrations:
-                        if state_migration.should_migrate(stream_state):
-                            # The state variable is expected to be mutable but the migrate method returns an immutable mapping.
-                            stream_state = dict(state_migration.migrate(stream_state))
+                    stream_state = self._migrate_state(declarative_stream, stream_state)
 
                     partition_router = declarative_stream.retriever.stream_slicer._partition_router
 
@@ -530,3 +524,14 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
                 if stream.stream.name not in concurrent_stream_names
             ]
         )
+
+    @staticmethod
+    def _migrate_state(
+        declarative_stream: DeclarativeStream, stream_state: MutableMapping[str, Any]
+    ) -> MutableMapping[str, Any]:
+        for state_migration in declarative_stream.state_migrations:
+            if state_migration.should_migrate(stream_state):
+                # The state variable is expected to be mutable but the migrate method returns an immutable mapping.
+                stream_state = dict(state_migration.migrate(stream_state))
+
+        return stream_state
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -226,7 +226,8 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
                     )
                     for state_migration in declarative_stream.state_migrations:
                         if state_migration.should_migrate(stream_state):
-                            stream_state = state_migration.migrate(stream_state)
+                            # The state variable is expected to be mutable but the migrate method returns an immutable mapping.
+                            stream_state = dict(state_migration.migrate(stream_state))
 
                     retriever = self._get_retriever(declarative_stream, stream_state)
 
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -224,6 +224,10 @@ def _group_streams(
                     stream_state = self._connector_state_manager.get_stream_state(
                         stream_name=declarative_stream.name, namespace=declarative_stream.namespace
                     )
+                    for state_migration in declarative_stream.state_migrations:
```
maxi297, line 227: nit: Can we extract this in a method so that if we have to change it, we don't forget to change it at the other place? Something like:
```
def _get_state(self, stream):
    for state_migration in declarative_stream.state_migrations:
        if state_migration.should_migrate(stream_state):
            # The state variable is expected to be mutable but the migrate method returns an immutable mapping.
            stream_state = dict(state_migration.migrate(stream_state))
    return stream_state
```

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -1231,6 +1231,235 @@ def test_read_with_concurrent_and_synchronous_streams_with_sequential_state():
...
+            ),
+        ),
+    ]
+    with HttpMocker() as http_mocker:
+        _mock_party_members_requests(http_mocker, party_members_slices_and_responses)
+        messages = list(
+            source.read(logger=source.logger, config=_CONFIG, catalog=catalog, state=state)
+        )
+    final_state = get_states_for_stream(stream_name="party_members", messages=messages)
+    assert state not in final_state
```
maxi297, line 1460: I'm not sure about this test. It seems to test that before and after a read, the state is different. This logic seems to make sense even without the state migration. Should we scope this test more and have something like:
* Instantiate a source that has a stream with a state migration
* call source._group_streams
* validate that the grouped stream that is concurrent also had a state migration when accessing `stream.cursor.state`

Right now, my fear is that `create_concurrent_cursor_from_datetime_based_cursor` does not receive the migrated state and therefore this change is not sufficient

maxi297 (commented):
I think there is something missing in my explanation during our sync. This PR updates the state that is used on the retriever (which seems to cover [this hack](https://github.com/airbytehq/airbyte-python-cdk/blob/08198acb9d924a15cad14826cfc2d688e5b568b8/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L484-L488)) but not the one on the cursor. I'm not even sure the hack needs to be maintained at that point. @brianjlai what do you think?

brianjlai:
> I think there is something missing in my explanation during our sync. This PR updates the state that is used on the retriever (which seems to cover [this hack](https://github.com/airbytehq/airbyte-python-cdk/blob/08198acb9d924a15cad14826cfc2d688e5b568b8/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L484-L488)) but not the one on the cursor. I'm not even sure the hack needs to be maintained at that point. @brianjlai what do you think?

Yeah that sounds correct. I think the missing piece here is that we need to also run a state migration after we read in state from `ModelToComponentFactory.create_concurrent_cursor_from_datetime_based_cursor()`.

That will also require that we pass in the migrations to the model factory. Or an idea that i kind of like is that when we parse the manifest into the DatetimeBasedCursorModel, we also append the state migrations to the pydantic model. And we have it available to run from `create_concurrent_cursor_from_datetime_based_cursor()`

unit_tests/sources/declarative/test_concurrent_declarative_source.py
darynaishchenko: updated test

darynaishchenko:
@maxi297 @brianjlai 
Updated pr with the changes to `create_concurrent_cursor_from_datetime_based_cursor`. Could you please take a look?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1236,6 +1243,7 @@ def create_concurrent_cursor_from_perpartition_cursor(
                 stream_namespace=stream_namespace,
                 config=config,
                 message_repository=NoopMessageRepository(),
+                stream_state_migrations=stream_state_migrations,
```
maxi297, line 1246: I don't know what we should expect here but with this implementation, we can only apply a migration to states within the per partition state i.e. given this example:
```
                {
                    "states": [
                        {
                            "partition": {
                                "partition_key": "value"
                            },
                            "cursor": {
                                "last_updated": "2023-05-27T00:00:00Z"
                            }
                        }
                    ],
                    "state": {
                        "last_updated": "2023-05-27T00:00:00Z"
                    },
                    lookback_window: 10,
                    "parent_state": {
                        "parent_stream_name": {
                            "last_updated": "2023-05-27T00:00:00Z"
                        }
                    }
                }
```
... we can only migrate the value within `states.X.cursor`. It feels like it would be more flexible if the migration applies on the top level and that the state migration applies to fields on the root level like `lookback_window` and `parent_state` too for example. WDYT?

maxi297 (commented):
One comment on the scope of the migration for perpartition. I think we can make it even more useful

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
darynaishchenko: @maxi297 , I'm not sure I understood your comment. `stream_state_migrations` is a list of StateMigration objects: [example](https://github.com/airbytehq/airbyte-python-cdk/pull/316/files#diff-175d190dfbd6dc4e669155da03423bb985c565c5a12b29ceb9170f9c4a47f7d2R3333) Looks like you want to change migrate method inside StateMigration implementation, right?

maxi297: We had a sync and are aligned on this. @darynaishchenko will work on the change ❤️

darynaishchenko: updated pr with this change

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -224,6 +224,7 @@ def _group_streams(
                     stream_state = self._connector_state_manager.get_stream_state(
                         stream_name=declarative_stream.name, namespace=declarative_stream.namespace
                     )
+                    stream_state = self._migrate_state(declarative_stream, stream_state)
```
maxi297, line 227: nit: can we move the `self._migrate_state` call inside `_get_retriever as we only do that for the retriever part?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -953,6 +954,11 @@ def create_concurrent_cursor_from_datetime_based_cursor(
             if "stream_state" not in kwargs
             else kwargs["stream_state"]
         )
+        if stream_state_migrations:
```
maxi297, line 957: nit: can we extract this to avoid duplication in `model_to_component_factory.py`?

maxi297 (approved):
Just small nits but I'm good with this one. Thanks Daryna!

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
tolik0: I think we should not pass the state migration to the cursor for partitions because this state has already been migrated with the state migration for the per-partition cursor. Additionally, since it's the same migration, there is no way to distinguish whether it is migrating the root-level state or the state for a specific partition.

darynaishchenko: I removed passing `state_migrations` to the `create_concurrent_cursor_from_perpartition_cursor`, but left migration logic inside the method, so in the case when state not migrated previously via `create_concurrent_cursor_from_datetime_based_cursor` it can be used. 
Tested with klaviyo, worked as expected.

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
darynaishchenko: for me it's more logic for `_connector_state_manager `, but if we don't want to change its method declaration I prefer to override `stream_state ` variable as we will be sure that next using of `stream_state` contains migrated and actual state.

tolik0 (approved):

PR #318, opened 2025-02-06
feat(low-code cdk): add StateDelegatingStream

lazebnyi:
## What 

Fixed: https://github.com/airbytehq/airbyte-internal-issues/issues/11589

This PR adds the `StateDelegatingStream` component to the low-code CDK.

This component determines the stream YAML description to be used for synchronization based on the state value passed to the source.  

## Summary by CodeRabbit

- **New Features**
	- Introduced an experimental state delegating stream feature that enables flexible data retrieval workflows, allowing users to configure streams for both full refresh and incremental sync based on dynamic state conditions.
	- Enhanced orchestration and caching behavior offers improved reliability and adaptability in stream processing.

- **Tests**
	- Added comprehensive tests to ensure robust performance of the new state-based data retrieval functionality across different scenarios.


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3150,6 +3151,42 @@ definitions:
...
+      - incremental_retriever
+    properties:
+      type:
+        type: string
+        enum: [ StateDelegatingRetriever ]
+      full_refresh_no_slice_in_params:
+        description: If set to true, a single slice will be used and its request options will be ignored when sending requests.
+        type: boolean
+        default: false
+      full_refresh_ignore_min_max_datetime:
```
maxi297, line 3169: In which context do we need this for Stripe?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1790,85 +1799,70 @@ def _build_stream_slicer_from_partition_router(
...
-                    partition_router=stream_slicer,
-                    stream_cursor=cursor_component,
-                )
+                ),
+                partition_router=stream_slicer,
+                stream_cursor=cursor_component,
+            )
         elif model.incremental_sync:
             if model.retriever.type == "AsyncRetriever":
-                if model.incremental_sync.type != "DatetimeBasedCursor":
```
maxi297, line 1864: 🙏 thanks for the cleanup

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -206,10 +210,34 @@ def _group_streams(
...
                 == "AsyncRetriever"
+                or name_to_stream_mapping[declarative_stream.name]["retriever"]["type"]
+                == "StateDelegatingRetriever"
             ):
                 incremental_sync_component_definition = name_to_stream_mapping[
                     declarative_stream.name
-                ].get("incremental_sync")
+                ].get("incremental_sync", {})
+
+                if (
```
maxi297, line 220: It seems like this logic is duplicated from [here](https://github.com/airbytehq/airbyte-python-cdk/pull/318/files#diff-1d9bc4ca384e5c00867da05e9db9d919aba908c59bb73aab83883b9ed20def05R1818). Do we need both? If so, can we avoid duplicating it?

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -206,10 +210,34 @@ def _group_streams(
...
+                    .get("retriever", {})
+                    .get("full_refresh_ignore_min_max_datetime", False)
+                    and incremental_sync_component_definition
+                ):
+                    incremental_sync_component_definition["start_datetime"]["max_datetime"] = None
+                    incremental_sync_component_definition["start_datetime"]["min_datetime"] = None
+                    incremental_sync_component_definition["end_datetime"]["max_datetime"] = None
+                    incremental_sync_component_definition["end_datetime"]["min_datetime"] = None
 
                 partition_router_component_definition = (
```
maxi297, line 242: We have a very interesting case with @tolik0 where we would need to have a case with source-monday where the full refresh is just an incremental sync and the incremental has a partition router. Surprisingly, the fact that it would be different state format (full refresh is a simple `{<cursor_field>: <cursor_value>}` but the incremental one is per partition) would be fine because we only case about the parent state in this case and the migration from "simple" to "per partition" sets `<cursor_value>` to the parent state. However in order for that to work, we would need to check if there is a state and if so if there is a partition router on the incremental retriever. Do you think it would be possible?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3150,6 +3151,42 @@ definitions:
...
+    type: object
+    required:
+      - type
+      - full_refresh_retriever
+      - incremental_retriever
+    properties:
+      type:
+        type: string
+        enum: [ StateDelegatingRetriever ]
+      full_refresh_no_slice_in_params:
```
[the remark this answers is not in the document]
lazebnyi: Yes, you are right. But for some endpoints (e.g., `/accounts`), we have restrictions on which query parameters can be passed. If we include slice parameters in the query, it will trigger an error.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
lazebnyi: This flag is needed because some endpoints have limitations for incremental reads (e.g., events can only be read up to 30 days in the past).

To avoid restricting the start/end date during a full refresh, we can use this flag.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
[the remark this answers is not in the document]
maxi297: Is this an issue because of RequestOptions only? If we do string interpolation, are we good with this? 

I'm not saying we should do that, I'm just trying to understand the parameters of the problem

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
lazebnyi: Yes, this is because we use these values multiple times—both when creating slicers in the constructor and when creating the partitions class inside `_group_streams`.

More details in the loom in the PR description.

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
lazebnyi: After reverting to the two-retriever instances approach, I think it could work.

Is this stream what you were describing?
**https://github.com/airbytehq/airbyte/blob/fe3ea86ec1508622b5e8e261586cb4b66b89eafd/airbyte-integrations/connectors/source-stripe/source_stripe/manifest.yaml#L702**

unit_tests/sources/declarative/test_state_delegating_stream.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Test full data retrieval (without state)
+        full_records = get_records(source, _CONFIG, configured_catalog)
+        expected_full = [
+            {"id": 1, "name": "item_1", "updated_at": "2024-07-13"},
+            {"id": 2, "name": "item_2", "updated_at": "2024-07-13"},
+        ]
+        assert expected_full == full_records
+
+        # Test incremental data retrieval (with state)
```
maxi297, line 230: nit: can we split this in two tests? It is interesting feedback for the users to see that full refresh is failing but not incremental. With the current test setup, it is not possible to know

maxi297 (approved):
One small nit but the rest make sense to me. Thanks for this change

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2950,7 +3000,14 @@ def _create_message_repository_substream_wrapper(
                 self._evaluate_log_level(self._emit_connector_builder_messages),
             ),
         )
-        return substream_factory._create_component_from_model(model=model, config=config)
+        child_state = (
+            self._connector_state_manager.get_stream_state(kwargs.get("stream_name", ""), None)
```
maxi297, line 3004: I had a brainfart after approving. 

It feels like this would be used to create the parent stream because I think this should be named as `parent_state`, right? It is quite interesting because before, we would assume that the state is only needed when calling [set_initial_state](https://github.com/airbytehq/airbyte-python-cdk/blob/5d82e1485fdba520102f5a618a8de9debd321ac5/airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py#L241). Now, we need to know the state before we actually instantiate this. I think for now, I would consider this case out-of-scope if we don't need it for source-stripe and I would add an exception somewhere (since the `StateDelegatingStreamModel` is only present in the config and there is no Python equivalent, I would assume it needs to be in `model_to_component_factory`) and we can try to support this case better later on. WDYT?

unit_tests/sources/declarative/test_state_delegating_stream.py
lazebnyi: Done

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: After a sync with @lazebnyi: It turns out we need this because we have streams like `transfer_reversals` which have a parent that actually have state delegation. Hence, we need to support that. The interesting thing is that we don't need to actually retrieve the parent state here. We just need to know that there is a state. Hence, we can have any dummy state as we don't use the content of the state but just the fact that it exists

PR #322, opened 2025-02-07
fix(low-code): Add parent state migration from global state

tolik0:
## Summary by CodeRabbit

- **Refactor**
	- Improved state migration during incremental synchronization to ensure only essential state data is retained, enhancing overall consistency.
- **Tests**
	- Expanded the test suite with new scenarios and parameterized cases to validate incremental state handling and robust error management.
	- Updated expected outputs in existing tests to align with new state management logic.

### DIFF
--- a/airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
+++ b/airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
@@ -299,23 +299,33 @@ class SubstreamPartitionRouter(PartitionRouter):
 
     def _migrate_child_state_to_parent_state(self, stream_state: StreamState) -> StreamState:
         """
-        Migrate the child stream state to the parent stream's state format.
+        Migrate the child or global stream state into the parent stream's state format.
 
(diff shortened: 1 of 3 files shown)


maxi297 (approved):
I fear that this might not be 100% true in the following case:
* T0: HTTP request for parent stream. To be synced are parent_1 and parent_2
* T1: Parent stream emit parent_1
* T2: HTTP request for child stream on slice parent_1
* T3: parent_3 is also updated which would now make it available in the parent stream HTTP request
* T4: a new record is added on child stream for parent_2
* T5: finished processing parent_1 on child level
* T6: process child stream for slice parent_2

In this case, we would set the parent state as T4 and the update on parent_3 would not be picked up. That being said, the "{<cursor field>: <cursor value>}" format was not safe for very similar reasons so I'm good with this as a migration.

tolik0:
@maxi297 I agree. Ideally, we should apply the lookback window to the parent stream as well. However, this migration is only meant for transitioning from the legacy format, so there’s no need to overcomplicate the logic since the lookback window won’t be present in that case.

PR #324, opened 2025-02-07
feat(OAuth): allow for access_token without expiration from the API

maxi297:
## What

See https://github.com/airbytehq/airbyte-enterprise/pull/82#discussion_r1943603994 for more details

## How

Allow for initializing an access_token while provided a date in the future for refresh.

Note that if the token_expiry_date is not far enough in the future, the connector might refresh the token on every call. It is the responsibility of the connector to put a date that is far enough in the future to avoid this.

## Summary by CodeRabbit

- **Bug Fixes**  
  - Enhanced the authentication flow to more reliably handle token expiration and refresh scenarios, ensuring smoother and more consistent access.  

- **Tests**  
  - Expanded test coverage to validate the improved token management, including scenarios with no access token but a future expiry date, helping maintain robust and dependable authentication behavior.

### DIFF
--- a/unit_tests/sources/declarative/auth/test_oauth.py
+++ b/unit_tests/sources/declarative/auth/test_oauth.py
@@ -301,6 +301,7 @@ class TestOauth2Authenticator:
             client_id="{{ config['client_id'] }}",
             client_secret="{{ config['client_secret'] }}",
             token_expiry_date=timestamp,
+            access_token_value="some_access_token",
             refresh_token="some_refresh_token",
             config={
                 "refresh_endpoint": "refresh_end",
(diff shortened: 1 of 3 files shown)


unit_tests/sources/declarative/auth/test_oauth.py
```diff
@@ -301,6 +301,7 @@ def test_initialize_declarative_oauth_with_token_expiry_date_as_timestamp(
             client_id="{{ config['client_id'] }}",
             client_secret="{{ config['client_secret'] }}",
             token_expiry_date=timestamp,
+            access_token_value="some_access_token",
```
maxi297, line 304: I had to change this because if the access_token is not initialized, the `get_token_expiry_date ` will be datetime.min to force a refresh.

The use case in this test didn't make sense to me as there were no access token but there was an expiry date. I'm not sure in which case this is possible.

lazebnyi (approved):
lgtm

PR #326, opened 2025-02-07
chore(decoder): clean decoders and make csvdecoder available

maxi297:
## What

https://github.com/airbytehq/airbyte-internal-issues/issues/11616

This is a breaking change but only for an experimental component or one that is only used in source-amplitude so I'm fine keeping this a minor change.

## Summary by CodeRabbit

- **New Features**  
  - Introduced a unified decoding framework supporting multiple data types (e.g., CSV, JSON, and compressed formats) for enhanced flexibility.
  - Added new decoders for handling CSV and Gzip decoding.

- **Refactor**  
  - Streamlined the data extraction process by consolidating redundant decoding components and improving handling for both streamed and non-streamed responses.

- **Tests**  
  - Expanded test coverage to validate improved response processing and enhanced memory usage efficiency under various conditions.
  - Added new test cases for the `CompositeRawDecoder` to ensure correct behavior with consumed and non-streamed responses.
  - Updated tests to reflect changes in decoder implementations and removed obsolete tests.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2984,6 +2984,7 @@ definitions:
         description: Component decoding the response so records can be extracted.
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
+          - "$ref": "#/definitions/CsvDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
@@ -3142,6 +3143,7 @@ definitions:
         description: Component decoding the response so records can be extracted.
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
+          - "$ref": "#/definitions/CsvDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
@@ -3152,6 +3154,7 @@ definitions:
         description: Component decoding the download response so records can be extracted.
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
+          - "$ref": "#/definitions/CsvDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2985,6 +2985,7 @@ definitions:
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
           - "$ref": "#/definitions/CsvDecoder"
+          - "$ref": "#/definitions/GzipDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
@@ -3144,6 +3145,7 @@ definitions:
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
           - "$ref": "#/definitions/CsvDecoder"
+          - "$ref": "#/definitions/GzipDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
@@ -3155,6 +3157,7 @@ definitions:
         anyOf:
           - "$ref": "#/definitions/CustomDecoder"
           - "$ref": "#/definitions/CsvDecoder"
+          - "$ref": "#/definitions/GzipDecoder"
           - "$ref": "#/definitions/JsonDecoder"
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2997,12 +2997,12 @@ definitions:
     type: object
     required:
       - type
-      - inner_decoder
+      - decoder
     properties:
       type:
         type: string
         enum: [GzipDecoder]
-      inner_decoder:
+      decoder:
         anyOf:
           - "$ref": "#/definitions/CsvDecoder"
           - "$ref": "#/definitions/GzipDecoder"
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2133,23 +2133,6 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
-  GzipJsonDecoder:
-    title: GzipJson Decoder
-    description: Use this if the response is Gzip compressed Json.
-    type: object
-    additionalProperties: true
-    required:
-      - type
-    properties:
-      type:
-        type: string
-        enum: [GzipJsonDecoder]
-      encoding:
-        type: string
-        default: utf-8
-      $parameters:
-        type: object
-        additionalProperties: true
   ZipfileDecoder:
     title: Zipfile Decoder
     description: Decoder for response data that is returned as zipfile(s).
@@ -3006,7 +2989,6 @@ definitions:
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
           - "$ref": "#/definitions/XmlDecoder"
-          - "$ref": "#/definitions/GzipJsonDecoder"
           - "$ref": "#/definitions/CompositeRawDecoder"
           - "$ref": "#/definitions/ZipfileDecoder"
       $parameters:
@@ -3206,7 +3188,6 @@ definitions:
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
           - "$ref": "#/definitions/XmlDecoder"
-          - "$ref": "#/definitions/GzipJsonDecoder"
           - "$ref": "#/definitions/CompositeRawDecoder"
           - "$ref": "#/definitions/ZipfileDecoder"
       download_decoder:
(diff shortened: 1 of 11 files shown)


airbyte_cdk/sources/declarative/decoders/json_decoder.py
```diff
@@ -10,48 +10,39 @@
...
-        return False
+        return self._decoder.is_stream_response()
 
     def decode(
         self, response: requests.Response
     ) -> Generator[MutableMapping[str, Any], None, None]:
         """
         Given the response is an empty string or an emtpy list, the function will return a generator with an empty mapping.
         """
+        has_yielded = False
```
maxi297, line 36: The default behavior of this decoder was a bit weird so I decided to keep it not to push the weird logic in `JsonParser`

airbyte_cdk/sources/declarative/decoders/json_decoder.py
```diff
@@ -69,43 +60,3 @@ def decode(
...
-        self, response: requests.Response
-    ) -> Generator[MutableMapping[str, Any], None, None]:
-        # TODO???: set delimiter? usually it is `\n` but maybe it would be useful to set optional?
-        #  https://github.com/airbytehq/airbyte-internal-issues/issues/8436
-        for record in response.iter_lines():
-            yield orjson.loads(record)
-
-
-@dataclass
-class GzipJsonDecoder(JsonDecoder):
```
maxi297, line 95: Removes because it wasn't used anyway. We can introduce this for source-amazon-seller-partner though

airbyte_cdk/sources/declarative/decoders/json_decoder.py
```diff
@@ -10,47 +10,39 @@
...
 
 
-@dataclass
 class JsonDecoder(Decoder):
     """
     Decoder strategy that returns the json-encoded content of a response, if any.
     """
 
-    parameters: InitVar[Mapping[str, Any]]
+    def __init__(self, parameters: Mapping[str, Any]):
```
natikgadzhi, line 24: Nit: a comment on what `_decoder` is would be really good to have a few months form now.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1720,18 +1701,13 @@ class Config:
         extra = Extra.allow
 
     type: Literal["ZipfileDecoder"]
-    parser: Union[GzipParser, JsonParser, JsonLineParser, CsvParser] = Field(
+    decoder: Union[CsvDecoder, GzipDecoder, JsonDecoder, JsonlDecoder] = Field(
```
natikgadzhi, line 1704: Asking to learn: why is it that `GzipDecoder` has `innner_decoder`, but `ZipfileDecoder` has just `decoder`? Should these names be consistent?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2081,33 +2057,30 @@ def create_iterable_decoder(
...
+        if isinstance(model, JsonDecoderModel):
+            # Note that the logic is a bit different from the JsonDecoder as there is some legacy that is maintained to return {} on error cases
+            return JsonParser()
+        elif isinstance(model, JsonlDecoderModel):
+            return JsonLineParser()
+        elif isinstance(model, CsvDecoderModel):
+            return CsvParser(encoding=model.encoding, delimiter=model.delimiter)
+        elif isinstance(model, GzipDecoderModel):
+            return GzipParser(
+                inner_parser=ModelToComponentFactory._get_parser(model.inner_decoder, config)
```
natikgadzhi, line 2076: Just to make sure I'm not tripping: this calls the func recursively, but for the inner decoder, right?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3002,79 +2984,38 @@ definitions:
...
-      - type
-    properties:
-      type:
-        type: string
-        enum: [JsonLineParser]
-      encoding:
-        type: string
-        default: utf-8
-  CsvParser:
+          - "$ref": "#/definitions/CsvDecoder"
```
natikgadzhi, line 3007: Nothing stops us from shoving XmlDecoder here as well, but we're just not sure if there is a usecase for it?

natikgadzhi (approved):
I haven't looked into what that would mean for our connectors and published manifests — I think the only concern is around `GzipJsonDecoder` (not used?) and `CompositeRawDecoded` that I believe is used in a few spots, but it's a very simple manifest change to update, right?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
maxi297: Those four entries have an associated parser. [XmlDecoder](https://github.com/airbytehq/airbyte-python-cdk/blob/74631d8850510194d90faeddbb489ada5ad3cb8a/airbyte_cdk/sources/declarative/decoders/xml_decoder.py#L19) does not as it was implemented as a Decoder and not a Parser. Hence, it can't be supported in underlying decoders like in GzipDecoder

airbyte_cdk/sources/declarative/decoders/json_decoder.py
maxi297: Added a description in the class documentation

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
maxi297: We should probably have those name consistent. As we are currently doing the breaking change, I can probably break this as well

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: Exactly

maxi297 (commented):
Adding comments on the code review

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3002,79 +2984,38 @@ definitions:
...
+          - "$ref": "#/definitions/GzipDecoder"
```
pnilan, line 3008: Is supporting a nested GzipDecoder necessary?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3202,24 +3143,22 @@ definitions:
         description: Component decoding the response so records can be extracted.
         anyOf:
```
pnilan, line 3144: Add `GzipDecoder` here?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3202,24 +3143,22 @@ definitions:
...
           - "$ref": "#/definitions/JsonlDecoder"
           - "$ref": "#/definitions/IterableDecoder"
           - "$ref": "#/definitions/XmlDecoder"
-          - "$ref": "#/definitions/GzipJsonDecoder"
-          - "$ref": "#/definitions/CompositeRawDecoder"
           - "$ref": "#/definitions/ZipfileDecoder"
       download_decoder:
         title: Download Decoder
         description: Component decoding the download response so records can be extracted.
         anyOf:
```
pnilan, line 3155: Add `GzipDecoder` here?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3002,79 +2984,38 @@ definitions:
         description: Component decoding the response so records can be extracted.
         anyOf:
```
pnilan, line 2985: Add GzipDecoder here?

pnilan (approved):
Approved -- just want to confirm if we should add `GzipDecoder` as a "top level" decoder in the `SimpleRetriever` and `AsyncRetriever`.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
maxi297: I don't know if this is needed but I don't know if we want to prevent it

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
maxi297: Good point!

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
maxi297: Good point!

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2081,33 +2057,30 @@ def create_iterable_decoder(
...
+            # Note that the logic is a bit different from the JsonDecoder as there is some legacy that is maintained to return {} on error cases
+            return JsonParser()
+        elif isinstance(model, JsonlDecoderModel):
+            return JsonLineParser()
+        elif isinstance(model, CsvDecoderModel):
+            return CsvParser(encoding=model.encoding, delimiter=model.delimiter)
+        elif isinstance(model, GzipDecoderModel):
+            return GzipParser(
+                inner_parser=ModelToComponentFactory._get_parser(model.decoder, config)
+            )
```
artem1205, line 2077: I'm a little bit confused how this should work:

I tried to run test with these CDK changes in my original PR: https://github.com/airbytehq/airbyte/pull/48449 and it failed. 
I have a GzipDecoder with a JsonParser inside, where non standard (utf) encoding is used.

https://github.com/airbytehq/airbyte/blob/df85c670ab8b8004c0bdfdb348bb728cdd99c455/airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/manifest.yaml#L2174-L2180

I cannot reuse `JsonParser` here because GzipDecoder pass **bytes** to the inner_parser.

artem1205 (changes requested):

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: @artem1205 and I synced to understand this issue and it was a mix of manifest change and of bad wiring up in the model_to_component factory. I pushed a fix and we should be good on this one

artem1205 (approved):
LGTM!

ChristoGrab:
Adding a link to the issue to add this new functionality to the builder, just for x-referencing purposes:

https://github.com/airbytehq/airbyte-internal-issues/issues/11679

PR #328, opened 2025-02-10
feat: check for request_option mapping conflicts in individual components

ChristoGrab:
## What

Although we check for duplicate keys when combining all request option mappings in a retriever, individual components that define more than one request option lack validations to confirm that there are no key mapping errors. Therefore, we currently override values when a conflict occurs, ie:

```
## Inputs ##
paginator.page_size_option: ["variables", "limit"] = 5
paginator.page_token_option: ["variables", "limit"] = 1

## Outputs ##
Expected: "ValueError: Duplicate keys found..."
Actual: {"variables": { "limit": 1 } }
```

This PR adds a new validation specifically for components that allow multiple request options, to ensure that mappings are combined and checked for each component individually.

## How

- Implements a new helper function `_validate_component_request_option_paths`. This function takes an arbitrary number of request option keys and reuses the existing `combine_mappings` implementation to check for duplicate keys. A dict with dummy values is constructed from the component's keys (since the actual values may not be available at time of init). Resolves [#11580](https://github.com/airbytehq/airbyte-internal-issues/issues/11580)
- Updates the duplicate key error message for clarity. Resolves [#11581](https://github.com/airbytehq/airbyte-internal-issues/issues/11581)

## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration validation for time-based queries and pagination, ensuring request options are correctly set before execution and improving overall data retrieval accuracy.
  
- **Bug Fixes**
  - Upgraded error messaging to provide clearer feedback when configuration issues arise, making it easier for users to identify and resolve duplicate or conflicting settings.

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
@@ -21,7 +21,7 @@ from airbyte_cdk.sources.declarative.requesters.request_option import (
 )
 from airbyte_cdk.sources.message import MessageRepository
 from airbyte_cdk.sources.types import Config, Record, StreamSlice, StreamState
-from airbyte_cdk.utils.mapping_helpers import _validate_multiple_request_options
+from airbyte_cdk.utils.mapping_helpers import _validate_component_request_option_paths
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


maxi297 (approved):
nit: it feels like we should have a test for datetime_based_cursor and default_paginator. I don't mind not doing the cursor one because we are slowly moving away from this implementation. I'll let you decide if it's worth for default_paginator

The rest LGTM! Thanks @ChristoGrab

PR #331, opened 2025-02-12
feat(concurrent perpartition cursor): Refactor ConcurrentPerPartitionCursor

tolik0:
This PR fixes an issue with ConcurrentPerPartitionCursor, where it could fail if the number of generated partitions at the start of the sync exceeded the limit.

**Key Changes**

- Removed implicit dependency on ThreadPool limits
- Moved global cursor switching logic to stream_slices, enabling switching while generating slices
- Refactored `observe` and `close_partitions` methods to update only the global cursor value when global cursor is enabled
- Added throttling for state emission – now limited to once per minute to fix errors with the platform


airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -245,7 +242,7 @@ def _ensure_partition_limit(self) -> None:
                             partition_key
                         )  # Remove the oldest partition
                         logger.warning(
-                            f"The maximum number of partitions has been reached. Dropping the oldest finished partition: {oldest_partition}. Over limit: {self._over_limit}."
+                            f"The maximum number of partitions has been reached. Dropping the oldest finished partition: {oldest_partition}. Over limit: {self._number_of_partitions}."
```
maxi297, line 245: Should this log be updated? It seems like we print the number of partitions and not the number over the limit. The same comment applies to the log below

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -355,15 +352,32 @@ def _set_global_state(self, stream_state: Mapping[str, Any]) -> None:
 
     def observe(self, record: Record) -> None:
         if not self._use_global_cursor and self.limit_reached():
+            logger.info(
```
maxi297, line 355: It's a very nice addition! ❤️

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -355,15 +352,32 @@ def _set_global_state(self, stream_state: Mapping[str, Any]) -> None:
...
+        record_cursor = self._connector_state_converter.output_format(
+            self._connector_state_converter.parse_value(self._cursor_field.extract_value(record))
+        )
+        self._update_global_cursor(record_cursor)
+        if not self._use_global_cursor:
+            self._cursor_per_partition[
+                self._to_partition_key(record.associated_slice.partition)
+            ].observe(record)
+
+    def _update_global_cursor(self, value: Mapping[str, Any]) -> None:
```
maxi297, line 375: Should the type of `value` here be `CursorValueType`? Else, it seems like we set `self._new_global_cursor` to `{<cursor field>: {<cursor field>: <cursor value>}}`

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -58,7 +58,8 @@ class ConcurrentPerPartitionCursor(Cursor):
     CurrentPerPartitionCursor expects the state of the ConcurrentCursor to follow the format {cursor_field: cursor_value}.
     """
 
-    DEFAULT_MAX_PARTITIONS_NUMBER = 10000
+    DEFAULT_MAX_PARTITIONS_NUMBER = 25_000
+    SWITCH_TO_GLOBAL_LIMIT = 1000
```
maxi297, line 62: Will this be changes again once we implement the throttling?

maxi297 (commented):
The global cursor logic change looks good to me even though I have a question about the throttling. One thing that is surprising to me is that tests don't change as I would assume that the previous solution would set the global cursor on the slice boundaries while this change would set it on the record cursor value. Am I missing something?

Once we are clear on the tests, I'll approve assuming that limits will be re-worked later so that if you want to release the throttling changes in another PR, you can release these here.

tolik0:
@maxi297 The previous implementation set the state as most recent value if present and slices start when it is not present. So the test shouldn't change.
https://github.com/airbytehq/airbyte-python-cdk/blob/e27cb81196fec42e2a17f6fc728c7f6f066be194/airbyte_cdk/sources/streams/concurrent/state_converters/abstract_stream_state_converter.py#L53-L64

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -99,9 +101,11 @@ def __init__(
         self._new_global_cursor: Optional[StreamState] = None
         self._lookback_window: int = 0
         self._parent_state: Optional[StreamState] = None
-        self._over_limit: int = 0
+        self._number_of_partitions: int = 0
```
maxi297, line 104: Should we clarify the naming of this variable? What confuses me right now is why it is not incremented during `_set_initial_state`. It feels like those partitions will be in-memory which therefore I would assume should count towards the _number_of_partitions count

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -141,21 +145,16 @@ def close_partition(self, partition: Partition) -> None:
             raise ValueError("stream_slice cannot be None")
 
         partition_key = self._to_partition_key(stream_slice.partition)
```
maxi297, line 147: nit and outside of scope: I like the fact that we compute this value only once so that performance should be better. Should we do the same on `_generate_slices_from_partition`?

maxi297 (approved):
LGTM! 

Given the complexity of the ConcurrentPerPartitionCursor solution in general, can we add more details on the PR description to make sure someone that has less knowledge can understand more easily? For example, the PR description mentions that it is `enhancing scalability`. In which case the current solution is not stable? What does increasing the number of partition does for stability? My understanding is that it avoids cases where fetching for a partition might fail [here](https://github.com/airbytehq/airbyte-python-cdk/pull/331/files#diff-40bb6219c9d2c70a591e9e3b2209087e1b01eb44bfe36c15ff22b43506ac69faR151) assuming a partition would take a very long time to process but it would be nice to have this explicit

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
maxi297: I see this change as part of https://github.com/airbytehq/airbyte-python-cdk/pull/343/files#diff-40bb6219c9d2c70a591e9e3b2209087e1b01eb44bfe36c15ff22b43506ac69faR239 so I'm going to resolve this here

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: You're right - we need to count partitions in `_set_initial_state` as well. Fixed

Issue #283, opened 2025-01-28
dead code: remove print buffer


natikgadzhi:
## Context

Assuming we've disabled Print Buffer in the new CDK version because of concurrency problems, we should either fix it or rip the code out. 
/cc @artem1205

PR #336, opened 2025-02-13
fix: fix PrintBuffer logic

artem1205:
## What

Resolve https://github.com/airbytehq/airbyte-internal-issues/issues/10136
Resolve https://github.com/airbytehq/airbyte-python-cdk/issues/283

## How

- remove flushing
- pinned and tested on [test WS](https://cloud.airbyte.com/workspaces/8ba9a84d-ef49-41fb-9aaa-9f75fb0f351e/connections/bf6ede5c-e346-4489-a965-63a381eb3c30/timeline)

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced buffered logging mechanism that improves how output is managed and displayed.
  - Added an explicit flush control to better synchronize log output during high-load scenarios.

- **Bug Fixes**
  - Resolved issues related to inconsistent output during concurrent operations, ensuring a smoother logging experience.


artem1205:
TODO: run regression test on source stripe to check for missing records

natikgadzhi:
!!!!1111 thank you Artem! <3

artem1205:
UPD:

pinned and tested on [stripe test WS](https://cloud.airbyte.com/workspaces/8ba9a84d-ef49-41fb-9aaa-9f75fb0f351e/connections/bf6ede5c-e346-4489-a965-63a381eb3c30/timeline) — no errors

airbyte_cdk/utils/print_buffer.py
```diff
@@ -73,3 +73,7 @@ def __exit__(
     ) -> None:
         self.flush()
         sys.stdout, sys.stderr = self.old_stdout, self.old_stderr
+
+    def flush_logger(self) -> None:
```
maxi297, line 77: Where is this called?

airbyte_cdk/logger.py
```diff
@@ -27,7 +30,7 @@
     "handlers": {
         "console": {
             "class": "logging.StreamHandler",
-            "stream": "ext://sys.stdout",
```
maxi297, line 30: Should we be worried that some logs would be emitted before (maybe during `run.py`) or after (when there is a breaking exception and we catch the uncaught exception) the `with PRINT_BUFFER` statement?

maxi297 (commented):
Just a couple of questions as comments. I would also like a more exhaustive description to understand what caused the issue before and how this change fixes the issue.

airbyte_cdk/utils/print_buffer.py
artem1205: removed, used during test

airbyte_cdk/logger.py
artem1205: > I would also like a more exhaustive description to understand what caused the issue before and how this change fixes the issue.

@maxi297 , my guess is that issue was caused due to the logger.




> Should we be worried that some logs would be emitted before (maybe during run.py) or after (when there is a breaking exception and we catch the uncaught exception) the with PRINT_BUFFER statement?

not really, since we share exactly the same `object` (stream) in both logger.write and print statements.

maxi297 (commented):
So the problem was that `PrintBuffer` wasn't used as a `handlers. stream` in the logging config, right? As I don't have more information on the original issue, are we able to reproduce the issue that happened before, apply this change and see that we don't have the issue anymore?

artem1205:
> are we able to reproduce the issue that happened before, apply this change and see that we don't have the issue anymore?

I'll test it

artem1205:
@maxi297 ,

UPD:

after deeeeep Dive into logs from [Brian's init sync](https://cloud.airbyte.com/workspaces/3adc4c3d-711d-4872-864c-670c40ef54b5/connections/21919235-b322-4286-8fbc-b3b0274569c4/timeline) I got the idea why it could happen and why we do not see the difference between old style implementation of PrintBuffer vs fixed one.

there are  many lines of Logger: `Failed to transform value ` that were written to stdout bypassing the PrintBuffer logic.
From talk with Gemini:
>[!NOTE]
> The RLock (reentrant lock) is primarily designed to prevent data corruption when multiple threads access a shared resource (like your StringIO buffer).  It ensures that only one thread can modify the buffer at a time, preventing garbled output.  However, it does not provide a guarantee of delivery of every single write in the face of abrupt process termination.

So my guess is that PrintBuffer tried to write to `stdout` at the moment it was working on logger warning message and some records were missing.

Rollout plan:

1. release dev version of CDK with this fix
2. create a PR for `source-stripe` with dev CDK version
3. start progressive rollout and check for errors in channel (**proj-detect-dropped-records-platform-counts**)

maxi297 (approved):
Thanks for the explanation. Given the tests for source-stripe is successful, I'm fine with releasing this change. Poking @lazebnyi as he is working on source-stripe right now and this might conflict with his release

lazebnyi (approved):

natikgadzhi:
@artem1205 I would LOVE to see this in prod. Again, thank you for pushing this over the finish line.

natikgadzhi:
@artem1205 how's testing going? Should we merge this in?

artem1205:
@natikgadzhi ,

testing stuck on progressive rollout:
image


not sure what to do next, @clnoll could you assist with it?

clnoll:
@artem1205 looks like you set the initial rollout percent to 0, which put the rollout in an error state. If you cancel and restart it (but with a non-zero initial rollout percent), it should work.

artem1205:
cancelled, then created new one and
got an error when pressed `start_automated` 
image

image

natikgadzhi:
I am very curious to see if we get a perf bump with this in
hardcoded-records. Update it to check?

PR #343, opened 2025-02-17
feat(concurrent perpartition cursor): Add parent state updates

tolik0:
## Summary by CodeRabbit

- **New Features**
  - Upgraded partition management improves the handling of incremental data processing with increased partition limits and enhanced state synchronization.
  - New test scenarios validate cursor behavior in various edge cases, ensuring accurate state management during partition processing.

- **Bug Fixes**
  - Improved state management ensures the cursor correctly handles updates and emissions based on partition status.

- **Refactor**
  - Enhanced internal logic refines the switching mechanism between different processing modes, ensuring a smoother and more resilient operational flow.


maxi297 (commented):
So let me rephrase the logic of this implementation in words so that you can confirm my understanding.

## Order of events
While consuming the parent stream, we keep track of the state after each parent record as part of an ordered list of states described by `self._partition_parent_state_map`. When a partition on the child stream is closed, we can check a couple of things:
* Have we generated all the slices for this partition (for example, let's assume we have parent_record_id, we might have slices `{"partition": <parent_record_id>, "cursor": <january>}` and `{"partition": <parent_record_id>, "cursor": <february>}`)? If not, we can't progress the state because it is impossible for <parent_record_id> to be fully processed
* Have we processed all the slices associated with the parent record id? This is done using `self._semaphore_per_partition[p_key]._value != 0`

Given the two conditions above, it means that we can progress the state. Here is a timeline of what events can look like:
* T0: Generate `{"partition": <parent_record_id>, "cursor": <january>}`, january is NOT last
* T1: Closing `{"partition": <parent_record_id>, "cursor": <january>}`, the parent state is not progressed because we still have slices to generate
* T2: Generate `{"partition": <parent_record_id>, "cursor": <february>}`, february is last
* T3: Closing `{"partition": <parent_record_id>, "cursor": <february>}`, we can progress the parent state

This works with unordered execution of slices with the same slice.partition but different slice.cursor:
* T0: Generate `{"partition": <parent_record_id>, "cursor": <january>}`, january is NOT last
* T1: Generate `{"partition": <parent_record_id>, "cursor": <february>}`, february is last
* T2: Closing `{"partition": <parent_record_id>, "cursor": <february>}`, the parent state is not progressed because we still have slices to process
* T3: Closing `{"partition": <parent_record_id>, "cursor": <january>}`, we can progress the parent state

This works with unordered execution of slices with different child slice.partition:
* T0: Generate `{"partition": <parent_record_id_1>, "cursor": <whatever>}`, whatever is last for parent_record_id_1
* T1: Generate `{"partition": <parent_record_id_2>, "cursor": <whatever>}`, whatever is last for parent_record_id_2
* T2: Closing `{"partition": <parent_record_id_2>, "cursor": <whatever>}`, the parent state is not progressed because parent_record_id_1 is first in the list and hasn't been cleared out
* T3: Closing `{"partition": <parent_record_id_1>, "cursor": <whatever>}`, we can progress the parent state to the parent state when `parent_record_id_2` was emitted

## Assumptions/concerns
* [I'm fine with this one] The state is updated as part of the partition router. We wanted to move away from DeclarativeStream traction their own state because of issues it created (calling twice `stream_slices` or `read_records` caused issues in the past). But we can have `SubstreamPartitionRouter` track the state using a cursor in the future and that would be just fine as well.
* [I think this is already the case but...] Given we don't support checkpointing state on records, the cursors should only progress the state when closing partitions. 
* The parent state is deep copied in order to avoid it being updated for all entries in `self._partition_parent_state_map` at the same time.
* [I think we need to evaluate the likeliness of issues related to this] Memory usage is not an issue. Do we need a backup plan if there are a lot of partitions?
    * An idea of optimization memory wise would be to remove the entry in _partition_parent_state_map if the state for the next entry is the same. For example, if records are emitted in the order ["parent_record_id_1", "parent_record_id_2", "parent_record_id_3"] and all records are from the same parent slice, it is very likely that the state will be the same because we only progress the state when closing the slice. This would mean that we would have the following sequence of events:
        * T0: Generate slice for `parent_record_id_1` with parent state <parent state 1>, self._partition_parent_state_map = {"parent_record_id_1": <parent state 1>}
        * T1: Generate slice for `parent_record_id_2` with parent state <parent state 1>, self._partition_parent_state_map = {"parent_record_id_2": <parent state 1>} instead of {"parent_record_id_1": <parent state 1>, "parent_record_id_2": <parent state 1>}
        * T2: Generate slice for `parent_record_id_3` with parent state <parent state 2>, self._partition_parent_state_map = {"parent_record_id_2": <parent state 1>, "parent_record_id_3": <parent state 2>} instead of {"parent_record_id_1": <parent state 1>, "parent_record_id_2": <parent state 1>, "parent_record_id_3": <parent state 2>}
    This would not necessarily be enough to prevent any memory creep so we might need to find other guardrails

## Benefits/drawbacks
Having the solution as part of ConcurrentPerPartitionCursor is good because:
* We already track most of the information needed (apart from the parent state) in the ConcurrentPerPartitionCursor
* The logic does not bleed (like in some proposal like [this one](https://airbytehq-team.slack.com/archives/D079WNVV2FN/p1739552676801859) where we would pass the information through the stream_slice)

Therefore, I think it is a smart solution and it feels we should go forward with it. What would be the next steps? Unit tests? Testing on the connection we struggled with for source-amazon-ads?

tolik0:
@maxi297 Yes, the next steps will be to implement the optimization you suggested, followed by adding unit tests and testing with Amazon Seller Partner users who previously experienced issues.

maxi297:
Regarding my optimization, I don't think it is possible as we need to know `parent_record_id_1`,  `parent_record_id_2` and `parent_record_id_3` have all been processed in order to move the cursor to `<parent state 2>`. Hence, we need to keep track of `parent_record_id_1` and `parent_record_id_2` somewhere (without necessarily having to have the parent state multiple times though). This highlights the fact that order of all parent records is important and we will need to keep track of that. In order words, optimizing memory is less straightforward than I envisioned.

maxi297:
Adding to the previous comment, we need to keep all semaphores in memory anyway because if one does not have its value == 0, it means we can't progress the global state. We assume all those semaphores to be very low in terms of memory usage given that they are only a wrapper over and integer. Hence, we can reduce memory usage by instead of `self._semaphore_per_partition[p_key]._value == 0`,  we validate that all semaphores before `p_key` have a value of 0

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -155,7 +159,45 @@ def close_partition(self, partition: Partition) -> None:
...
+        """
+        last_closed_state = None
+
+        while self._partition_parent_state_map:
+            # Look at the earliest partition key in creation order
+            earliest_key = next(iter(self._partition_parent_state_map))
+
+            # Verify ALL partitions from the left up to earliest_key are finished
+            all_left_finished = True
+            for p_key, sem in self._semaphore_per_partition.items():
```
maxi297, line 181: nit: do we want to keep an indice of the last semaphore we cleaned the parent state for? I seems like if we don't have that, we will iterate over X items for nothing, X growing as we progress in the sync. I think I don't mind too much but since this is under a lock and therefore will block any other threads from generating slices, I fear it might be slightly important to optimize that

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -3017,3 +3027,260 @@ def test_state_throttling(mocker):
...
+        "lookback_window": 0,
+        "states": [],
+    }
+    assert len(cursor._cursor_per_partition) == 0
+    assert len(cursor._semaphore_per_partition) == 0
+    assert len(cursor._partition_parent_state_map) == 0
+    assert mock_cursor.stream_slices.call_count == 0  # No calls since no partitions
+
+
+def test_given_new_partition_mid_sync_when_close_partition_then_update_state():
```
maxi297, line 3080: What do we mean by `new_partition_mid_sync`? This test seems to generate all the partitions (because of `slices = list(cursor.stream_slices())` which will generate everything) before closing them all (the for loop). Hence, I'm not sure where is the new partition here

maxi297 (approved):
Should we have a test where the parent state is partially progressed because the last partition that was generated has not yet been completed but the first one has?

Apart from those small comments, I think we're good!

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: I added deleting semaphores when their value reaches 0, since we only need them to check if no errors occurred.

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
tolik0: I added two new tests to cover cases with unfinished partitions.

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
maxi297: Just to precise, the code also checks that the partition is done generating on top of having the semaphore value to 0. Hence, I'm good with this change. It will also help for memory management which is great. Thanks!

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -196,18 +251,25 @@ def _emit_state_message(self, throttle: bool = True) -> None:
         self._message_repository.emit_message(state_message)
 
     def stream_slices(self) -> Iterable[StreamSlice]:
+        print("stream_slices")
```
maxi297, line 254: nit: should we remove this print?

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -3017,3 +3027,428 @@ def test_state_throttling(mocker):
...
+    ]
+
+    slices = list(cursor.stream_slices())
+    # Close all partitions except from the first one
+    for slice in slices[1:]:
+        cursor.close_partition(
+            DeclarativePartition("test_stream", {}, MagicMock(), MagicMock(), slice)
+        )
+    cursor.ensure_at_least_one_state_emitted()
+    print(cursor.state)
```
maxi297, line 3147: nit: should we remove this print?

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -3017,3 +3027,428 @@ def test_state_throttling(mocker):
...
+    ]
+
+    slices = list(cursor.stream_slices())
+    # Close all partitions except from the first one
+    for slice in slices[:-1]:
+        cursor.close_partition(
+            DeclarativePartition("test_stream", {}, MagicMock(), MagicMock(), slice)
+        )
+    cursor.ensure_at_least_one_state_emitted()
+    print(cursor.state)
```
maxi297, line 3232: nit: should we remove this print?

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
maxi297: Just one confirmation I'd like from you: Conceptually, we can still access `self._semaphore_per_partition[p_key]` in most places because when we `del self._semaphore_per_partition[p_key]` during the clean up process, we don't expect anyone to access this semaphore anymore, right? I only saw one place where we were doing this validation [here](https://github.com/airbytehq/airbyte-python-cdk/pull/343/files#diff-40bb6219c9d2c70a591e9e3b2209087e1b01eb44bfe36c15ff22b43506ac69faR331)

maxi297 (approved):
Given the CI passes, I'm good with this change. The comments are just nits/validation. Thanks @tolik0

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
tolik0: In the current implementation, `self._semaphore_per_partition[p_key]` is accessed directly by key in a few places, such as `_check_and_update_parent_state` (for cleanup) and `_ensure_partition_limit` (to check if the partition is finished). I’ve added a check to handle cases where the semaphore might already be deleted: if it’s missing from `_semaphore_per_partition`, it implies its value was 0 and the partition is fully finished (i.e., all slices processed with no errors). In this state, neither `observe` nor `close_partition` would be called for that partition, as all its slices have already been closed. This cleanup happens under the lock, ensuring thread safety, so other threads won’t access a deleted semaphore unexpectedly.

PR #346, opened 2025-02-18
feat: add IncrementingCountCursor

artem1205:


artem1205:
UPD:

found an interesting bug:
concurrentcursor is used only during the `read`, if we'll run `check` command on stream with the `IncrementingCountCursor` we'll face a bug.

we need to implement IncrementingCountCursor to make it pass

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -397,11 +411,18 @@ def _is_datetime_incremental_without_partition_routing(
...
-            and incremental_sync_component_definition.get("type", "")
-            == DatetimeBasedCursorModel.__name__
+            and (
+                incremental_sync_component_definition.get("type", "")
+                in (DatetimeBasedCursorModel.__name__, IncrementingCountCursorModel.__name__)
+            )
             and hasattr(declarative_stream.retriever, "stream_slicer")
             and (
                 isinstance(declarative_stream.retriever.stream_slicer, DatetimeBasedCursor)
+                # IncrementingCountCursorModel is hardcoded to be of type DatetimeBasedCursor
```
maxi297, line 421: Can we update the method name to reflect that? I'm not sure what would be the best name but my understanding is that it's basically `_is_concurrent_cursor_incremental_without_partition_routing`

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -777,6 +777,44 @@ definitions:
       type:
         type: string
         enum: [LegacyToPerPartitionStateMigration]
+  IncrementingCountCursor:
+    title: Incrementing Count Cursor
+    description: Cursor that allows for incremental sync according to a continuously increasing number
```
maxi297, line 782: nit: a `number` in json schema language can be a float. Should we be precise here and say it's integers only?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1622,6 +1693,31 @@ def create_declarative_stream(
...
+                )
+                if cursor_model.start_value_option  # type: ignore # mypy still thinks cursor_model of type DatetimeBasedCursor
+                else None
+            )
+
+            # The concurrent engine defaults the start/end fields on the slice to "start" and "end", but
+            # the default DatetimeBasedRequestOptionsProvider() sets them to start_time/end_time
+            partition_field_start = "start"
+
+            request_options_provider = DatetimeBasedRequestOptionsProvider(
```
maxi297, line 1715: The naming here is weird. Should we generify the `DatetimeBasedRequestOptionsProvider` to be a `StartEndRequestOptionsProvider` or something like that? Would this allow us to have only one method for building both `DatetimeBasedCursor` and `IncrementingCountCursor`? The comments above refer to `DatetimeBasedCursor` which makes me think there is duplication

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2111,6 +2207,22 @@ def create_gzip_decoder(
             stream_response=False if self._emit_connector_builder_messages else True,
         )
 
+    @staticmethod
```
maxi297, line 2210: Adding a comment to document what we said about that: We want to remove the usage on the declarative cursor package but there is some pre-work to be done for that like https://github.com/airbytehq/airbyte-internal-issues/issues/11771. We will design this part more thoroughly soon and this code should be remove during this work

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
```diff
@@ -0,0 +1,92 @@
...
+        if self.is_state_message_compatible(stream_state):
+            return sync_start, stream_state
+
+        # Create a slice to represent the records synced during prior syncs.
+        # The start and end are the same to avoid confusion as to whether the records for this slice
+        # were actually synced
+        slices = [
+            {
+                self.START_KEY: start if start is not None else sync_start,
+                self.END_KEY: sync_start,  # this may not be relevant anymore
```
maxi297, line 53: Why wouldn't it be relevant anymore? Can we just remove it if it's the case?

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
```diff
@@ -0,0 +1,92 @@
...
+                self.MOST_RECENT_RECORD_KEY: sync_start,
```
maxi297, line 54: I would have assumed that the value there would be the most_recent_cursor_value we've seen, not the start of the sync

maxi297 (commented):
Should we have tests that ensures that a manifest model creates the expected behavior?

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
artem1205: i guess we reused the logic from `DateTimeStreamStateConverter` https://github.com/airbytehq/airbyte-python-cdk/pull/52

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
artem1205: this method is used only in `ConcurrentCursor._get_concurrent_state`  https://github.com/airbytehq/airbyte-python-cdk/blob/44592437b3eb42c5103c7708de52312f893a0aa7/airbyte_cdk/sources/streams/concurrent/cursor.py#L191-L201

and, since we don't have any "old-style" states (`is_state_message_compatible` -> `True`) we may never use `convert_from_sequential_state` for this cursor.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1189,6 +1196,70 @@ def create_concurrent_cursor_from_datetime_based_cursor(
...
+        )
+
+        interpolated_cursor_field = InterpolatedString.create(
+            incrementing_count_cursor_model.cursor_field,
+            parameters=incrementing_count_cursor_model.parameters or {},
+        )
+        cursor_field = CursorField(interpolated_cursor_field.eval(config=config))
+
+        connector_state_converter = IncrementingCountStreamStateConverter(
+            is_sequential_state=True,  # ConcurrentPerPartitionCursor only works with sequential state
```
maxi297, line 1247: This is not concurrent per partition though. I think the real reason to keep it as sequential state is that since the slice_range is never defined for IncrementalCountCursor, there is not value in having partitioned state

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
maxi297: I think we might use the `convert_from_sequential_state` if we can find a way to have slice_range. For now, I'm fine with this.

airbyte_cdk/sources/streams/concurrent/state_converters/incrementing_count_stream_state_converter.py
maxi297: I didn't understand the context properly and this is during conversion from sequential to sliced state. Ignore my comment

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -777,6 +777,44 @@ definitions:
       type:
         type: string
         enum: [LegacyToPerPartitionStateMigration]
+  IncrementingCountCursor:
```
maxi297, line 780: @lmossman would you have a look at the YAML definition of this component to see if anything seems odd to you?

maxi297 (commented):
Can we have tests that ensures that a manifest model creates the expected behavior?

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -777,6 +777,44 @@ definitions:
...
+          - type: string
+          - type: integer
+        interpolation_context:
+          - config
+        examples:
+          - 0
+          - "{{ config['start_value'] }}"
+      start_value_option:
+        title: Inject Start Value Into Outgoing HTTP Request
+        description: Optionally configures how the start datetime will be sent in requests to the source API.
```
lmossman, line 813:
```suggestion
        description: Optionally configures how the start value will be sent in requests to the source API.
```

lmossman (commented):
Had one small suggestion on adjusting the description of a field, otherwise the YAML schema changes make sense to me!

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
artem1205: Done

maxi297 (approved):
Thanks for the updates @artem1205. I'm good with this change.

Regarding the slice_range, I think you are right on how to process with that but this is definitely out of scope. Thanks for digging into this!

PR #358, opened 2025-02-20
fix: (CDK) (CsvParser) - Fix the `\\` escaping when passing the `delimiter` from Builder's UI

bazarnov:
## What
Sometimes there is a case when the `csv` file is encoded with the `\t` or other delimiters supported (basically any character) and should be passed with the `escape` character alongside. This breaks the `CsvParser` implementation when the input goes from the Builder's UI.

More context here: https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1740003282758399

## How
- added pre-parsing for the `delimiter` to decode the `escaping_character` and normalize the input before decoding records.

## User Impact
No impact is expected, this is not a Breaking change.

## Summary by CodeRabbit

- **New Features**
  - Enhanced CSV file parsing to correctly interpret escaped delimiter characters. This improvement provides a more robust approach to processing CSV files, resulting in better data ingestion and compatibility with varied CSV formats.
- **Tests**
  - Updated test cases to validate the new delimiter handling and ensure reliable, consistent CSV data processing. Additional validations were added to safeguard integration performance.

### DIFF
--- a/airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
+++ b/airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
@@ -107,6 +107,16 @@ class CsvParser(Parser):
     encoding: Optional[str] = "utf-8"
     delimiter: Optional[str] = ","
 
+    def _get_delimiter(self) -> Optional[str]:
+        """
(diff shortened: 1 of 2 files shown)


ChristoGrab:
Thanks for the fix @Baz!

PR #364, opened 2025-02-21
fix: update csv parser for builder compatibility

ChristoGrab:
## What

Addresses a couple issues related to enabling the `CsvParser` in the Builder UI.

1. When trying to parse a CSV file using the declarative `CsvParser`, the following error gets thrown:

```
ValueError: I/O operation on closed file.
```

This seems to be because when we `yield from reader`, the file handle is somehow closed before the data has been read. Iterating and yielding each row seems to resolve the issue.

2. Specifically when running a test read in the builder, even when getting a successful response with the CSV data included, 0 records are being read:

Screenshot 2025-02-20 at 3 33 38 PM

The issue appears to be because the contents of the record generator are consumed when the records are emitted during read. So no data is subsequently passed to the Builder to be displayed in the testing panel. By flagging the component to specifically not emit messages when in the Builder environment, we can get accurate test results again.

## Summary by CodeRabbit

- **Refactor**
	- Updated the CSV processing logic to use explicit iteration, enhancing clarity while maintaining the same end results.
	- Modified the component creation process to support dynamic behavior based on instance state.

### DIFF
--- a/airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
+++ b/airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
@@ -126,7 +126,8 @@ class CsvParser(Parser):
         """
         text_data = TextIOWrapper(data, encoding=self.encoding)  # type: ignore
         reader = csv.DictReader(text_data, delimiter=self._get_delimiter() or ",")
-        yield from reader
+        for row in reader:
+            yield row
 
 
 @dataclass
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2103,10 +2103,12 @@ class ModelToComponentFactory:
             parser=ModelToComponentFactory._get_parser(model, config), stream_response=True
         )
 
-    @staticmethod
-    def create_gzip_decoder(model: GzipDecoderModel, config: Config, **kwargs: Any) -> Decoder:
+    def create_gzip_decoder(
+        self, model: GzipDecoderModel, config: Config, **kwargs: Any
+    ) -> Decoder:
         return CompositeRawDecoder(
-            parser=ModelToComponentFactory._get_parser(model, config), stream_response=True
+            parser=ModelToComponentFactory._get_parser(model, config),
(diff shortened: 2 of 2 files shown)


airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -126,7 +126,8 @@ def parse(
         """
         text_data = TextIOWrapper(data, encoding=self.encoding)  # type: ignore
         reader = csv.DictReader(text_data, delimiter=self._get_delimiter() or ",")
-        yield from reader
+        for row in reader:
```
bazarnov, line 129: This is the same as `yield from reader` actually, why did we changed it?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2091,10 +2091,10 @@ def create_dynamic_schema_loader(
     def create_json_decoder(model: JsonDecoderModel, config: Config, **kwargs: Any) -> Decoder:
         return JsonDecoder(parameters={})
 
-    @staticmethod
-    def create_csv_decoder(model: CsvDecoderModel, config: Config, **kwargs: Any) -> Decoder:
+    def create_csv_decoder(self, model: CsvDecoderModel, config: Config, **kwargs: Any) -> Decoder:
```
bazarnov, line 2094: we should do the same for the `create_gzip_decoder` method.

ChristoGrab: Added 👍

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
ChristoGrab: This is related to the error both Wass and I were seeing previously:

```
ValueError: I/O operation on closed file.
```

For some reason the file handle was closing before the operation completes and causing the connector to crash. For whatever reason this seems to fix the problem 🤷

bazarnov: I'm fine with this. Thanks.

PR #366, opened 2025-02-25
fix: prevent raw decoder from closing before sync completion

ChristoGrab:
## What

We've run into an issue with the `CsvDecoder` closing files prematurely during syncs, leading to crashes when syncing and/or setting up sources:

```
ValueError: I/O operation on closed file
```

The problem seems to be that the connection is automatically closed once the last chunk is being processed, preventing the process from gracefully completing. By explicitly closing the connection AFTER we have yielded the data, we can ensure that the connection remains until all data has been parsed and read.

Note that not only csv fails with the same error but pandas as well which seems to indicate an issue regarding how urllib3 handles closing of IO rather than an issue with the Python csv standard lib.

>[!NOTE] 
>From [docs](https://urllib3.readthedocs.io/en/stable/user-guide.html#using-io-wrappers-with-response-content)
> Sometimes you want to use [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) or similar objects like a CSV reader directly with [HTTPResponse](https://urllib3.readthedocs.io/en/stable/reference/urllib3.response.html#urllib3.response.HTTPResponse) data. Making these two interfaces play nice together requires using the [auto_close](https://urllib3.readthedocs.io/en/stable/reference/urllib3.response.html#urllib3.response.HTTPResponse.auto_close) attribute by setting it to False. By default HTTP responses are closed after reading all bytes, this disables that behavior:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved streamed response handling to ensure that the data stream remains available during processing and is properly closed afterward, enhancing overall stability and resource management.

- **New Features**
  - Introduced a new test server class for testing CSV parsing capabilities of the `CompositeRawDecoder`.
  - Added a new test function to validate the CSV parsing functionality in a controlled environment.
  - Added a utility function to find available ports for testing purposes.

### DIFF
--- a/unit_tests/sources/declarative/decoders/test_composite_decoder.py
+++ b/unit_tests/sources/declarative/decoders/test_composite_decoder.py
@@ -226,9 +226,8 @@ def test_composite_raw_decoder_csv_parser_without_mocked_response():
     httpd = HTTPServer(("localhost", 8080), TestServer)
     thread = Thread(target=httpd.serve_forever, args=())
     thread.start()
-    thread.start()
     try:
-        response = requests.get(f"http://localhost:{port}", stream=True)
+        response = requests.get(f"http://localhost:8080", stream=True)
         result = list(CompositeRawDecoder(parser=CsvParser()).decode(response))
 
         assert len(result) == 1
--- a/unit_tests/sources/declarative/decoders/test_composite_decoder.py
+++ b/unit_tests/sources/declarative/decoders/test_composite_decoder.py
@@ -4,6 +4,7 @@
 import csv
 import gzip
 import json
+import socket
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from io import BytesIO, StringIO
 from threading import Thread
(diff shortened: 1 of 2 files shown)


maxi297:
@maxi297 to add a test on that

unit_tests/sources/declarative/decoders/test_composite_decoder.py
```diff
@@ -202,6 +204,30 @@ def test_composite_raw_decoder_csv_parser_values(requests_mock, encoding: str, d
...
+        self.end_headers()
+        self.wfile.write(bytes("col1,col2\nval1,val2", "utf-8"))
+
+
+def test_composite_raw_decoder_csv_parser_without_mocked_response():
+    """
+    This test reproduce a `ValueError: I/O operation on closed file` error we had with CSV parsing. We could not catch this with other tests because the closing of the mocked response from requests_mock was not the same as the one in requests.
+    """
+    # start server
+    httpd = HTTPServer(("localhost", 8080), TestServer)
```
maxi297, line 219: We should probably have a better way to find a port. Is there a port finder in Python?

artem1205: I guess we can use `socket` package to do it

artem1205 (approved):
LGTM!
approving to unblock, we might need to introduce another buffer to the CSV parser in order to handle and check for \n at the end of the line.

unit_tests/sources/declarative/decoders/test_composite_decoder.py
maxi297: Let's go with this for now and if we see issues, we can migrate to socket directly

PR #370, opened 2025-02-27
fix: (CDK) (HttpRequester) - Make the `path` arg optional

bazarnov:
## What
Resolving:
- https://github.com/airbytehq/airbyte-internal-issues/issues/11210

## How
- TODO

## User Impact
This update introduces the `Requester` interface change, which adds the ability for `get_url_base()` to utilize the interpolation context passed, such as `stream_state,` `stream_slice` / `stream_interval`, and `next_page_token.`

Sources that reuse the interface may need to update the interface for the `get_url_base()` method if it has been overridden. Otherwise, we anticipate no breaking changes from this CDK update.

### DIFF
--- a/airbyte_cdk/sources/declarative/requesters/http_requester.py
+++ b/airbyte_cdk/sources/declarative/requesters/http_requester.py
@@ -25,7 +25,7 @@ from airbyte_cdk.sources.message import MessageRepository, NoopMessageRepository
 from airbyte_cdk.sources.streams.call_rate import APIBudget
 from airbyte_cdk.sources.streams.http import HttpClient
 from airbyte_cdk.sources.streams.http.error_handlers import ErrorHandler
-from airbyte_cdk.sources.types import Config, StreamSlice, StreamState
+from airbyte_cdk.sources.types import Config, EmptyString, StreamSlice, StreamState
 from airbyte_cdk.utils.mapping_helpers import combine_mappings
 
 
@@ -49,9 +49,10 @@ class HttpRequester(Requester):
 
     name: str
     url_base: Union[InterpolatedString, str]
-    path: Union[InterpolatedString, str]
     config: Config
     parameters: InitVar[Mapping[str, Any]]
+
+    path: Optional[Union[InterpolatedString, str]] = None
     authenticator: Optional[DeclarativeAuthenticator] = None
     http_method: Union[str, HttpMethod] = HttpMethod.GET
     request_options_provider: Optional[InterpolatedRequestOptionsProvider] = None
@@ -66,7 +67,9 @@ class HttpRequester(Requester):
 
     def __post_init__(self, parameters: Mapping[str, Any]) -> None:
         self._url_base = InterpolatedString.create(self.url_base, parameters=parameters)
-        self._path = InterpolatedString.create(self.path, parameters=parameters)
+        self._path = InterpolatedString.create(
+            self.path if self.path else EmptyString, parameters=parameters
+        )
         if self.request_options_provider is None:
             self._request_options_provider = InterpolatedRequestOptionsProvider(
                 config=self.config, parameters=parameters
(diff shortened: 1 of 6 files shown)


lmossman (commented):
YAML changes look good to me! Thanks Baz

airbyte_cdk/sources/declarative/requesters/http_requester.py
```diff
@@ -112,27 +115,50 @@ def exit_on_rate_limit(self, value: bool) -> None:
...
+            # update the context with extra fields, if passed.
             **(
                 stream_slice.extra_fields
                 if stream_slice is not None and hasattr(stream_slice, "extra_fields")
                 else {}
             ),
         }
-        path = str(self._path.eval(self.config, **kwargs))
+
+    def get_url_base(
```
maxi297, line 135: I'm not sure how this can work. It seems like we query `get_url_base` [here](https://github.com/airbytehq/airbyte-python-cdk/blob/406542d10bdc843096c9da727756094be50352f5/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L2519-L2523) in order to create the paginator. This is needed to remove the base_url from the next_page_token [here](https://github.com/airbytehq/airbyte-python-cdk/blob/406542d10bdc843096c9da727756094be50352f5/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py#L155-L157). This means that we either need to:
* not rely on params we don't have at the start of the main.py
* find another way to do that in the paginator

I would prefer to do the first option as I don't think we have a case to interpolate on anything than the config, right?

maxi297 (commented):
I shared a concern which I think is blocking for this PR

airbyte_cdk/sources/declarative/requesters/http_requester.py
bazarnov: From the PR description:
> This update introduces the Requester interface change, which adds the ability for get_url_base() to utilize the interpolation context passed, such as stream_state, stream_slice / stream_interval, and next_page_token.

At runtime, we now have the capability to interpolate from `stream_state,` `stream_slice,` `stream_interval,` and `next_page_token.` However, this interpolation is not necessary for component creation because none of these values are available before the component is created and begins executing.
The ability to utilize the values from `{{ config[x] }}` for the `url_base` remains intact, ensuring that there are no regressions during testing or component creation in other locations.  

> not rely on params we don't have at the start of the main.py
This is how is was before, and this is how we do after this particular change we discuss.

bazarnov: > It seems like we query get_url_base [here](https://github.com/airbytehq/airbyte-python-cdk/blob/406542d10bdc843096c9da727756094be50352f5/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L2519-L2523) in order to create the paginator. This is needed to remove the base_url from the next_page_token [here](https://github.com/airbytehq/airbyte-python-cdk/blob/406542d10bdc843096c9da727756094be50352f5/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py#L155-L157).

We didn't attempt to push the update to every single place, but to the necessary ones only, like the general one: `HttpRequester,` which operates with all the parts we need; these parts are referenced correctly at runtime, which seems correct to me. The specific parts like `paginator`, should remain intact and re-factored only when we need to.

maxi297: Can you do a test with a manifest that has interpolation of stream slice on url_base and a paginator where the next_page_token is a full URL? Assuming url_base being `https://api.sendgrid.com/{stream_slice.campaign_id}`, my guess is that the pagination will not work because the string replace [here](https://github.com/airbytehq/airbyte-python-cdk/blob/406542d10bdc843096c9da727756094be50352f5/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py#L155-L157) is `paginator.path({"next_page_token": "https://api.sendgrid.com/campaign_1/next_page_token"})` would return `'https://api.sendgrid.com/campaign_1/next_page_token'` which is not a path but a full URL. Hence, I think string interpolation on stream slice does not work in this case.

bazarnov: Covered with `airbyte-python-cdk/unit_tests/sources/declarative/requesters/paginators/test_default_paginator.py`:
- test_path_returns_none_when_no_token
- test_path_returns_none_when_option_not_request_path
- test_path_with_additional_interpolation_context

Specifically, the `test_path_with_additional_interpolation_context` is the test you've asked for.

maxi297 (approved):
I was able to confirm that things are fine with [this test](https://github.com/airbytehq/airbyte-python-cdk/compare/baz/cdk/http-requester-make-path-optional...maxi297/ensure-interpolation-on-path?expand=1). I'm good with this change. Thanks for the fix @bazarnov

PR #378, opened 2025-03-04
fix: (CDK) (AsyncRetriever) - Use the `Nested Decoders` to decode the streaming responses, instead of `ResponseToFileExtractor`

bazarnov:
## What
Resolving: 
- https://github.com/airbytehq/airbyte-internal-issues/issues/11212
- https://github.com/airbytehq/airbyte-internal-issues/issues/11869

## How

- **GzipDecoder Updates:**
   - Added the ability to fall back to reading records "as is" if the data is not actually gzip-compressed.
- Updated the `decoder` notation in `model_to_component_factory.py` to distinguish between the decoder used where the `ResponseToFileExtractor` support is terminated. This is the deprecated component that should be replaced by the "nested decoders" in the future. Please refer to the example below for more information.
- The `download_extractor` should be supplied as a part of the manifest, only when we have to deal with the field pointer to the data (basically when data is nested within the `response`), the example is [here](https://github.com/airbytehq/airbyte/blob/4667527f1d7b301d73e7b774fc9315c309e6d5ac/airbyte-integrations/connectors/source-amazon-seller-partner/source_amazon_seller_partner/manifest.yaml#L1944-L1946)

Side-Fix:
- Fixed the https://github.com/airbytehq/airbyte-internal-issues/issues/11869 as a small regression after [this CDK update](https://github.com/airbytehq/airbyte-python-cdk/pull/370)
   - reverted the test case to confirm the fix works and included more examples of what is expected from the `_join_url` method.

# User Impact

Before we supported the CSV-only:
```yaml
download_extractor:
  type: ResponseToFileExtractor
```

After:

For JSON-Lines:
```yaml
download_decoder:
    type: JsonlDecoder
```
For CSV
```yaml
download_decoder:
    type: CsvDecoder
```
For Gzipped JSON-Lines:
```yaml
download_decoder:
    type: GzipDecoder
    decoder:
      type: JsonlDecoder
```
For Gzipped CSV:
```yaml
download_decoder:
    type: GzipDecoder
    decoder:
      type: CsvDecoder
```

## Summary by CodeRabbit

- **New Features**
  - Introduced support for handling compressed responses in various decoder classes.
  - Added methods to check if responses are compressed, enhancing data processing capabilities.
  - New test cases for gzip handling in composite raw decoder, including scenarios for both compressed and non-compressed responses.
  - Added a new abstract base class for parsers to standardize parsing functionality.

- **Refactor**
  - Updated JSONL decoding to utilize instance state for consistent streaming behavior across data formats.
  - Improved documentation and readability in HTTP requester methods.
  - Streamlined method signatures and return types across various decoder classes.

- **Bug Fixes**
  - Corrected URL construction logic to retain trailing slashes in paths.


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1678,6 +1678,10 @@ definitions:
       type:
         type: string
         enum: [ResponseToFileExtractor]
+      file_type:
+        title: The file type in which the response data is storred. Supported types are [csv, jsonl].
+        type: string
+        default: csv
```
maxi297, line 1684: Should we use [the GzipDecoder](https://github.com/airbytehq/airbyte-python-cdk/blob/472199fd7489f7c972673da793b5a0a686d184cf/airbyte_cdk/sources/declarative/declarative_component_schema.yaml#L3210-L3224) instead? I think at the time we re-used the parsing that was in source-salesforce because we didn't have the parser for it and the file base CDK wasn't re-usable but now we have parsers to allow that so I'm wondering if we are to add a new parameter in the declarative language if it should be a "standard" one

maxi297 (commented):
I just want to make sure we don't introduce something in the declarative language that we would need to revert therefore creating a breaking change

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
bazarnov: Basically, we can have it working like this, with no other modification, and we don't need the `download_extractor.type == ResponseToFileExtractor` to be used here at all.

For JSON-Lines:
```yaml
download_decoder:
    type: JsonlDecoder
```
For CSV
```yaml
download_decoder:
    type: CsvDecoder
```
For Gzipped JSON-Lines:
```yaml
download_decoder:
    type: GzipDecoder
    decoder:
      type: JsonlDecoder
```
For Gzipped CSV:
```yaml
download_decoder:
    type: GzipDecoder
    decoder:
      type: CsvDecoder
```

My primary concern is that the server may transmit the asynchronously collected data in a different format.
- If the data is small, it is typically not compressed using `gzip`. Therefore, we require a mechanism to dynamically select between `GzipDecoder` and `CsvDecoder` (for CSV-expected responses) or other decoders based on the data format.

To address this, we should modify the `GzipParser` to instruct it to handle non-encoded data provided to the `gzip` decoder. In the event that a small chunk of data is not compressed by the server, the parser should fall back to parsing the data as is.

After implementing this modification, we can confidently rely on the `download_decoder` to correctly parse the data from the `{{ download_target }}`.

WDYT?

bazarnov: Here is the loom demo on how it can work without the `ResponseToFileExtractor`: https://www.loom.com/share/8b968bdaaaf8404caff91f458f870ffd

maxi297: Cool, that makes sense. What it means for me is:
* As long as we can stream, we can avoid `ResponseToFileExtractor`
* By avoiding `ResponseToFileExtractor`, we can use decoders

So let's stop supporting `ResponseToFileExtractor` until we see an actual need. We can migrate `ResponseToFileExtractor` to decoders once de CSV parsing supports null and other functionalities that are currently supported by `ResponseToFileExtractor`

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -35,15 +35,36 @@ def parse(
...
     inner_parser: Parser
 
+    def _reset_reader_pointer(self, data: BufferedIOBase) -> None:
+        """
+        Reset the reader pointer to the beginning of the data.
+
+        Note:
+            - This is necessary because the gzip decompression will consume the data stream.
+        """
+        data.seek(0)
```
maxi297, line 45: Can we confirm that this works with a real HTTP response? I don't think this operation is supported based on my local tests:
```
PyDev console: starting.
Python 3.10.11 (main, Jun 28 2023, 11:54:29) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import requests
>>> response = requests.get("https://www.stats.govt.nz/assets/Uploads/Business-operations-survey/Business-operations-survey-2023/Download-data/business-operations-survey-2023-innovation.csv", stream=True)
>>> response.raw.seek(0)
  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
io.UnsupportedOperation: seek
>>> type(response.raw)
<class 'urllib3.response.HTTPResponse'>
```

Note that we had discrepancies between streaming using requests_mock and actually hitting the API like [this](https://github.com/airbytehq/airbyte-python-cdk/pull/366) so unit tests would only work if they rely on starting an actual HTTP server

maxi297: Else, the solution I could see is to do like ResponseToFileExtractor does [here](https://github.com/airbytehq/airbyte-python-cdk/blob/e8ec2334f0f4b3728f8c1691f4a1ce4c70980feb/airbyte_cdk/sources/declarative/extractors/response_to_file_extractor.py#L98-L110) and consume the response by chunk but I don't know how feasible this is for outputting this as a BufferedIOBase

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2782,29 +2787,29 @@ def _get_download_retriever() -> SimpleRetrieverTestReadDecorator | SimpleRetrie
                 parameters={},
             )
 
-        decoder = (
+        operational_decoder = (
```
maxi297, line 2790: What does "operational" means here?

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -35,15 +35,36 @@ def parse(
...
+        IMPORTANT:
+            - If the data is not gzipped, reset the pointer and pass the data to the inner parser as is.
+
+        Note:
+            - The data is not decoded by default.
         """
-        with gzip.GzipFile(fileobj=data, mode="rb") as gzipobj:
-            yield from self.inner_parser.parse(gzipobj)
+
+        try:
```
maxi297, line 61: Should we add a test for this new logic?

maxi297 (commented):
A couple of concerns that I'm curious to have your input on

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
bazarnov: Fixed using headers checks, works as expected.

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
bazarnov: This logic is no longer present, the rest is covered with either existing tests or newly added one in:
- `unit_tests/sources/declarative/decoders/test_composite_decoder.py`
   - `test_composite_raw_decoder_gzip_jsonline_parser_decodes_non_gzipped_raw_response`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
bazarnov: I've reverted this change, it was probably not worth it.

airbyte_cdk/sources/declarative/decoders/composite_raw_decoder.py
```diff
@@ -24,6 +28,7 @@ class Parser(ABC):
     def parse(
         self,
         data: BufferedIOBase,
+        compressed: Optional[bool] = False,
```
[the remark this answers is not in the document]
bazarnov: Updated as proposed.

maxi297 (approved):
If CI passes, I'm all in on this change. Thanks for diving in the requests lib shenanigans with me ❤️

PR #387, opened 2025-03-06
Fix: State handling when passed in read() method

marianob-span1:
Low-code connectors instantiate the source without a state and pass the state via the read() method. This PR updates ConcurrentDeclarativeSource to support this behavior.

Changes are:
- Instantiate the ConnectorStateManager with the state passed in read() method.
- Update the ModelToComponentFactory to use the stream_state parameter when creating a cursor.
- Add a test to verify the behavior: instantiate the source without a state and pass the state via the read() method.

## Summary by CodeRabbit

- **New Features**
  - Enhanced state management during concurrent stream processing to ensure more consistent and reliable data synchronization.
  
- **Documentation**
  - Improved inline explanations to clarify how the `stream_state` variable is utilized in low-code scenarios.
  
- **Tests**
  - Introduced new tests verifying proper state handling during data extraction, reinforcing robust performance.

### DIFF
--- a/unit_tests/sources/declarative/test_concurrent_declarative_source.py
+++ b/unit_tests/sources/declarative/test_concurrent_declarative_source.py
@@ -848,7 +848,9 @@ def test_concurrent_cursor_with_state_in_read_method():
         streams=[
             ConfiguredAirbyteStream(
                 stream=AirbyteStream(
-                    name="party_members", json_schema={}, supported_sync_modes=[SyncMode.incremental]
+                    name="party_members",
+                    json_schema={},
+                    supported_sync_modes=[SyncMode.incremental],
                 ),
                 sync_mode=SyncMode.incremental,
                 destination_sync_mode=DestinationSyncMode.append,
@@ -856,21 +858,21 @@ def test_concurrent_cursor_with_state_in_read_method():
         ]
     )
 
-
     def get_source():
         return ConcurrentDeclarativeSource(
             source_config=_MANIFEST, config=_CONFIG, catalog=catalog, state=[]
         )
 
-
     source = get_source()
     with HttpMocker() as http_mocker:
         _mock_party_members_requests(http_mocker, _NO_STATE_PARTY_MEMBERS_SLICES_AND_RESPONSES)
-        messages_iterator = source.read(logger=source.logger, config=_CONFIG, catalog=catalog, state=state)
-    
+        messages_iterator = source.read(
+            logger=source.logger, config=_CONFIG, catalog=catalog, state=state
+        )
+
     # Get the first message only to get the stream running so that we can check the cursor
     first_message = next(messages_iterator)
-        
+
     concurrent_streams, _ = source._group_streams(config=_CONFIG)
     party_members_stream = [s for s in concurrent_streams if s.name == "party_members"][0]
 
@@ -880,17 +882,21 @@ def test_concurrent_cursor_with_state_in_read_method():
     assert isinstance(party_members_cursor, ConcurrentCursor)
     assert party_members_cursor._stream_name == "party_members"
     assert party_members_cursor._cursor_field.cursor_field_key == "updated_at"
-    
+
     cursor_value = AirbyteDateTime.strptime("2024-09-05", "%Y-%m-%d")
-    
+
     assert len(party_members_cursor._concurrent_state["slices"]) == 1
-    assert party_members_cursor._concurrent_state["slices"][0]["most_recent_cursor_value"] == cursor_value
-    
+    assert (
+        party_members_cursor._concurrent_state["slices"][0]["most_recent_cursor_value"]
+        == cursor_value
+    )
+
     messages = list(messages_iterator)
     party_members_records = get_records_for_stream("party_members", messages)
     # There is only one record after 2024-09-05
     assert len(party_members_records) == 1
     assert party_members_records[0].data["id"] == "yoshizawa"
+
 
 def test_check():
     """
--- a/unit_tests/sources/declarative/test_concurrent_declarative_source.py
+++ b/unit_tests/sources/declarative/test_concurrent_declarative_source.py
@@ -825,6 +825,73 @@ def test_create_concurrent_cursor():
     assert incremental_counting_cursor._end_provider() == math.inf
 
 
+@freezegun.freeze_time(_NOW)
+def test_concurrent_cursor_with_state_in_read_method():
+    """
+    This test mimicks the behavior of a source in a real life low-code connector
+    where the source is instantiated without a state and the state is provided
+    via the read() method.
+
+    Note: this test specifically checks that for DatetimeBasedCursor
+    """
+    state = [
+        AirbyteStateMessage(
+            type=AirbyteStateType.STREAM,
+            stream=AirbyteStreamState(
+                stream_descriptor=StreamDescriptor(name="party_members", namespace=None),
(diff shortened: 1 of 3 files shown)


unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -825,6 +825,80 @@ def test_create_concurrent_cursor():
...
+        source_config=_MANIFEST, config=_CONFIG, catalog=catalog, state=[]
+    )
+
+    with HttpMocker() as http_mocker:
+        _mock_party_members_requests(http_mocker, _NO_STATE_PARTY_MEMBERS_SLICES_AND_RESPONSES)
+        messages = list(
+            source.read(logger=source.logger, config=_CONFIG, catalog=catalog, state=state)
+        )
+
+    concurrent_streams, _ = source._group_streams(config=_CONFIG)
```
maxi297, line 871: This method is private and only called once per source hence this does not reflect the normal usage of a source. Could you please comment on what this is trying to achieve? 

The consequences of using this private method on the test seem important as the test read performed above is applied to another instance of `party_members_stream` and the read should not influence the state that is tested below. The reason the test passes right now is the read swap the instance of `_connector_state_manager` and the next time `_group_stream` is called, it uses the new state. 

In order to properly validate that the input state is taken into account, we would have to call `http_mocker._validate_all_matchers_called()` to ensure that the HTTP requests that were performed considered the state as an input.

maxi297 (commented):
Hi @marianob-span1! Thanks for the contribution. I would like to ask more information regarding this change to make sure I understand what would be the right solution. I'm asking because we have been slowly fading out the use of state through parameters to encapsulate the state into [the Cursor interface](https://github.com/airbytehq/airbyte-python-cdk/blob/b79177b5b7bc3f32f5bdf2c8b44dae7d9cd3e748/airbyte_cdk/sources/streams/concurrent/cursor.py#L50-L82). 

The reasons are:
* There are a bunch of maintenance problems that comes with multiple classes having access to the state:
    * When we are adding capabilities to state management, the change can affect all those classes and each usage needs to be validated which is long and tedious
    * It is very hard to debug what updates the state if the state is passed everywhere and we've had surprising behavior associated to that
* We are moving to use concurrency more and more to speed up our sources and having a dict passed around and being able to be modified by many component is not thread safe

Hence, my questions would be:
* What is the problem you are facing that brings you to create this PR? 
* Why passing the state at the construction of the source is now a valid solution in your case?

In the meanwhile, I still added a couple of comments on the PR to try to bridge the gap between my understanding and yours.

marianob-span1:
Hi @maxi297 !

The current issue we are experiencing with low code connectors and the CDK is that, when using incremental streams with a DatetimeBasedCursor, the connector ignores the previous stream state that is passed by the platform or a `--state` argument if I'm working on my local dev env and it performs a full refresh instead of continue reading from the point it left off last time.

The reason seems to be that the point where a low code connector passes the current state is in the read() call in `entrypoint.py`:

```python
elif cmd == "read":
    config_catalog = self.source.read_catalog(parsed_args.catalog)
    state = self.source.read_state(parsed_args.state)
    
    yield from map(
        AirbyteEntrypoint.airbyte_message_to_string,
        self.read(source_spec, config, config_catalog, state),
    )
```

So my changes introduced here where aimed at making sure that state passed in that read() call reaches the place where the Concurrent Cursor is instantiated, that's why I added the argument `stream_state` [here](https://github.com/airbytehq/airbyte-python-cdk/blob/cdde727dd1df4d8a20e594d97fa546a90e44dfda/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L272) and had to instantiate a new `ConnectorStateManager` at the start of `ConnectorStateManager.read()`.

Given your question and looking at the whole CDK code, it seems that it is expected that the state is passed when the source is instantiated (not 100% sure about this though), which I'm not sure how that could be done at least for low code connectors where the source is instantiated without any parameters, this is the `main.py` from the new connector where I ran into this issue:

```python
import sys

from airbyte_cdk.entrypoint import launch
from source_linear import SourceLinear

if __name__ == "__main__":
    source = SourceLinear()
    launch(source, sys.argv[1:])
```

It could also be the case that we are using the CDK in a way that it's not expected to be and that's why the state is blank at the moment the read operation occurs, but this happened in a low code connector recently generated using instructions found in here: https://docs.airbyte.com/connector-development/config-based/tutorial/create-source and I'm not aware of any other guides to create a low code connector. I'd love to get some guidance on how a low code connector should be structured if the issue is that there is something wrong with the connector itself

Thanks!
Mariano

unit_tests/sources/declarative/test_concurrent_declarative_source.py
marianob-span1: It's just a way to obtain the streams that I saw in other tests in this same file. I'm happy to find another way to get the streams in order to make the assertions I was intending to do if the general approach of the fix included in this PR is okay.

maxi297:
Ok, that makes a lot of sense. Thanks for sharing this context!

You can check how it is done for other low-code sources like [source-instagram](https://github.com/airbytehq/airbyte/blob/8aaaad470e5f2aac147bc7afc13411bd2608d6b7/airbyte-integrations/connectors/source-instagram/source_instagram/run.py#L18-L46). Don't forget to initialize the uncaught exception handler like [this](https://github.com/airbytehq/airbyte/blob/8aaaad470e5f2aac147bc7afc13411bd2608d6b7/airbyte-integrations/connectors/source-instagram/source_instagram/run.py#L50). If your source is all defined within a manifest and a components.py file for the custom components, I would suggest you move to the `manifest-only` format like [source-chargee](https://github.com/airbytehq/airbyte/tree/8aaaad470e5f2aac147bc7afc13411bd2608d6b7/airbyte-integrations/connectors/source-chargebee) as you would not have to worry about the instantiation and the state management. 

Let me know if this does not solve your issue and why and we will figure it out! 💪

marianob-span1:
Hi @maxi297 that reply is just gold! I wasn't aware of how the run() of a connector should be properly structured. 

I think for now we'll just replicate Instagram's structure given that it's easier for development purposes to just run `poetry run connector read --config secrets/config.json --catalog catalog.json` in a straightforward way while that using the source-declarative manifest I would have to build the connector image and then do a read operation via a `docker run` command, is that correct?

I'm closing the PR but will appreciate if you could reply to this last comment, thank you so much!

maxi297:
> it's easier for development purposes to just run poetry run connector read --config secrets/config.json --catalog catalog.json in a straightforward way while that using the source-declarative manifest I would have to build the connector image and then do a read operation via a docker run command

This is mostly true but not 100% true. Everything under the hood is Python so you could create a `main.py` which instantiate a `YamlDeclarativeSource` and you would have the same capabilities. We do that in our testing [here](https://github.com/airbytehq/airbyte/pull/54715/files#diff-846353691f46e8a007c3b7bc6f9989a057c9eac482556e3e5d3ab06eb97fab13R43-R44). Note that the weird logic for the manifest path [here](https://github.com/airbytehq/airbyte/pull/54715/files#diff-846353691f46e8a007c3b7bc6f9989a057c9eac482556e3e5d3ab06eb97fab13R26-R34) is for CI only meaning that the testing is done as part of the source-declarative-manifest container for manifest only sources hence why we check for both paths but if it's just for a local `main.py`, you should be fine with just the `manifest.yaml` file path.

If you have any other questions, please ping me and we will figure it out! ❤️

PR #388, opened 2025-03-06
feat: add formatting for epoch_microseconds

artem1205:
## What

Resolving https://github.com/airbytehq/airbyte/pull/54151

## How

add `%epoch_microseconds` to parser. Reason: concurrent curosor stateconverter uses parser to operate state values

## Summary by CodeRabbit

- **New Features**
  - Enhanced date handling by introducing a new timestamp format with microsecond-level precision for parsing and formatting dates.

- **Tests**
  - Updated test cases to validate the new microsecond timestamp conversion, ensuring reliable and accurate date representation.
  - Streamlined test function signatures and added new scenarios for parsing and formatting datetime values.
  - Adjusted test parameters to align with the new microsecond format.

### DIFF
--- a/airbyte_cdk/sources/declarative/interpolation/macros.py
+++ b/airbyte_cdk/sources/declarative/interpolation/macros.py
@@ -173,7 +173,7 @@ def format_datetime(
     )
     if format == "%s":
         return str(int(dt_datetime.timestamp()))
-    elif format == "%ms":
+    elif format == "%epoch_microseconds":
         return str(int(dt_datetime.timestamp() * 1_000_000))
     return dt_datetime.strftime(format)
 
--- a/airbyte_cdk/sources/declarative/interpolation/macros.py
+++ b/airbyte_cdk/sources/declarative/interpolation/macros.py
@@ -12,6 +12,8 @@ import pytz
 from dateutil import parser
 from isodate import parse_duration
 
+from airbyte_cdk.sources.declarative.datetime.datetime_parser import DatetimeParser
+
 """
 This file contains macros that can be evaluated by a `JinjaInterpolation` object
 """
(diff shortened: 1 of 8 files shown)


airbyte_cdk/sources/declarative/interpolation/macros.py
```diff
@@ -173,7 +173,7 @@ def format_datetime(
     )
     if format == "%s":
         return str(int(dt_datetime.timestamp()))
-    elif format == "%ms":
```
maxi297, line 176: I'm confused about this change ; shouldn't we continue supporting `%ms`? Doesn't that indicate that there is a bug in the formating for epoch millis as it would return an epoch microseconds string representation?

maxi297 (commented):
It feels like there is a problem somewhere but I think it is in the formatting of epoch millis aka %ms. I'm kind of worried because it used in a couple connectors today (dixa, appcues, front, lever-hiring, linkedin-pages, mixmax...) and in most cases, it would mean that we send a epoch microsecond to the API instead and fixing that would be a breaking change

airbyte_cdk/sources/declarative/interpolation/macros.py
artem1205: it was added in https://github.com/airbytehq/airbyte-python-cdk/pull/373/files

but then I realised that we already have "%ms" placeholder that represents **milli**seconds instead of **micro**seconds.
To avoid confusion I decided to name them explicitly: `%epoch_microseconds`

maxi297: Would it help maintenance and reduce confusion if `format_datetime` would call `DatetimeParser.format`? It would also ensure that the same formatting capabilities are the same everywhere

artem1205: agreed, refactored

maxi297 (approved):
LGTM! Thanks for the improvement too

PR #391, opened 2025-03-06
feat: AsyncRetriever: Enable configurability of max concurrent job count

pnilan:
Closes: https://github.com/airbytehq/airbyte-internal-issues/issues/11343

## What
- Adds `max_concurrent_jobs` field to `AsyncRetriever` component.
    - Can take either an `integer` or interpolated string
    - Min count of 1
- Updates and adds tests for interpolation and validation

## Summary by CodeRabbit

- **New Features**
  - Enhanced job tracking by allowing the job limit to be set dynamically with either a number or a configurable string value.
  - Introduced a new, configurable maximum concurrent jobs setting that lets users fine-tune performance based on their processing requirements.

- **Bug Fixes**
  - Added validation to ensure the job limit is at least 1, raising an error if this condition is not met.

- **Tests**
  - Updated tests to reflect changes in the `JobTracker` initialization, ensuring proper handling of new parameters and configurations.


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3376,6 +3376,16 @@ definitions:
           - "$ref": "#/definitions/IterableDecoder"
           - "$ref": "#/definitions/XmlDecoder"
           - "$ref": "#/definitions/ZipfileDecoder"
+      max_concurrent_jobs:
+        title: Maximum Conccurent Job Count
+        description: Maximum number of concurrent jobs to run.
```
natikgadzhi, line 3381: Nit: it'd be useful to explain what those jobs do — are those parsing already downloaded archives, or are those threads that request multiple reports in parallel? etc

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3376,6 +3376,16 @@ definitions:
...
+        anyOf:
+          - type: integer
+          - type: string
```
natikgadzhi, line 3384: I'm not convinced you _really_ need this in config, and therefore you need strings here. But, up to you.

unit_tests/sources/declarative/partition_routers/test_async_job_partition_router.py
```diff
@@ -58,7 +58,7 @@ def test_stream_slices_with_parent_slicer():
         job_orchestrator_factory=lambda stream_slices: AsyncJobOrchestrator(
             MockAsyncJobRepository(),
             stream_slices,
-            JobTracker(_NO_LIMIT),
+            JobTracker(_NO_LIMIT, config={}),
```
natikgadzhi, line 61: Seems like JobTracker already has config={} default in the initializer? Could avoid changes here I think?

natikgadzhi (approved):
Nothing stands out as wrong, so approving to unblock.

bazarnov (commented):
Everything looks good to me, but I think it would be a good idea to add a dedicated unit test that tests more than one `max_concurrent_jobs`. For example, we could test 2-3 `max_concurrent_jobs` to ensure that we process the records correctly. What do you think?

pnilan:
Closing in favor of source-scoped configurability: https://github.com/airbytehq/airbyte-python-cdk/pull/405

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
pnilan: Yeah, I think you're right. Looks like Salesforce defines a static max across all accounts.

PR #392, opened 2025-03-07
fix(concurrent-cdk): StreamPartition handles AirbyteRecords

girarda:
Update the StreamPartition so it properly handles `AirbyteRecord`s.

On master, we send all AirbyteMessages on the repository, which is wrong if the message is an `AirbyteRecord`. We should instead create a record from it and yield it.

## Summary by CodeRabbit

- **New Features**
  - Enhanced message processing to reliably extract records from `AirbyteMessage` objects.
- **Tests**
  - Introduced a test to validate the behavior of the `StreamPartition` class when reading `AirbyteMessage` records.

### DIFF
--- a/airbyte_cdk/sources/streams/concurrent/adapters.py
+++ b/airbyte_cdk/sources/streams/concurrent/adapters.py
@@ -276,7 +276,7 @@ class StreamPartition(Partition):
     def read(self) -> Iterable[Record]:
         """
         Read messages from the stream.
-        If the StreamData is a Mapping, it will be converted to a Record.
+        If the StreamData is a Mapping or an AirbyteMessage of type RECORD, it will be converted to a Record.
         Otherwise, the message will be emitted on the message repository.
         """
         try:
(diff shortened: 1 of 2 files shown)


natikgadzhi (approved):
Nit: can you update the docstring of read_records too? It seems outdated.

airbyte_cdk/sources/streams/concurrent/adapters.py
```diff
@@ -302,6 +302,12 @@ def read(self) -> Iterable[Record]:
                         stream_name=self.stream_name(),
                         associated_slice=self._slice,  # type: ignore [arg-type]
                     )
+                elif isinstance(record_data, AirbyteMessage) and record_data.record is not None:
```
maxi297, line 305: I'm not sure how this works with the file based universal file transfer records. It seems like right now, we are excluding those from this condition. Is that expected?

girarda: file based connectors have their own [facade](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/file_based/stream/concurrent/adapters.py). I'll add a comment mentioning we need to support them here if we want to reconcile the facades.

PR #395, opened 2025-03-07
Proposal for generic composite raw decoder

maxi297:
Proposal following compression logic added in https://github.com/airbytehq/airbyte-python-cdk/pull/378


maxi297:
Closing as the proposal has been applied in another PR

PR #402, opened 2025-03-10
feat(file-based): new AbstractFileBasedStreamPermissionsReader

aldogonzalez8:
## What

Introduced an `AbstractFileBasedStreamPermissionsReader` class to handle permission-related tasks for file-based streams.

This should fix to have to make noops like [here](https://github.com/airbytehq/airbyte/pull/55202/files) ISP!! 

I already have two PRs waiting for this change:

- [Source Google Drive](https://github.com/airbytehq/airbyte/pull/55689)
- [Source Microsoft Sharepoint](https://github.com/airbytehq/airbyte/pull/54175/files)

## How

- Created an abstract base class, `AbstractFileBasedStreamPermissionsReader`, to encapsulate permission-checking logic.
- This class defines methods that subclasses must implement to verify necessary permissions for file-based streams.
- Refactored existing code to utilize this new class, promoting cleaner and more maintainable code by separating permission logic from other functionalities.

# Review guide

Here are the relevant files to review:

1. `airbyte_cdk/sources/file_based/file_based_source.py`:
   - Integrates the new permission reader into the file-based source logic.
2. `airbyte_cdk/sources/file_based/file_based_stream_permissions_reader.py`:
   - Defines the `AbstractFileBasedStreamPermissionsReader` and its core permission-handling logic.
3. `airbyte_cdk/sources/file_based/file_based_stream_reader.py`:
   - Updates the stream reader to delegate permission verification to the new permission reader class.
4. `airbyte_cdk/sources/file_based/stream/identities_stream.py`:
   - Modifies the identities stream to align with the new permission-handling structure.
5. `airbyte_cdk/sources/file_based/stream/permissions_file_based_stream.py`:
   - Introduces or modifies the permission-based stream functionality to work with the new reader.

These files reflect the core changes introduced in this pull request, focusing on the integration of the `AbstractFileBasedStreamPermissionsReader` across relevant file-based stream implementations.

## User Impact

Users should experience no changes in functionality. This refactoring is internal and does not alter any external interfaces or behaviors.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated mechanism for handling file permissions in file-based streams.

- **Refactor**
  - Unified stream identity and permissions functionalities by integrating the new permissions management.
  - Removed legacy methods to streamline permission-related processes.

- **Tests**
  - Updated test suites to validate the new permissions handling and ensure consistent stream behavior.


airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -100,8 +103,10 @@ def __init__(
         cursor_cls: Type[
             Union[AbstractConcurrentFileBasedCursor, AbstractFileBasedCursor]
         ] = FileBasedConcurrentCursor,
+        stream_permissions_reader: Optional[AbstractFileBasedStreamPermissionsReader] = None,
```
aaronsteers, line 106: Sidebar: We do a LOT of dependency injection via constructor args. The same can be accomplished with class members with less complexity and with looser coupling between classes constructing objects of other classes. Sometime I'd like to discuss a design proposal where: 

1. The class defines a default class (or None) for special functionality. 
2. Subclasses can override that class in their own declarations to provide their own custom implementations. 
3. Then, constructor args are less needed. They can optionally be passed (in class or object form) but they aren't needed, per se, as they are today. 

I'm probably not explaining this well but I'll make a note to bring this up in future dialogue. (No action needed as of now.)

To be specific though, we could add a member to this class with the same name, except suffixed with "_type" or "_class", with the `| None = None` typing because this is optional. Then, in the base class constructor, we check if our type is None, initializing it if not. This approach doesn't require expanding the constructor input args. If we _really_ want to also accept an input arg in the constructor, we can do so, and in that case it works override the class-level behavior, if defined.

airbyte_cdk/sources/file_based/file_based_stream_permissions_reader.py
```diff
@@ -0,0 +1,131 @@
...
+            for group in groups_api.list():
+                group_obj = my_identity_model(id=group.id, name=groups.name, email_address=user.email, type="group").dict()
+                for member in members_api.list(group=group):
+                    group_obj.member_email_addresses = group_obj.member_email_addresses or []
+                    group_obj.member_email_addresses.append(member.email)
+                yield group_obj.dict()
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not implement load_identity_groups(). To support identities, implement this method and update identities_schema."
+        )
```
aaronsteers, line 80: Nit: Since this is an abstract method, the `raise` statement is basically impossible to reach. You can instead just use `...` as the method body. 

Another approach is to remove the `abstractmethod` decorator and allow classes to be created without defining an implementation. This approach is helpful if the class can (in theory) be valid without implementing the respective behaviors - such as when the code path is not strictly required for all implementations.

PR #405, opened 2025-03-10
feat: Enable JobTracker limit configurability

pnilan:
## What
Enables source-level job tracking and job count limit for async jobs managed by each stream's `AsyncJobOrchestrator`.

- Adds new optional `max_concurrent_async_job_count` top-level property in manifest and passes to `ModelToComponentFactory` where a `JobTracker` component is instantiated in its conctructor w/ the defined job count value.
- Closes https://github.com/airbytehq/airbyte-internal-issues/issues/11343

## Recommended Review Order
1. `declarative_component_schema.yaml`
2. `manifest_declarative_source.py`
3. `model_to_component_factory.py`
4. `job_tracker.py`

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration option, "Maximum Concurrent Async Jobs", allowing users to specify the maximum number of asynchronous jobs.
  - Enhanced the handling of job tracking to ensure consistent usage of the same tracker instance across components.

- **Bug Fixes**
  - Improved job limit handling by enforcing a minimum value of 1, preventing scenarios with invalid job limits.

- **Tests**
  - Added new test coverage for the `JobTracker` class to validate behavior when initialized with limits less than 1.

### DIFF
--- a/airbyte_cdk/sources/declarative/async_job/job_tracker.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_tracker.py
@@ -17,10 +17,6 @@ class ConcurrentJobLimitReached(Exception):
 class JobTracker:
     def __init__(self, limit: int):
         self._jobs: Set[str] = set()
-        if limit < 1:
-            LOGGER.warning(
-                f"The `max_concurrent_async_job_count` property is less than 1: {limit}. Setting to 1. Please update the source manifest to set a valid value."
-            )
         self._limit = 1 if limit < 1 else limit
         self._lock = threading.Lock()
 
--- a/airbyte_cdk/sources/declarative/async_job/job_tracker.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_tracker.py
@@ -3,7 +3,7 @@
 import logging
 import threading
 import uuid
-from typing import Set
+from typing import Optional, Set
 
 from airbyte_cdk.logger import lazy_log
 
@@ -15,9 +15,9 @@ class ConcurrentJobLimitReached(Exception):
 
 
 class JobTracker:
-    def __init__(self, limit: int):
+    def __init__(self, limit: Optional[int]):
         self._jobs: Set[str] = set()
-        self._limit = limit
+        self._limit = limit if isinstance(limit, int) and limit >= 1 else 1
         self._lock = threading.Lock()
 
     def try_to_get_intent(self) -> str:
--- a/airbyte_cdk/sources/declarative/async_job/job_tracker.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_tracker.py
@@ -17,6 +17,10 @@ class ConcurrentJobLimitReached(Exception):
 class JobTracker:
     def __init__(self, limit: int):
         self._jobs: Set[str] = set()
+        if limit < 1:
+            LOGGER.warning(
+                f"The `max_concurrent_async_job_count` property is less than 1: {limit}. Setting to 1. Please update the source manifest to set a valid value."
+            )
         self._limit = 1 if limit < 1 else limit
         self._lock = threading.Lock()
 
--- a/airbyte_cdk/sources/declarative/async_job/job_tracker.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_tracker.py
@@ -3,7 +3,7 @@
 import logging
 import threading
 import uuid
-from typing import Optional, Set
+from typing import Set
 
 from airbyte_cdk.logger import lazy_log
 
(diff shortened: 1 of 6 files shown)


pnilan:
Grooming note:
- verify api budget instantiation location

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1898,6 +1903,11 @@ class Config:
     spec: Optional[Spec] = None
     concurrency_level: Optional[ConcurrencyLevel] = None
     api_budget: Optional[HTTPAPIBudget] = None
+    max_concurrent_async_job_count: Optional[int] = Field(
+        None,
+        description="Maximum number of concurrent async jobs to run. This property is only relevant for sources/streams that support asynchronous job execution (e.g. a Report stream that requires that initiates a job, polls the job status, and then fetches the job results). This is often set by the API's maximum number of concurrent jobs on the account level.",
```
maxi297, line 1908: nit: Can we reference the AsyncRetriever here? Something like `that support asynchronous job execution through the AsyncRetriever`

airbyte_cdk/sources/declarative/async_job/job_tracker.py
```diff
@@ -17,7 +17,7 @@ class ConcurrentJobLimitReached(Exception):
 class JobTracker:
     def __init__(self, limit: int):
         self._jobs: Set[str] = set()
-        self._limit = limit
+        self._limit = 1 if limit < 1 else limit
```
maxi297, line 20: nit: Can we have a warning log here to explain that we aren't using the limit that was provided?

maxi297 (approved):
Tagging @bazarnov to have its input on how we expect to handle this in the Connector Builder and if we should change things regarding this. I don't think it should be a blocker here but I would prefer to consider this earlier than later.

The rest looks good to me!

bazarnov (approved):
Looks good to me! Just a couple of dummy questions from my side @pnilan.

bazarnov:
> Tagging @bazarnov to have its input on how we expect to handle this in the Connector Builder and if we should change things regarding this. I don't think it should be a blocker here but I would prefer to consider this earlier than later.
> 
> The rest looks good to me!

Looks like passing the value from the `self.max_concurrent_jobs` to the `JobTracker` sets the limit correctly, thus should be fine when the `max_concurrent_jobs` is available in Builder. Since we expect this to be set `per source` - we should be fine.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2928,8 +2930,7 @@ def _get_download_retriever() -> SimpleRetrieverTestReadDecorator | SimpleRetrie
             job_orchestrator_factory=lambda stream_slices: AsyncJobOrchestrator(
                 job_repository,
                 stream_slices,
-                JobTracker(1),
-                # FIXME eventually make the number of concurrent jobs in the API configurable. Until then, we limit to 1
+                self._job_tracker,
```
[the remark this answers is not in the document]
pnilan: I believe a new AsyncRetriever is instantiated for each stream that has one, so therefore a new JobTracker would be instantiated per stream. Moving the instantiation to the ModelToComponentFactory constructor ensures that the self._job_tracker would track jobs across all streams that use it.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -540,6 +541,7 @@ def __init__(
         )
         self._connector_state_manager = connector_state_manager or ConnectorStateManager()
         self._api_budget: Optional[Union[APIBudget, HttpAPIBudget]] = None
+        self._job_tracker: JobTracker = JobTracker(max_concurrent_async_job_count or 1)
```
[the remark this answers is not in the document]
pnilan: See previous comment instantiation one JobTracker per source.

For example, Salesforce limits jobs on the account-level (not endpoint level) so we would need to track jobs across streams:

> The system can process up to five queued or active jobs simultaneously for each organization. 

[Source](https://trailhead.salesforce.com/content/learn/modules/asynchronous_apex/async_apex_monitoring#monitoring-queued-jobs-with-soql)

My thought here is if we *do* run into sources that limit per-stream, then we can add in a stream-scoped job tracker.

airbyte_cdk/sources/declarative/async_job/job_tracker.py
pnilan: Added.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
pnilan: Added.

PR #407, opened 2025-03-11
fix: (CDK) (manifest-only) - add ability to debug `manifest-only` sources / destinations

bazarnov:
## What
Frequently, we had cases when we need/have to `debug` a `manifest-only` source/destination, but we don't have a fast and relatively easy way of doing that to reproduce the issue or introduce new features to the `declarative_component_schema`, etc.

## How
- Previously, debugging `manifest-only` sources and destinations required creating a copy of the `source-declarative-manifest,` modifying the `pyproject.toml`, installing an environment, and ensuring compatibility with versioning of the manifest and other related elements.
- With this change, we have gained a highly flexible approach to debugging `manifest-only` sources and destinations by providing only the necessary components, such as `config, catalog, state` and the `manifest` itself.
- Furthermore, the ability to configure the debugger based on your IDE further enhances this capability, allowing you to utilize `breakpoints` and gain a comprehensive understanding of the runtime state to identify and resolve any issues effectively.
- see the `README.md` for more information.

## Summary by CodeRabbit

- **New Features**
  - Introduced a debug utility that enables manifest debugging with customizable command-line parameters.

- **Documentation**
  - Added detailed guidance for setting up and running debugging sessions within your IDE, covering necessary configuration and resource management.

- **Chores**
  - Improved repository file management to exclude extraneous files while preserving essential directory structures.


bazarnov:
Automation ideas for the future:
- autogenerated `catalog`
- extract `testing values` > `config.json` (or similar) during the `export manifest` in Builder

Ideally, it would be very handy if the export functionality had a feature called `Export for Debugging` (for admins only) that comes as a `zip` package with `manifest + config`, which we can directly use while debugging.

PR #409, opened 2025-03-12
feat(low-code cdk): add condition to add fields

lazebnyi:
Fixed : https://github.com/airbytehq/airbyte-internal-issues/issues/11938

## Summary by CodeRabbit

- **New Features**
	- Introduced a new conditional logic option for adding fields, allowing record transformations to include extra fields only when defined expressions evaluate to true. This enhancement provides users with greater flexibility in tailoring how and when fields are applied based on dynamic conditions.


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -110,6 +110,19 @@ definitions:
         type: array
         items:
           "$ref": "#/definitions/AddedFieldDefinition"
+      condition:
+        description: Fields will be added if expression is evaluated to True.,
```
bazarnov, line 114: nit: remove trailing `,`

bazarnov (approved):
Looks good!

PR #412, opened 2025-03-12
fix: (CDK) (HttpRequester) - fix `trailing slash` for `url_base` when no `path` has been provided

bazarnov:
## What
Resolving regression after:
- https://github.com/airbytehq/airbyte-python-cdk/pull/370

Resolving OC:
- https://github.com/airbytehq/oncall/issues/7584

## How
- moved from using the `os.path.join()` in the `_join_url()` method, use it when the `path` is provided and there is no `/` at the end of the `url_base` to join the URL parts correctly

## User Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the construction and formatting of URLs to ensure consistent handling of trailing slashes, resulting in more predictable URL outputs.
- **Tests**
  - Updated verification checks to align with the refined URL formatting, ensuring that all endpoints now meet the expected behavior.


airbyte_cdk/sources/declarative/requesters/http_requester.py
```diff
@@ -370,13 +370,18 @@ def _join_url(cls, url_base: str, path: str) -> str:
...
 
         # return a full-url if provided directly from interpolation context
         if path == EmptyString or path is None:
-            return url_base.rstrip("/")
+            return url_base
+        else:
+            # since we didn't provide a full-url, the url_base might not have a trailing slash
+            # so we join the url_base and path correctly
+            if not url_base.endswith("/"):
+                url_base = os.path.join(url_base, EmptyString)
```
lmossman, line 384: @bazarnov is this whole `else` block even necessary? It seems to me like we have two cases:
1. Only a URL base is provided
2. Both a URL base and a non-empty path are provided
(notice I don't care about any slashes in either case)

The `if` block handles the first case, so that if no path is provided, then we return url_base as-is.

For the second case, it seems like we could just simply return `urljoin(url_base, path)` to combine the two, and the `url_join` should properly handle slashes in between the two, right?

So why do we need this `else` block at all?

bazarnov: If both: `url_base` doesn't have the `trailing slash` or/and the `path` doesn't have the `leading slash` - it would not join the URL parts correctly, meaning that we have to set the `/` in either case inside the `else` block, which we check and do. Otherwise, there might be more issues.

lmossman: Oh okay, I had assumed `urljoin` would place a slash between the two in that case. Is there a reason we don't just put that logic into `urljoin`?

lmossman: Ah, urljoin isn't our method, looks like it comes from urllib. Can you double check if it has that logic baked in or not?

bazarnov: The thing is:

```python
urljoin('http://example.com/test', 'page') # >> 'http://example.com/page'
```

this method swallows the middle part of `/test`, and doesn't add the `trailing slash` automatically, which is non-expected behavior to me. I'll investigate.

bazarnov: The issue lies in how the `urllib.parse.urljoin()` function handles the `base path` and the relative `path`. Specifically, the function removes the last segment of the base path if it is not a directory. In your case, 'http://example.com/test' is treated as a file, not a directory, so the `/test` part is removed.

To fix this, you can ensure that the base URL ends with a slash (/) to indicate that it is a directory. The fix looks like: 
```python
urljoin('http://example.com/test/', 'page') # >> 'http://example.com/test/page'
```

bazarnov: I'd leave that else block as the minimal fix required to make everything work as expected @lmossman

lmossman: Got it, your logic makes sense to me then.

One more question: if the path has a slash at the beginning, e.g. `urljoin('http://example.com/test/', '/page')`, does this properly avoid a double-slash, i.e. it produces `http://example.com/test/page`?

bazarnov: Guess what? We remove it [here](https://github.com/airbytehq/airbyte-python-cdk/blob/649c6d03481ecab06e437c23566ea2ea9fea3e4f/airbyte_cdk/sources/declarative/requesters/http_requester.py#L150), beforehand)

lmossman (approved):
Logic makes sense to me, though I did not test this myself

PR #413, opened 2025-03-12
feat(low-code cdk): add transform_before_filtering to partition router schema

lazebnyi:
Fixed: https://github.com/airbytehq/airbyte-internal-issues/issues/11933

## Summary by CodeRabbit

- **New Features**  
  - Added a configurable option that lets users choose to apply data transformations before filtering records, providing enhanced control over the data processing order.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2564,7 +2564,9 @@ class ModelToComponentFactory:
             else None
         )
 
-        transform_before_filtering = False
+        assert model.transform_before_filtering is not None  # for mypy
+
+        transform_before_filtering = model.transform_before_filtering
         if client_side_incremental_sync:
             record_filter = ClientSideIncrementalRecordFilterDecorator(
                 config=config,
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2564,7 +2564,9 @@ def create_record_selector(
             else None
         )
 
-        transform_before_filtering = False
+        assert model.transform_before_filtering is not None  # for mypy
```
bazarnov, line 2567: Can we avoid using `assert` here, it's great we can use it, but if it's just to satisfy the `MyPy` checker - let's use `# type: ignore` instead. WDYT?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2564,7 +2564,9 @@ def create_record_selector(
...
+
+        transform_before_filtering = model.transform_before_filtering
```
bazarnov, line 2569: nit: `model.transform_before_filtering or False`

bazarnov (approved):
Looks good to me, feel free to ignore the comments if it's blocking you in certain ways.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
lazebnyi: Yes, we have same parts throughout this file. I think we need to replace them all at once. For now, let’s leave it as is.

PR #417, opened 2025-03-12
fix(file): Support column named type

maxi297:
Addresses https://airbytehq-team.slack.com/archives/C03GD9SV36E/p1741359423619149

Basically, a user is having issue when the column name is the reserved word `type`. We can support that if we check the type of the value when we see the key `type` during the schema inference.

Discovered catalog before:
image

Discovered catalog after: 
image

## Summary by CodeRabbit

- **Refactor**
	- Enhanced data processing to more robustly handle missing values within various data configurations, including improvements for nested data structures. This update ensures improved data normalization and compatibility for file-based integrations.


aldogonzalez8 (approved):
APPROVED

PR #418, opened 2025-03-13
feat(low-code cdk): add lazy read to simple retriver

lazebnyi:
Fixed: https://github.com/airbytehq/airbyte-internal-issues/issues/11473?issue=airbytehq%7Cairbyte-internal-issues%7C11942

This PR introduces `LazySimpleRetriever`, built on top of `SimpleRetriever` with a new `lazy_read_pointer` field. When this field is set, the retriever switches to a lazy read mode. This means all full refresh syncs and incremental initial reads will follow the lazy flow, reducing unnecessary child requests and improving performance — especially for streams with limited child data exposure (e.g., Stripe `bank_accounts`).

## Summary by CodeRabbit

- **New Features**
	- Added an optional incremental sync option to enhance data retrieval with flexible cursor support.
	- Introduced a lazy-loading retrieval mechanism that efficiently handles parent-child record relationships with improved pagination and response safety.
	- Added a new property for lazy reading functionality in the schema definitions.
	- Expanded the public interface by including the new `LazySimpleRetriever` class for improved data retrieval capabilities.
	- Added unit tests to validate lazy loading and incremental sync behaviors.

- **Bug Fixes**
	- Improved error handling for invalid configurations related to lazy reading pointers.

- **Documentation**
	- Updated schema definitions to include new properties and their intended functionalities.


airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
```diff
@@ -131,6 +143,42 @@ def _get_request_option(
                         parent_config.request_option.inject_into_request(params, value, self.config)
         return params
 
+    def process_parent_record(
```
maxi297, line 146: I'm wondering if we could have kept this internal to SubstreamPartitionRouter if we would have had the `lazy_read_pointer` on the SubstreamPartitionRouter level and it could have set extra_fields as:
```
{
  <partition_field>: <partition_value>,
  "parent_slice": <parent_partition>,
  "extra_fields": { "child_response": <response> }
}
```

The benefits would be that:
* we can leverage concurrency which would be useful in terms of performance if many slices require pagination
* the logic of looping over the stream slices would still be hidden as part of the SubstreamPartitionRouter and therefore probably easier to maintain

We would still need to have a new retriever though but it would analyze the first page as being `slice.extra_fields.child_response`.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -3242,6 +3242,15 @@ definitions:
           - "$ref": "#/definitions/IterableDecoder"
           - "$ref": "#/definitions/XmlDecoder"
           - "$ref": "#/definitions/ZipfileDecoder"
+      lazy_read_pointer:
```
maxi297, line 3245: I really like the fact that this is decoupled from the Python implementation as adding a retriever has caused development in the Connector Builder in the past ❤️

maxi297 (commented):
The logic seems good to me but I'm wondering if we can still use stream slices to leverage the concurrency of our framework

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2534,6 +2541,7 @@ def create_parent_stream_config(
             incremental_dependency=model.incremental_dependency or False,
             parameters=model.parameters or {},
             extra_fields=model.extra_fields,
+            lazy_read_pointer=model.lazy_read_pointer
```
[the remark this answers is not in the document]
lazebnyi: This handle in __post_init__

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -618,3 +634,166 @@ def _fetch_next_page(
...
+    partition_router: SubstreamPartitionRouter = field(init=True, repr=False, default=None)  # type: ignore[assignment]  # 'partition_router' is required for LazySimpleRetriever and is validated in the constructor
+    lazy_read_pointer: Optional[List[InterpolatedString]] = None
+
+    def _read_pages(
+        self,
+        records_generator_fn: Callable[[Optional[requests.Response], Optional[StreamSlice]], Iterable[Record]],
+        stream_state: Mapping[str, Any],
+        stream_slice: StreamSlice,
+    ) -> Iterable[Record]:
+        parent_stream_config = self.partition_router.parent_stream_configs[-1]
```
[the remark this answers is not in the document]
lazebnyi: I have removed this as we don't need `parent_stream_config` values in lazy retriever anymore

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -618,3 +634,166 @@ def _fetch_next_page(
...
             ),
         )
+
+
+class SafeResponse(requests.Response):
+    def __getattr__(self, name: str) -> Any:
+        return getattr(requests.Response, name, None)
+
+    @property
+    def content(self) -> Optional[bytes]:
```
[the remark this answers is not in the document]
lazebnyi: I have added validation in the constructor that LazySimpleRetriever support only `JsonDecoder`.

airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
lazebnyi: Nice proposal!

I’ve moved child response handling to `partition_router`, update `LazySimpleRetriever` and move `lazy_read_pointer` to `parent_config`. Now, if a user has multiple sources/parents for child records, they can set up lazy_read_pointer for each one individually.

airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
```diff
@@ -381,3 +421,16 @@ def get_stream_state(self) -> Optional[Mapping[str, StreamState]]:
     @property
     def logger(self) -> logging.Logger:
         return logging.getLogger("airbyte.SubstreamPartitionRouter")
+
+
+class SafeResponse(requests.Response):
```
maxi297, line 426: nit: can we add a description of what is this object? `Safe` doesn't tell me much in this case.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -618,3 +630,74 @@ def _fetch_next_page(
...
+    """
+
+    def _read_pages(
+        self,
+        records_generator_fn: Callable[[Optional[requests.Response]], Iterable[Record]],
+        stream_state: Mapping[str, Any],
+        stream_slice: StreamSlice,
+    ) -> Iterable[Record]:
+        response = stream_slice.extra_fields["child_response"]
+        print(f"LOL, {response.json()}")
```
maxi297, line 652: LOL

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -618,3 +630,74 @@ def _fetch_next_page(
...
+                yield from self._paginate(
+                    next_page_token,
+                    records_generator_fn,
+                    stream_state,
+                    stream_slice,
+                )
+
+            yield from []
+        else:
+            yield from self._read_pages(records_generator_fn, stream_state, stream_slice)
```
maxi297, line 671: I'm confused: In which case can we end up in this else statement?

maxi297 (approved):
I'm fine with this change. I think it is a great and simple implementation for a feature that might seem complex at first. 

I just added a couple of comments to understand some stuff but I'm not blocking the release on these

airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py
lazebnyi: Added doc string

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
maxi297: How does that work? I assume lazy read only works with one parent with the current logic. With board parents, I would assume it is a mistake from the connector developer or it is a case we just don't support

PR #419, opened 2025-03-13
fix: (CDK) (AsyncRetriever) - fix `availability strategy` issues, during `check connection`

bazarnov:
## What
Original Slack conversation: https://airbytehq-team.slack.com/archives/C07JMAAE620/p1741825218622419

Problem Statement:
- Attempting to utilize the name of an asynchronous stream in CheckStream resulted in an error during the check request process when attempting to establish a Source for that manifest.
- The issue is associated with the `availability strategy` that searches for the `retriever.requester.exit_on_rate_limits`, which is not available for the `AsyncRetriever`.

## How
- created dedicated property methods to expose the `creation_requester.exit_on_rate_limits` and re-use it

## User Impact
With this update, customers who utilize the `AsyncRetriever` will now have the capability to specify the asynchronous stream as the one for checking the connection. This was not the case previously.

## Summary by CodeRabbit

- **New Features**
  - Introduced an asynchronous retrieval mode with enhanced rate limit handling, providing improved flexibility in managing job creation when limits are reached.
  
- **Documentation**
  - Updated guidance and explanations around rate limit management in asynchronous operations for clearer understanding.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_stream.py
+++ b/airbyte_cdk/sources/declarative/declarative_stream.py
@@ -14,6 +14,7 @@ from airbyte_cdk.sources.declarative.incremental import (
 from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
 from airbyte_cdk.sources.declarative.migrations.state_migration import StateMigration
 from airbyte_cdk.sources.declarative.retrievers import SimpleRetriever
+from airbyte_cdk.sources.declarative.retrievers.async_retriever import AsyncRetriever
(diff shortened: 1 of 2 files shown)


maxi297 (approved):
LGTM! I've started a brainstorm [here](https://airbytehq-team.slack.com/archives/C07JMAAE620/p1741876588994959?thread_ts=1741825218.622419&cid=C07JMAAE620) to try to find a way for us to avoid to do this for every retrievers we add

PR #429, opened 2025-03-18
fix: (CDK) (AsyncRetriever) - fix bug when `TIMEOUT` Jobs are retried, ignoring the `polling_job_timeout` setting

bazarnov:
## What
Original Slack thread:
- https://airbytehq-team.slack.com/archives/C07JMAAE620/p1742257344979549?thread_ts=1742236041.905529&cid=C07JMAAE620

Additional side-fix for the:
- https://github.com/airbytehq/airbyte/pull/55745#issuecomment-2731139288

## How
- make the `TIMEOUT` Job status to be `Terminal`
- make the `TIMEOUT` Job status non-retriable

- add more interpolation context required by certain sources, like `source-amazon-ads`, relying on `creation_response['headrs']['My-Header-Value']`. Also valid for the `polling_response` case.

## User Impact
No impact is expected, this is not a Breaking Change.

## Summary by CodeRabbit

- **New Features**
  - Improved job management now distinguishes between failed and timed-out jobs, ensuring more precise job reallocation and clearer error notifications when timeouts occur.
  - Centralized handling of job responses streamlines the processing of creation and polling data.

- **Tests**
  - Updated testing ensures that timeout errors provide accurate and informative feedback.

PR #432, opened 2025-03-19
fix: (CDK) (AsyncRetriever) - fix the regression. The TIMEOUT Job Status should Retry on server status

bazarnov:
## What 
Given this dialog: https://airbytehq-team.slack.com/archives/C07JMAAE620/p1742346710533189?thread_ts=1742236041.905529&cid=C07JMAAE620

Related to:
- https://github.com/airbytehq/airbyte-python-cdk/pull/429

### DIFF
--- a/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
+++ b/airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
@@ -44,16 +44,21 @@ class AsyncPartition:
     This bucket of api_jobs is a bit useless for this iteration but should become interesting when we will be able to split jobs
     """
 
-    _MAX_NUMBER_OF_ATTEMPTS = 3
+    _DEFAULT_MAX_JOB_RETRY = 3
 
-    def __init__(self, jobs: List[AsyncJob], stream_slice: StreamSlice) -> None:
+    def __init__(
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
```diff
@@ -400,7 +403,7 @@ def _remove_completed_or_timed_out_jobs(self, partition: AsyncPartition) -> None
             partition (AsyncPartition): The partition to process.
         """
         for job in partition.jobs:
-            if job.status() in [AsyncJobStatus.COMPLETED, AsyncJobStatus.TIMED_OUT]:
+            if job.status() == AsyncJobStatus.COMPLETED:
```
maxi297, line 406: nit: should we update the name of this method to remove `timed_out_jobs`? I think there is a comment where this is called that would need to be updated as well.

maxi297 (approved):
Confirming that this fixes the tests in source-salesforce. Thanks a lot @bazarnov ❤️ !

airbyte_cdk/sources/declarative/async_job/job_orchestrator.py
bazarnov: changed.

PR #434, opened 2025-03-19
fix: update decoder title & descriptions for improved tooltip

pnilan:
## What
This PR adds titles/descriptions to decoders where missing and updates existing descriptions as necessary to more closely align with the Builder UI.

Current tooltip provides insufficient and incorrect data as not all decoders title/description fields are properly populated:

Screenshot 2025-03-19 at 11 07 53 AM

## Summary by CodeRabbit

- **Documentation**
  - Enhanced guidance for selecting decoders based on response formats such as JSON, JSON Lines, Iterable, XML, ZIP File, Gzip, and CSV.

- **New Features**
  - Introduced new instructions for handling Gzip-compressed data and CSV-formatted data, including configuration options for inner data processing and formatting settings.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2193,7 +2193,7 @@ definitions:
         additionalProperties: true
   JsonDecoder:
     title: Json Decoder
-    description: Select 'JSON' as the decoder to parse the response as JSON.
+    description: Select 'JSON' if the response is formatted as a JSON object.
     type: object
     required:
       - type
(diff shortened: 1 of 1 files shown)


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2193,6 +2193,7 @@ definitions:
         additionalProperties: true
   JsonDecoder:
     title: Json Decoder
```
pnilan, line 2195: @lmossman Want to call this out here -- these titles that are rendered in the tooltip don't properly reflect the options available in the builder ("JSON" vs "Json Decoder"). I don't think it's worth actually aligning right now with this PR as I imagine this is covered by your upcoming work?

lmossman: Yeah I noticed that inconsistency too - I think it is something that you could just go ahead and correct in this PR if you'd like (e.g. rename `Json Decoder` to just `Json`, etc.) since that is probably what the change I would make would look like

lmossman (approved):
Changes LGTM

PR #443, opened 2025-03-25
PoC for emit file reference record

aldogonzalez8:


airbyte_cdk/sources/declarative/retrievers/file_uploader.py
```diff
@@ -40,5 +42,20 @@ def upload(self, record: Record) -> None:
         if self._content_extractor:
             raise NotImplementedError("TODO")
         else:
-            with open(str(Path(__file__).parent / record.data["file_name"]), "ab") as f:
+            files_directory = Path(get_files_directory())
+            # TODO:: we could either interpolate record data if some relative_path is provided or
+            #  use partition_field value in the slice {"partition_field": some_value_id} to create a path
+            file_relative_path = Path(record.stream_name) / record.data["file_name"]
```
maxi297, line 48: It is interesting. So it means that there is one other thing missing from the YAML declarative manifest and it is `filename_extractor` or something like that, right?

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
```diff
@@ -74,7 +74,8 @@ def read(self) -> Iterable[Record]:
                     )
                 )
                 if self._file_uploader:
-                    self._file_uploader.upload(record)
+                    file_reference = self._file_uploader.upload(record)
+                    record.file_reference = file_reference
```
maxi297, line 78: I think we should do this as part of the file_uploader. I don't see a world where the file uploader uploads a file but we don't want the file reference as part of the record

maxi297 (commented):
Adding comments following @aldogonzalez8 's changes

airbyte_cdk/sources/declarative/retrievers/file_uploader.py
aldogonzalez8: Yes, that would be very interesting to have, not a priority for PoC but we need to resolve in someway.

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
aldogonzalez8: Yeah, makes sense to me, will change this.

airbyte_cdk/sources/declarative/retrievers/file_uploader.py
aldogonzalez8: I trust your memory, but we could fallback to create sub-issues [here](https://github.com/airbytehq/airbyte-internal-issues/issues/12161). I just added [this one](https://github.com/airbytehq/airbyte-internal-issues/issues/12196).

airbyte_cdk/sources/declarative/retrievers/file_uploader.py
```diff
@@ -40,5 +44,25 @@ def upload(self, record: Record) -> None:
...
+            full_path.parent.mkdir(parents=True, exist_ok=True)
+
+            with open(str(full_path), "wb") as f:
                 f.write(response.content)
+            file_size_bytes = full_path.stat().st_size
+
+            logger.info("File uploaded successfully")
+            logger.info(f"File path: {full_path} ")
+            logger.info(f"File size: {file_size_bytes / 1024} KB")
+            logger.info(f"File download target: {download_target}")
```
aldogonzalez8, line 62: Im adding noise logging to debug in case we see issues moving the file.

airbyte_cdk/sources/declarative/retrievers/file_uploader.py
aldogonzalez8: https://github.com/airbytehq/airbyte-python-cdk/pull/453

aldogonzalez8: This has been added.

maxi297 (approved):
LGTM!

PR #446, opened 2025-03-26
Add file_based information in discovered catalog

maxi297:


aldogonzalez8:
ok component becomes the flag; that makes sense to me.

aldogonzalez8 (approved):
APPROVED

PR #448, opened 2025-03-27
feat(low code cdk): updated format_datetime macros to format timestamps

darynaishchenko:
### What
Needed for export stream migration to low code in source mixpanel https://github.com/airbytehq/airbyte/pull/55189
in python implementation of export stream cursor field `time` was transformed from timestamp to string in format "%Y-%m-%dT%H:%M:%SZ".
### How
As `format_datetime` macros works only with str and datetime objects , logic to parse timestamp to datetime if input value is instance of int was added.

## Summary by CodeRabbit

- **New Features**
	- Expanded date formatting capabilities to now handle integer timestamps along with string and datetime inputs.
- **Tests**
	- Enhanced test coverage with additional cases for converting Unix timestamps and epoch formats to formatted date strings.

### DIFF
--- a/airbyte_cdk/sources/declarative/interpolation/macros.py
+++ b/airbyte_cdk/sources/declarative/interpolation/macros.py
@@ -172,7 +172,7 @@ def format_datetime(
         return dt.strftime(format)
 
     if isinstance(dt, int):
-        dt_datetime = datetime.datetime.fromtimestamp(dt, tz=datetime.timezone.utc)
+        dt_datetime = DatetimeParser().parse(dt, input_format if input_format else "%s")
     else:
         dt_datetime = (
             datetime.datetime.strptime(dt, input_format) if input_format else str_to_datetime(dt)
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/interpolation/macros.py
```diff
@@ -170,9 +170,13 @@ def format_datetime(
     """
     if isinstance(dt, datetime.datetime):
         return dt.strftime(format)
-    dt_datetime = (
-        datetime.datetime.strptime(dt, input_format) if input_format else str_to_datetime(dt)
-    )
+
+    if isinstance(dt, int):
+        dt_datetime = datetime.datetime.fromtimestamp(dt, tz=datetime.timezone.utc)
```
maxi297, line 175: Should we distinguish between epoch seconds, milliseconds, microseconds and nanoseconds?

maxi297 (commented):
Just wondering if we could have multiple format for the integer here

airbyte_cdk/sources/declarative/interpolation/macros.py
darynaishchenko: input_format should be provided in this case, default "%s"

maxi297 (approved):
LGTM! Thanks for the added improvement

PR #449, opened 2025-03-27
feat(file-mode-api): move file uploader to record selector level.

aldogonzalez8:
Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/12226

Right now, the file uploader is at the schema level, but it needs to be at least at the end of the RecordSelector meaning. If we have client-side filtering, we only want to download the files if we know the record will get synced.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -207,19 +207,9 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
             # these legacy Python streams the way we do low-code streams to determine if they are concurrent compatible,
             # so we need to treat them as synchronous
 
-            file_uploader = None
-            if isinstance(declarative_stream, DeclarativeStream):
-                file_uploader = (
-                    self._constructor.create_component(
-                        model_type=FileUploader,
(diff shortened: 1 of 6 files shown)


maxi297 (approved):
The code looks good to me so I would say :shipit: 

However, should we be concerned about all the source tests failing with `Test failed for connector 'source-pokeapi' on step 'BuildConnectorImages'.`? I'm not sure how this would be related to this PR but it still feels off

aldogonzalez8:
> The code looks good to me so I would say :shipit:
> 
> However, should we be concerned about all the source tests failing with `Test failed for connector 'source-pokeapi' on step 'BuildConnectorImages'.`? I'm not sure how this would be related to this PR but it still feels off

My understanding is that because we are using the test pypy data classes source, once that change is merged, it should be able to build the image:

``` 
INFO: pip is looking at multiple versions of airbyte-cdk to determine which version is compatible with other requirements. This could take a while.
Stderr:
ERROR: Could not find a version that satisfies the requirement airbyte-protocol-models-dataclasses==0.14.1337.dev1742858109 (from airbyte-cdk) (from versions: 0.13.0, 0.13.1, 0.14.0, 0.14.1, 0.14.2)
ERROR: No matching distribution found for airbyte-protocol-models-dataclasses==0.14.1337.dev1742858109
```

PR #452, opened 2025-03-30
feat(Low-Code CDK Property Chunking): Allow fetching query properties and property chunking for low-code sources

brianjlai:
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/12145

## What

Adds support for the `QueryProperties` concept which allows for low-code connectors to inject either a static list of properties (source-linkedin-ads) or a set of properties retrieved from an API endpoint (`source-hubspot`) into the `HttpRequester.request_properties`

See the spec for more details about what this feature is and how it is used by connectors:
https://docs.google.com/document/d/1B_WcIEHTzctwHX_QjZgONXVF9g0i7BwpKzwrHMIA59s/edit?tab=t.0#heading=h.5dmjlrjauglv

## How

This PR has all the usual things associated with new components being added to the low-code framework. That includes updating `declarative_component_schema.yaml`, regenerating pydantic models, and updating `model_to_component_factory.py` to parse the models into runtime components.

The part I want to call out around model parsing is that we do something of a unique (and maybe a bit confusing) transformation of the `HttpRequester.request_parameters.query_properties` model definition by moving it under the `SimpleRetriever.additional_query_properties` for runtime. The optimal developer experience is to define it under `request_parameters` because that is ultimately where it gets injected. However, the SimpleRetriever needs to have context into the `QueryProperties` component because it orchestrates the requests being made and merging records back together.

At runtime, PR can be broken down into the two pieces of functionality mentioned above

### Defining or retrieving the complete set of request properties to inject:

The `QueryProperties` runtime class acts as the orchestrator behind getting the list of properties that need to be queried for either through the statically defined list or the `PropertiesFromEndpoint` class. It then leverages `PropertyChunking` if defined, otherwise it returns a single set of properties. This way, all `QueryProperty` output is processed in the same way by the `SimpleRetriever` regardless of whether property chunking is defined.

Some connectors like linkedin ads have a unique identifier based on property fields that must be requested and therefore they must be included in every grouping of properties. The `always_include_properties` field allows us to define fields we know must be added. And this allows us to avoid using interpolation to build a complex query field such as: `fields: "{{ ','.join(['required', stream_partition.extra_fields['query_properties']) }}"`

### Performing chunking into smaller sets of properties for APIs with restrictions around requests:

The `PropertyChunking` runtime class is responsible for dictating when to break the complete set of components into smaller groups either by the character count or by the number of properties per group. It also supplies the `record_merge_strategy` which would only be needed when there are multiple groups

### Loading the `query_properties` into the the outbound API request

This is probably one of the more awkward parts of the code. We need some mechanism of passing the current chunk of `query_properties` to inject from the SimpleRetriever.read_records` to the `InterpolatedRequestOptionsProvider`. The most convenient way to avoid changing the interface is to use `extra_properties` which won't be persisted back as state.

Even though jinja interpolation _could_ work and might be more consistent so that every request_param is a string, it feels backwards to convert things back into interpolated strings when we can just operate on the `StreamSlice` directly

## Summary by CodeRabbit

- **New Features**
  - Introduced advanced record merging strategies based on configurable keys.
  - Enhanced property extraction by supporting both static definitions and dynamic retrieval from external APIs.
  - Implemented property chunking to efficiently handle API limits during data retrieval.
  - Improved API request handling with additional query property support for flexible data acquisition.

- **Tests**
  - Expanded test coverage to validate the new merging, property extraction, and chunking functionality.
  - Added tests for handling additional query properties and ensuring proper functionality of the `SimpleRetriever` class.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2765,6 +2882,53 @@ def create_simple_retriever(
...
+                    and request_parameter.get("type") == "QueryProperties"
+                ):
+                    query_properties_key = key
+                    query_properties_definitions.append(request_parameter)
+                elif not isinstance(request_parameter, str):
+                    raise ValueError(
+                        f"Each element of request_parameters should be of type str or QueryProperties, but received {request_parameter.get('type')}"
+                    )
+
+            if len(query_properties_definitions) > 1:
```
brianjlai, line 2906: For simplicity its easier to just only ever worry about a single `QueryProperties` component defined per-requester. I don't currently see a use case. Unfortunately, we also can't rely on a auto-validations because as far as i know json schema can't enforce this granular a check of a specify elements type multiple times.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2765,6 +2882,53 @@ def create_simple_retriever(
...
+            query_properties_definitions = []
+            for key, request_parameter in model.requester.request_parameters.items():
+                if (
+                    isinstance(request_parameter, Mapping)
+                    and request_parameter.get("type") == "QueryProperties"
+                ):
+                    query_properties_key = key
+                    query_properties_definitions.append(request_parameter)
+                elif not isinstance(request_parameter, str):
+                    raise ValueError(
```
brianjlai, line 2902: I couldn't get json schema to pydantic model conversions to enforce `str, `QueryProperties`. It always would get rendered as `Any` regardless of what combo I made once we included both a `#ref` and concrete `type`

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    property_list: Optional[Union[List[str], PropertiesFromEndpoint]]
+    always_include_properties: Optional[List[str]]
+    property_chunking: Optional[PropertyChunking]
+    config: Config
+    parameters: InitVar[Mapping[str, Any]]
+
+    def get_request_property_chunks(
+        self, stream_slice: Optional[StreamSlice] = None
```
maxi297, line 29: In which case does the list of properties depends on the stream_slice?

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
```diff
@@ -0,0 +1,51 @@
...
+            fields = self.property_list if self.property_list else []
+
+        if self.property_chunking:
+            yield from self.property_chunking.get_request_property_chunks(
+                property_fields=fields, always_include_properties=self.always_include_properties
+            )
+        else:
+            yield from [list(fields)]
+
+    def has_multiple_chunks(self, stream_slice: Optional[StreamSlice]) -> bool:
```
maxi297, line 44: nit: It feels like we could let that to the consumer of the query properties (in our case, SimpleRetriever) we don't need to maintain this new method. Do we know in which other case we would need this method? I'm mostly worried about maintaining multiple methods that might generate multiple same HTTP requests multiple times

maxi297 (commented):
I just need a bit more information and I'll review this again. I can definitely say is that this YAML interface is clean and I really like it

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
brianjlai: I think that's fair and it does give the impression this is a "cheap" method, but in reality makes an API call. The main reason i ended up adding it was that we needed to introspect the iterable to detect if there were at least two elements. And we've been burned in the past with `tee()` not deallocating slices properly.

But I think we can move this method up into the retriever and if we combine with caching `get_request_property_chunks()` that should make things less require fewer outbound API calls. Thanks for the suggestion

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
brianjlai: This is for consistency so that the `SimpleRetriever`/`HttpRequester`, etc that is defined to fetch properties from the API endpoint have access to the stream slice object of the current partition. Hubspot specifically doesn't need access to the slice. 

If we don't supply this, then if a future connector were to attempt to use jinja interpolation like we do for query params, this would get resolved to None. So even though we don't current have a use case, I would advocate that we keep this since it doesn't hurt for it to exist

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
```diff
@@ -0,0 +1,40 @@
...
+    parameters: InitVar[Mapping[str, Any]]
+
+    def __post_init__(self, parameters: Mapping[str, Any]) -> None:
+        self._property_field_path = [
+            InterpolatedString(string=property_field, parameters=parameters)
+            for property_field in self.property_field_path
+        ]
+
+    def get_properties_from_endpoint(self, stream_slice: Optional[StreamSlice]) -> Iterable[str]:
+        response_properties = self.retriever.read_records(
```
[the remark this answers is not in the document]
brianjlai: It shouldn't be every time. We get the set of properties once at the beginning when we process a slice. But it would be the case that we perform this fetching for every slice. It still makes sense to look into caching this request.

This might be ignorance on my part, but how do we enable caching? Is it through the model `HttpRequester.use_cache` field? I see how it enables the sqllite cache on the `HttpClient`, but I as far as I can tell the default is off, and we have manifests that enabled it.

Are you saying that we should be marking `use_cache=True` under the hood for both dynamic schema discovery and for property retrieval? Or am I completely off base and there's something I can do to turn it on annotations or new Python code

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: > Should we improve the error message to describe what the alternative should be?

I'm not sure we have an alternative? But I also don't currently see a use case for needing to pull properties from two endpoints, load them into two different query parameters, and needing to perform property chunking on both. And if that's a thing, I'm okay waiting until we see this to address it.

> Also, this might be far fetched but isn't it a case we implicitly support if PropertiesFromEndpoint returns only one property? Is this a problem?

This should already be supported if `PropertiesFromEndpoint` returns only has one property. This restriction to `len(query_properties_definitions) > 1` refers to specifically if there is more than 1 definition of a field to the set of properties to inject

So this would invalid:
```
request_parameters:
  request_field_1:
    type: QueryProperties
    ...
  request_field_2:
    type: QueryProperties
```

But if there was just `request_field_1` that returned only one property in the set then that would work fine. Unless I just misunderstood your question

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
```diff
@@ -0,0 +1,40 @@
...
+            records_schema={}, stream_slice=stream_slice
+        )
+        for property_obj in response_properties:
+            path = [
+                node.eval(self.config) if not isinstance(node, str) else node
+                for node in self._property_field_path
+            ]
+
+            yield dpath.get(property_obj, path)  # type: ignore # extracted will be a MutableMapping, given input data structure
```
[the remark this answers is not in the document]
brianjlai: fixed

unit_tests/sources/declarative/test_manifest_declarative_source.py
```diff
@@ -627,7 +627,6 @@ def test_source_missing_checker_fails_validation(self):
                     },
                 }
             ],
-            "check": {"type": "CheckStream"},
```
brianjlai, line 630: I was getting test failures here and my suspicion is that because we no longer made `stream_names` a required field since it also supports dynamic streams check configs. This was in https://github.com/airbytehq/airbyte-python-cdk/pull/450

So then this component is actually valid and does not throw the Validation error anymore. Getting rid of this component entirely causes the test to correctly fail.

What I don't understand is why the tests didn't fail on the above PR. Nonetheless getting rid of this should fix my build

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1044,6 +1044,18 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
+  EmitPartialRecordMergeStrategy:
+    title: Emit Partial Record
+    description: Record merge strategy where in the case where multiple requests are needed to retrieve all properties, properties are not consolidated back into a single record and are instead emitted as separate groups of properties. This strategy should only be used when records do not have a unique identifier like a primary key.
```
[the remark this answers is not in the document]
maxi297: image

Source: https://github.com/airbytehq/airbyte/pull/13596 which was related this [this oncall issue](https://github.com/airbytehq/oncall/issues/259)

Checking more comments, it seems like the properties history streams didn't have pk before:
* https://github.com/airbytehq/oncall/issues/259#issuecomment-1149506504
* [PropertyHistory was as full fledge stream](https://github.com/airbytehq/airbyte/blob/0817fae7b899048a8792f253c678228fb4c6422a/airbyte-integrations/connectors/source-hubspot/source_hubspot/source.py#L119) which wasn't delineated into more precise propertyhistory like `CompaniesPropertyHistory` where [no PKs were defined](https://github.com/airbytehq/airbyte/blob/0817fae7b899048a8792f253c678228fb4c6422a/airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py#L1319-L1329)

One interesting thing I see is that the timestamp is on the primary key (example for [CompaniesPropertyHistory](https://github.com/airbytehq/airbyte/blob/48b96cb6731209db452f83fdbe030215be258ddb/airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py#L2212)). The same applies to [ContactsPropertyHistory](https://github.com/airbytehq/airbyte/blob/48b96cb6731209db452f83fdbe030215be258ddb/airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py#L2150-L2152) even though it's not PropertyHistoryV3. 

Honestly, it sounds to me like it only works because of the timestamp in the PK and therefore, we could have the merge strategy for this as well. WDYT?

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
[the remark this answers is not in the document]
maxi297: `HttpRequester.use_cache` should pass the information to the `HttpClient`, yes. That being said, both `HttpClient` will need the same name because [the HttpClient name is used for the cache name](https://github.com/airbytehq/airbyte-python-cdk/blob/caa1e7d938803b37d6b851d2b811a3176203a38c/airbyte_cdk/sources/streams/http/http_client.py#L124-L130).

Note that we also dynamically set caching on parent streams [here](https://github.com/airbytehq/airbyte-python-cdk/blob/caa1e7d938803b37d6b851d2b811a3176203a38c/airbyte_cdk/sources/declarative/manifest_declarative_source.py#L177-L219) which I would prefer this to be logic in the model_to_component_factory but :shrug:

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
maxi297: I'm ok with that but I would document the reasoning as a docstring for this method so that it's less surprising for other users. Right now, I think my head can wrap the idea that different slices would request for different properties but at the same time, there are weird APIs out there

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
maxi297: > we needed to introspect the iterable to detect if there were at least two elements

We could also increment a variable to know how many chunks were generated and take action on this. I don't have a preference though

unit_tests/sources/declarative/test_manifest_declarative_source.py
maxi297: What were the test failures? Does that means that https://github.com/airbytehq/airbyte-python-cdk/pull/450 was a breaking change?

brianjlai: basically this test was no longer raising a `ValidationError` despite expecting it. 
```
Failed: DID NOT RAISE <class 'jsonschema.exceptions.ValidationError'>
```

I wouldn't consider it to be a breaking change because we were relaxing the restrictions on what throws an error. So no existing manifests should break. Only tests that were asserting that the manifest validations raise errors like this test does. Make sense?

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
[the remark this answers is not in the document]
brianjlai: > That being said, both HttpClient will need the same name because [the HttpClient name is used for the cache name](https://github.com/airbytehq/airbyte-python-cdk/blob/caa1e7d938803b37d6b851d2b811a3176203a38c/airbyte_cdk/sources/streams/http/http_client.py#L124-L130).

Welllll, we just call it `""` on the dynamic schema loader... , good catch though, if we want to share the caching then we can just call it `dynamic_properties` since it's effectively what it is. I could be mistaken, but I think the dynamic loader is actually only invoked during discover not read, so having them share the same cache probably isn't a massive performance improvement in comparison to making sure we always cache the `PropertiesFromEndpoint`

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
brianjlai: yep i'll add it to the docstring

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
brianjlai: I thought about that too, but one problem with that is that is we want to know whether we need to start accumulating records in the merge dictionary on the first chunk, so the incrementing variable only allows us to know on the second loop. So that was why the alternative was a `tee()`. 

Alternatively, I don't love it, but I would also expect the property list to generally not be that big, so we could convert it to a list and just get the length. but I still lean towards just using this method twice since it should be cached

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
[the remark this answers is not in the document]
brianjlai: mmk yep you've convinced me, I can take out this strategy, no need to support it if it doesn't feel like we need it or hubspot can be modified to not use this strategy at all

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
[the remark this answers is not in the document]
maxi297: > I could be mistaken, but I think the dynamic loader is actually only invoked during discover not read 

I also don't know but I fear some calls [here](https://github.com/airbytehq/airbyte-python-cdk/blob/caa1e7d938803b37d6b851d2b811a3176203a38c/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L273) could trigger the loading of the json schema. I also fear some custom code should have access to `get_json_schema` and call it multiple times but I have less reasons to believe the latter case

unit_tests/sources/declarative/test_manifest_declarative_source.py
maxi297: Yes, that makes sense, thanks for the precision!

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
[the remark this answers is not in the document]
brianjlai: yeah that's true. either way, i have the cache and the same name setup for requesters/clients when we build both components in the model factory so we should be all set here

airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
brianjlai: i think that makes sense, I tweaked the design a bit to be agnostic to how many chunks. And also made a few adjustments so that even if we don't get a merge key, we still ensure that we emit the record

maxi297 (approved):
LGTM! Thanks for the many improvements ❤️

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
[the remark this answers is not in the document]
maxi297: I have been playing with migrating the Connector Builder to the Concurrent CDK (because I ended up finding it is still using the declarative state) and I can confirm that this is an issue as I'm getting the following issue once I use `ConcurrentDeclarativeSource` instead of `ManifestDeclarativeSource`: 

```
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/connector_builder/test_reader/reader.py", line 389, in _read_stream
    yield from AirbyteEntrypoint(source).read(
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/entrypoint.py", line 244, in read
    for message in self.source.read(self.logger, config, catalog, state):
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/sources/declarative/concurrent_declarative_source.py", line 138, in read
    concurrent_streams, _ = self._group_streams(config=config)
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/sources/declarative/concurrent_declarative_source.py", line 333, in _group_streams
    declarative_stream.get_json_schema(),
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/sources/declarative/declarative_stream.py", line 170, in get_json_schema
    return self._schema_loader.get_json_schema()
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/sources/declarative/schema/default_schema_loader.py", line 39, in get_json_schema
  File "/Users/maxime/devel/code/airbyte-python-cdk/airbyte_cdk/sources/declarative/schema/json_file_schema_loader.py", line 58, in get_json_schema
    raw_json_file = pkgutil.get_data(resource, schema_path)
  File "/Users/maxime/.pyenv/versions/3.10.14/lib/python3.10/pkgutil.py", line 621, in get_data
    spec = importlib.util.find_spec(package)
  File "/Users/maxime/.pyenv/versions/3.10.14/lib/python3.10/importlib/util.py", line 103, in find_spec
    return _find_spec(fullname, parent_path)
  File "<frozen importlib._bootstrap>", line 945, in _find_spec
  File "/Users/maxime/devel/code/airbyte-python-cdk/.venv/lib/python3.10/site-packages/_pytest/assertion/rewrite.py", line 110, in find_spec
    if self._early_rewrite_bailout(name, state):
  File "/Users/maxime/devel/code/airbyte-python-cdk/.venv/lib/python3.10/site-packages/_pytest/assertion/rewrite.py", line 211, in _early_rewrite_bailout
    path = PurePath(*parts).with_suffix(".py")
  File "/Users/maxime/.pyenv/versions/3.10.14/lib/python3.10/pathlib.py", line 782, in with_suffix
    raise ValueError("%r has an empty name" % (self,))
ValueError: PurePosixPath('.') has an empty name
```

This indicates that in the previous implementation of the test, the schema loader was not configured properly but the code path that would evaluate json schema was never exercised.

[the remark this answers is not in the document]
brianjlai: wait so what is the issue? that we need to cache or there's something functionally wrong with this part of the code?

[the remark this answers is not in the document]
maxi297: I'm not sure this is an issue but the difference is that `get_json_schema` wasn't called during read for the tests I was running while when migrating to the ConcurentDeclarativeSource, it is called in `_group_streams` at the initialization of the source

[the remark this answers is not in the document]
brianjlai: ah gotcha so more that it validates that we do in fact use this during a read and could benefit from caching. i use the same stream name on both schema loader and properties retriever, so caching should hit on the same entry for both

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
bazarnov: @brianjlai This is the problem: 

nit: not `$ref"`, but `"$ref"`

here: https://github.com/airbytehq/airbyte-python-cdk/blame/2f29eff3f5d74862e1667f6274a0f23d1b3a6c0c/airbyte_cdk/sources/declarative/declarative_component_schema.yaml#L2036

PR #453, opened 2025-03-31
feat(file-mode-api: add filename extractor component

aldogonzalez8:
Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/12196

### DIFF
--- a/airbyte_cdk/sources/declarative/retrievers/file_uploader.py
+++ b/airbyte_cdk/sources/declarative/retrievers/file_uploader.py
@@ -24,7 +24,6 @@ class FileUploader:
         self._content_extractor = content_extractor
 
     def upload(self, record: Record) -> None:
-        # TODO validate record shape - is the transformation applied at this point?
         mocked_response = SafeResponse()
         mocked_response.content = json.dumps(record.data).encode("utf-8")
         download_target = list(self._download_target_extractor.extract_records(mocked_response))[0]
--- a/airbyte_cdk/sources/declarative/retrievers/file_uploader.py
+++ b/airbyte_cdk/sources/declarative/retrievers/file_uploader.py
@@ -1,7 +1,16 @@
-import json
-from pathlib import Path
-from typing import Optional
+#
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+#
 
+import json
+import logging
+from dataclasses import InitVar, dataclass, field
+from pathlib import Path
+from typing import Optional, Mapping, Union, Any
+
+from airbyte_cdk.sources.declarative.interpolation.interpolated_string import (
+    InterpolatedString,
+)
 from airbyte_cdk.models import AirbyteRecordMessageFileReference
 from airbyte_cdk.sources.declarative.extractors.record_extractor import RecordExtractor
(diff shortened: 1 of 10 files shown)


airbyte_cdk/sources/declarative/retrievers/file_uploader.py
```diff
@@ -1,50 +1,68 @@
...
             files_directory = Path(get_files_directory())
-            # TODO:: we could either interpolate record data if some relative_path is provided or
-            #  use partition_field value in the slice {"partition_field": some_value_id} to create a path
-            file_relative_path = Path(record.stream_name) / record.data["file_name"]
+
+            relative_path = self._filename_extractor.eval(self.config, record=record)
+            relative_path = relative_path.lstrip("/")
+            file_relative_path = Path(relative_path)
 
             full_path = files_directory / file_relative_path
```
maxi297, line 67: Should we have the stream name somewhere in there? It feels like multiple streams could have a file with the same name.

Even more than this, should we have a unique ID per file? It feels like there could even be two files in the same stream with the same name...

maxi297: Ok so what we are saying is that it is the developer's responsibility to make sure there are no clash. Could we remove this concern from the developer's and do it ourselves? 

Regarding timing: I'm not 100% sure we need this right now and maybe we can make `filename_extractor` optional in the future when we find a solution this this. On the top of my head, I can only see one way and it is when the stream declares a PK which seems to be common when I checked for Confluence, Jira and Salesforce so maybe this is viable in the future

maxi297 (approved):
Accepting under the premise that we are fine that the connector developer ensure no file collisions for now

airbyte_cdk/sources/declarative/retrievers/file_uploader.py
aldogonzalez8: > so what we are saying is that it is the developer's responsibility to make sure there are no clash. Could we remove this concern from the developer's and do it ourselves?

No, I didn't make myself clear. I'm sorry about that. To reduce the risk of collisions, I will add the stream name + unique ID on the backend (CDK).

maxi297: What is the logic for the unique ID? Autogenerated UUID?

aldogonzalez8: I think we can:

- Add ourselves stream name to the path reducing collision risk
- Make filename_extractor optional so the developer can include a unique ID. There is a risk that he could mess up, but we can add some documentation to the component. 
- Use Autogenerated UUID if filename_extractor is not present.

aldogonzalez8:
> Accepting under the premise that we are fine that the connector developer ensure no file collisions for now

@maxi297  This is still true with slight modifications that we can refine in the future:

- We added the stream name to the path, reducing collision risk.
- Make filename_extractor optional so the developer can include a unique ID. There is a risk that dev could mess up the interpolation, but we added documentation to the component.
- Use Autogenerated UUID if filename_extractor is not present.

PR #454, opened 2025-04-01
feat(low code cdk): add path to filter interpolation in AddFields condition

darynaishchenko:
### What
Needed for mixpanel dynamic schemas https://github.com/airbytehq/airbyte/pull/55189
When stream has static fields in schema, they should be added only when dynamic schema doesn't have these fields in it.
### How
To check if field already exists in schema a condition in AddFields transformation can be used, but now it doesn't have path value in interpolation arguments.
Added  `path=parsed_field.path` to `_filter_interpolator.eval`.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced field processing by refining the transformation logic to incorporate additional context during condition checks. This update improves how fields are added based on record content, ensuring more consistent and reliable outcomes during data processing. These enhancements contribute to a smoother and more predictable overall experience when working with records.


maxi297 (approved):
:shipit:

PR #457, opened 2025-04-03
feat(PoC): adjust file-based and file uploader component to latest protocol changes.

aldogonzalez8:
## What

This PR updates the file-based and file uploader components in the Airbyte Python CDK to align with the [file transfer record protocol](https://github.com/airbytehq/airbyte-protocol/pull/119/files) changes introduced in the platform. It introduces schema refinements, file path handling improvements, and new test cases.

Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/12364

## How
- Updated file transfer logic to reflect new protocol structure.
- Adjusted the uploader to handle file references at the record selector level.
- Performed refactoring and linting cleanup across modules.

## Review guide
File based changes:
1. `airbyte_cdk/models/airbyte_protocol.py`: remove hacked protocol
2. `airbyte_cdk/models/file_transfer_record_message.py`: remove hacked protocol
3. `airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py`: remove hacked protocol
4. `airbyte_cdk/sources/file_based/file_based_stream_reader.py`: change method verb and return type to AirbyteRecordMessageFileReference, also make _get_file_transfer_paths support method return a dict with path fields.
5. `airbyte_cdk/sources/file_based/file_record_data.py`: helper model for record (metadata) of files.
6. `airbyte_cdk/sources/file_based/file_types/file_transfer.py`: update to return record and file reference data.
7. `airbyte_cdk/sources/file_based/schema_helpers.py`: schema of records (metadata) for file-based connectors.
8. `airbyte_cdk/sources/file_based/stream/concurrent/adapters.py`: pass file_reference
9.  `airbyte_cdk/sources/file_based/stream/default_file_based_stream.py`: introduce changes to default file based stream to handle new file reference and records data besides fixed schema.
10. `airbyte_cdk/sources/file_based/stream/permissions_file_based_stream.py`: update call to `stream_data_to_airbyte_message`
11. `airbyte_cdk/sources/types.py`: remove old `is_file_transfer_message` flag
12. `airbyte_cdk/sources/utils/record_helper.py`: remove handling of `is_file_transfer_message` flag
13. `airbyte_cdk/test/mock_http/response_builder.py`: add helper method to get binary data from file for testing

File-api changes:
1. `airbyte_cdk/sources/declarative/retrievers/file_uploader.py`: update latest protocol fields names.

## User Impact
Developers using the file-based CDK and file uploader in declarative functionality will benefit from file_reference protocol support. 

- [ ] YES 💚
- [x] NO ❌


airbyte_cdk/sources/file_based/file_record_data.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+#
+
+from datetime import datetime
+from typing import Optional
+
+from pydantic.v1 import BaseModel
+
+
+class FileRecordData(BaseModel):
```
maxi297, line 11: Maybe future us problem: Can/Should we allow for additional properties so that sources can add whatever they want? If we do, I think it'll have impact on airbyte_cdk/sources/file_based/schema_helpers.py which seems a bit more challenging as it can't be modified based on the source as it is right now

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -97,14 +97,7 @@ def _filter_schema_invalid_properties(
         self, configured_catalog_json_schema: Dict[str, Any]
     ) -> Dict[str, Any]:
         if self.use_file_transfer:
```
maxi297, line 99: nit: Does it make sense to simplify this logic and always call _filter_schema_invalid_properties? The platform probably filter them out anyway right? 

My concern is that we wouldn't have to maintain and do changes here if the logic wasn't depending on `use_file_transfer`

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -166,26 +151,21 @@ def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Airbyte
             raise MissingSchemaError(FileBasedSourceError.MISSING_SCHEMA, stream=self.name)
         # The stream only supports a single file type, so we can use the same parser for all files
         parser = self.get_parser()
+        file_transfer = FileTransfer()
```
maxi297, line 154: Why has it been moved up there?

nit: If this would be as a field instead of a method variable, we would avoid the allocation every time it is called

airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -162,25 +169,41 @@ def get_file(
...
+                   - source_file_relative_path (str): The relative path to the referenced file in source.
         """
         ...
 
-    def _get_file_transfer_paths(self, file: RemoteFile, local_directory: str) -> List[str]:
+    def _get_file_transfer_paths(
+        self,
+        file: RemoteFile,
+        local_directory: str,
+        parse_file_path_from_uri: Optional[Callable[[str], str]] = None,
```
maxi297, line 183: A couple of things regarding this parameter:
* I don't see where this is used. In which case does it make sense?
* Should we add a test to document an actual use case for this parameter? 
* Once I understand a bit more how it is used, I might propose things like having this as an abstract method of AbstractFileBasedStreamReader

maxi297 (commented):
I asked a couple of questions to make sure I understand everything but I don't think anything will be blocking.

Also in theory, this is a breaking change but it feels like we will update all the connectors that populates `AirbyteFileTransferRecordMessage` in `record` or call `AbstractFileBasedStreamReader.get_file`, right? Are there other breaking changes we should be worried about?

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
aldogonzalez8: I think you are right on this; we can remove this override that was introduced [here](https://github.com/airbytehq/airbyte-python-cdk/commit/522f9c88dc65b0da249cad805cc5adff1d6375ae#diff-8556db57d11c88793c91887a919c48e97bae0262889abd405713c0a380ab4bb9R78).

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
aldogonzalez8: > Why has it been moved up there?

We will create in every call to `read_records_from_slice` instead of every iteration of the loop `for file in stream_slice["files"]`

> If this would be as a field instead of a method variable, we would avoid the allocation every time it is called

I think this is better. Let me do that.

airbyte_cdk/sources/file_based/file_based_stream_reader.py
aldogonzalez8: In reflection, this approach (mine) was wrong from the initial implementation; we used URI as the relative path, which causes the expectation URI to be that instead of the resource identifier. Then, when a connector is passing a real URI, it messes with the logic. 

I will simplify this to the connector to resolve the relative path so Stream Reader doesn't need any helper method to try to translate URI => relative path but just resolve the directory path for the staging drive.

airbyte_cdk/sources/file_based/file_record_data.py
aldogonzalez8: > Can/Should we allow for additional properties so that sources can add whatever they want?

For now, this is not a concern[ per the discussion in our last meetings](https://docs.google.com/document/d/1xIVlidNME1Jh9zyXaVi166TPyRNhQu2hBgqCT7eaOYs/edit?tab=t.0#heading=h.1qnv42k86dbu) with the team.

Users are getting this metadata "for free" (nobody requested it), but if we need to allow it in the future, we could provide an interface in stream reader similar to what we do to [model permissions schema](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/file_based/stream/permissions_file_based_stream.py#L94).

I agree this would mean maintenance, but we don't have a current use case, so my belief is that we are ok as it is.

cc @girarda @tryangul

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
aldogonzalez8: Ok, already did.

airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
aldogonzalez8: Ok, removed.

aldogonzalez8:
> I asked a couple of questions to make sure I understand everything but I don't think anything will be blocking.
> 
> Also in theory, this is a breaking change but it feels like we will update all the connectors that populates `AirbyteFileTransferRecordMessage` in `record` or call `AbstractFileBasedStreamReader.get_file`, right? Are there other breaking changes we should be worried about?

There are no other things to worry about, as previously, we had a schema, but we ignored such schema (which actually is/was messy) as the only point of interest was to move the file. From now on, we will move data along the file.

We will need to refresh the schema for these.

Old (messy,, but we didn't care):

image

New (need to update pre-dev as I'm adding source-uri but same idea):

image

airbyte_cdk/sources/file_based/file_record_data.py
girarda: If allowing additional fields is difficult, I'd suggest adding an "additional_properties" field that sources can populate however they want

airbyte_cdk/sources/file_based/file_record_data.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    A record in a file-based stream.
+    """
+
+    folder: str
+    filename: str
+    bytes: int
+    source_uri: str
+    id: Optional[str] = None
+    updated_at: Optional[str] = None
```
girarda, line 21: should we also have a `created_at`?

aldogonzalez8: Is now there

airbyte_cdk/sources/file_based/file_record_data.py
aldogonzalez8: Ok, [issue created](https://github.com/airbytehq/airbyte-internal-issues/issues/12418). I also added `created_at` to the schema, I feel this should be enough for this iteration.

airbyte_cdk/test/standard_tests/__init__.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
+'''FAST Airbyte Standard Tests
```
maxi297, line 2: Why do we have CATs tests related changes here?

maxi297 (approved):
I'm good with this change assuming the CAT changes were out of scope

airbyte_cdk/test/standard_tests/__init__.py
aldogonzalez8: I'm a terrible person and I'm taking credit for @aaronsteers work :( Sorry.

JK, it is because it is a stacked PR where both branches run behind main. Let me fix that.

maxi297: It's fine! not a blocker for going forward with this

PR #460, opened 2025-04-07
poc: connector builder using concurrent cdk

maxi297:


airbyte_cdk/sources/concurrent_source/concurrent_source.py
```diff
@@ -44,6 +44,7 @@ def create(
         slice_logger: SliceLogger,
         message_repository: MessageRepository,
         timeout_seconds: int = DEFAULT_TIMEOUT_SECONDS,
+        queue: Optional[Queue[QueueItem]] = None
```
maxi297, line 47: Since the MessageRepository also needs to have access to the queue, we need to have the queue passed here instead of being created

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -360,16 +423,19 @@ def _group_streams(
                     and incremental_sync_component_definition.get("type", "")
                     == DatetimeBasedCursorModel.__name__
                     and hasattr(declarative_stream.retriever, "stream_slicer")
-                    and isinstance(
-                        declarative_stream.retriever.stream_slicer, PerPartitionWithGlobalCursor
+                    and (
```
maxi297, line 426: All the changes in this clause are caused by the `TestReadSlicerDecorator` that wraps the "real" `partition_router`. It affects how we call `create_concurrent_cursor_from_perpartition_cursor` (we need the inner partition router when creating it) and how we create the `StreamSlicerPartitionGenerator` as we need to re-wrap the stream_slicer in the `TestReadSlicerDecorator`

In a world where we don't have declarative cursors, the cursor would already be wrapper and we wouldn't have to do shenanigans here

airbyte_cdk/sources/declarative/interpolation/macros.py
```diff
@@ -177,6 +177,7 @@ def format_datetime(
         dt_datetime = (
             datetime.datetime.strptime(dt, input_format) if input_format else str_to_datetime(dt)
         )
+    dt_datetime = dt_datetime.replace(tzinfo=pytz.utc)
```
maxi297, line 180: Outside of the scope of this PR. See https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1743795253674379 for more details

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -623,6 +619,7 @@ def _fetch_next_page(
                 stream_slice=stream_slice,
                 next_page_token=next_page_token,
             ),
+            # FIXME remove this implementation and have the log_formatter depend on the fact that the logger is debug or not
```
maxi297, line 622: I would like to remove the SimpleRetriever decorator eventually. We just need to have the log_formatter be optional be an optional field of the SimpleRetriever and we should be fine.

In theory, this seems good because it would allow for test reads configuration to apply to any retriever, not just the SimpleRetriever

airbyte_cdk/sources/declarative/schema/default_schema_loader.py
```diff
@@ -37,7 +37,7 @@ def get_json_schema(self) -> Mapping[str, Any]:
 
         try:
             return self.default_loader.get_json_schema()
-        except OSError:
+        except (OSError, ValueError):
```
maxi297, line 40: Outside of the scope of this PR. See https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1743795253674379 for more details

airbyte_cdk/test/entrypoint_wrapper.py
```diff
@@ -157,6 +157,8 @@ def _run_command(
     stream_handler.setFormatter(AirbyteLogFormatter())
     parent_logger = logging.getLogger("")
     parent_logger.addHandler(stream_handler)
+    if "--debug" not in args:
```
maxi297, line 160: Not needed for tests to pass. Just enabling more logs in the tests for debugging purposes

unit_tests/connector_builder/test_connector_builder_handler.py
```diff
@@ -140,6 +142,19 @@
             "type": "DeclarativeStream",
             "$parameters": _stream_options,
             "retriever": "#/definitions/retriever",
+            "incremental_sync": {
```
maxi297, line 145: Added to test that the state is properly emitted. I also had to add `updated_at` fields in the records to avoid things from breaking

unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -2902,7 +2902,7 @@ def test_use_request_options_provider_for_datetime_based_cursor():
         parameters={},
     )
 
-    connector_builder_factory = ModelToComponentFactory(emit_connector_builder_messages=True)
```
maxi297, line 2905: I could have updated the assertion to check for a `assert isinstance(retriever.stream_slicer, TestReadStreamSlicerDecorated)` or disable `emit_connector_builder_messages`. I arbitrarily chose the latter

unit_tests/sources/streams/concurrent/test_concurrent_read_processor.py
```diff
@@ -171,90 +168,19 @@ def test_handle_partition(self):
...
 
         self._thread_pool_manager.submit.assert_called_with(
-            self._partition_reader.process_partition, self._a_closed_partition
+            self._partition_reader.process_partition, self._a_closed_partition, self._another_stream.cursor
         )
         assert (
             self._a_closed_partition in handler._streams_to_running_partitions[_ANOTHER_STREAM_NAME]
         )
 
-    def test_handle_partition_emits_log_message_if_it_should_be_logged(self):
```
maxi297, line 188: slice logging has been moved to PartitionReader

unit_tests/sources/streams/concurrent/test_concurrent_read_processor.py
```diff
@@ -171,90 +168,19 @@ def test_handle_partition(self):
...
-
-        self._thread_pool_manager.submit.assert_called_with(
-            self._partition_reader.process_partition, self._an_open_partition
-        )
-        self._message_repository.emit_message.assert_called_with(self._log_message)
-
-        assert self._an_open_partition in handler._streams_to_running_partitions[_STREAM_NAME]
-
-    @freezegun.freeze_time("2020-01-01T00:00:00")
-    def test_handle_on_partition_complete_sentinel_with_messages_from_repository(self):
```
maxi297, line 214: I think the part of this test that checks if the message repository is still consumed might be still relevant but I assume that the implementation is always `ConcurrentMessageRepository` now and `MessageRepository.consume_queue` might not be useful in a concurrent world...

What is sure is that `self._stream.cursor.close_partition.assert_called_once()` is not useful anymore because it has been moves to the concurrent read processor

unit_tests/sources/streams/concurrent/test_concurrent_read_processor.py
```diff
@@ -302,55 +227,6 @@ def test_handle_on_partition_complete_sentinel_yields_status_message_if_the_stre
             )
         ]
         assert messages == expected_messages
-        self._another_stream.cursor.close_partition.assert_called_once()
-
-    @freezegun.freeze_time("2020-01-01T00:00:00")
-    def test_given_exception_on_partition_complete_sentinel_then_yield_error_trace_message_and_stream_is_incomplete(
```
maxi297, line 308: This was used in case `close_partition` was failing. We don't call `close_partition` anymore hence this test is useless now

maxi297 (commented):
Commenting on the PoC...

PR #462, opened 2025-04-07
fix: no longer require token expiration from the OAuth resource server

dbgold17:
fixes https://github.com/airbytehq/airbyte/issues/54198

question: what sort of larger integration / end-to-end testing should be done on this? Should I run a sync for instance?

The proposal is to create a default expiration (currently 1 hour) that we use if we do not receive a token expiration back from the resource server when requesting / refreshing access tokens. This number is somewhat arbitrary but set to be a middle ground to avoid overly aggressive refreshes and a case where an access token expires before our fabricated expiration.  

a future improvement would be to check several places for the token expiration (potentially another API endpoint or within the access token JWT)

Also, in writing this, I opted to try to make as small a change as possible to minimally interfere with the existing code paths. As a result I updated the extraction method as opposed to adding code to an existing method that does expiration date parsing `_parse_token_expiration_date` which is not used by all code paths. 

## Summary by CodeRabbit

- **New Features**
  - Introduced default token expiration handling to improve token refresh behavior when expiration details are missing.

- **Tests**
  - Enhanced test coverage with scenarios for missing or expired token expiration fields.
  - Added fixed-time decorators for consistent expiry date testing.
  - Improved realism in token refresh tests by mocking HTTP responses.
  - Added parameterized tests for token expiry absence and updated expiry assertions for accuracy.

### DIFF
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -361,7 +361,7 @@ class AbstractOauth2Authenticator(AuthBase):
         """
         if current_depth > max_depth:
             # this is needed to avoid an inf loop, possible with a very deep nesting observed.
-            message = f"The maximum level of recursion is reached. Couldn't find the speficied `{key_name}` in the response."
+            message = f"The maximum level of recursion is reached. Couldn't find the specified `{key_name}` in the response."
             raise ResponseKeysMaxRecurtionReached(
                 internal_message=message, message=message, failure_type=FailureType.config_error
             )
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -147,7 +147,7 @@ class AbstractOauth2Authenticator(AuthBase):
     # ----------------
     # PRIVATE METHODS
     # ----------------
-    
+
     def _default_token_expiry_date(self) -> str:
         """
         Returns the default token expiry date
@@ -327,7 +327,9 @@ class AbstractOauth2Authenticator(AuthBase):
         Returns:
             The extracted token_expiry_date or None if not found.
         """
-        expires_in = self._find_and_get_value_from_response(response_data, self.get_expires_in_name())
+        expires_in = self._find_and_get_value_from_response(
+            response_data, self.get_expires_in_name()
+        )
         # If the access token expires in is None, we do not know when the token will expire
         # 1 hour was chosen as a middle ground to avoid unnecessary frequent refreshes and token expiration
         if expires_in is None:
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -130,7 +130,7 @@ class AbstractOauth2Authenticator(AuthBase):
         headers = self.get_refresh_request_headers()
         return headers if headers else None
 
-    def refresh_access_token(self) -> Tuple[str, Optional[str]]:
+    def refresh_access_token(self) -> Tuple[str, Union[str, int]]:
         """
         Returns the refresh token and its expiration datetime
 
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -130,7 +130,7 @@ class AbstractOauth2Authenticator(AuthBase):
         headers = self.get_refresh_request_headers()
         return headers if headers else None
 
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -130,17 +130,17 @@ def build_refresh_request_headers(self) -> Mapping[str, Any] | None:
         headers = self.get_refresh_request_headers()
         return headers if headers else None
 
-    def refresh_access_token(self) -> Tuple[str, Union[str, int]]:
+    def refresh_access_token(self) -> Tuple[str, Union[str, int, None]]:
```
dbgold17, line 133: got a little overzealous (ignore for now)

unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
```diff
@@ -4,9 +4,9 @@
 
 import json
 import logging
-from datetime import timedelta, timezone
```
dbgold17, line 7: unrelated unused import

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -308,17 +312,30 @@ def _extract_refresh_token(self, response_data: Mapping[str, Any]) -> Any:
         """
         return self._find_and_get_value_from_response(response_data, self.get_refresh_token_name())
 
-    def _extract_token_expiry_date(self, response_data: Mapping[str, Any]) -> Any:
+    def _extract_token_expiry_date(self, response_data: Mapping[str, Any]) -> AirbyteDateTime:
```
dbgold17, line 315: I decided to refactor the method to return a `AirbyteDateTime` in order to consolidate the expiration date parsing, validation, default behavior all in one method. The alternative would require having the default expiry set in multiple places in child classes and have to know about the expected expiry format (duration vs time of expiry)

unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
```diff
@@ -620,9 +674,16 @@ def test_given_message_repository_when_get_access_token_then_emit_message(
             token_expiry_date_format="YYYY-MM-DD",
             message_repository=message_repository,
         )
-        authenticator.refresh_access_token = mocker.Mock(
-            return_value=("new_access_token", "2023-04-04", "new_refresh_token")
+        # Mock the response from the refresh token endpoint
+        resp.status_code = 200
+        mocker.patch.object(
```
dbgold17, line 679: changed to mock at the request-level as opposed to the `refresh_access_token` method.

unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
```diff
@@ -614,15 +674,26 @@ def test_given_message_repository_when_get_access_token_then_emit_message(
         message_repository = Mock()
         authenticator = SingleUseRefreshTokenOauth2Authenticator(
             connector_config,
-            token_refresh_endpoint="foobar",
+            token_refresh_endpoint="https://refresh_endpoint.com",
             client_id=connector_config["credentials"]["client_id"],
             client_secret=connector_config["credentials"]["client_secret"],
+            token_expiry_is_time_of_expiration=True,
```
dbgold17, line 680: this is potentially a breaking change @bazarnov 

my understanding is that the intended behavior is that we require this parameter to be set to `True` in order to attempt to parse the expiry as a date (not a duration). 

However, it seems that the previous implementation of `SingleUseRefreshTokenOauth2Authenticator` did not do that and instead used the existence of the `token_expiry_date_format` parameter (not the actual value) to determine how to parse the expiry. Other OAuth classes implemented this differently, equivalent to how I'm doing it now.

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -308,17 +312,30 @@ def _extract_refresh_token(self, response_data: Mapping[str, Any]) -> Any:
...
+        )
+        # If the access token expires in is None, we do not know when the token will expire
+        if expires_in is None:
+            # If the token expiry date is set and the token is not expired, continue using the existing token expiry date
+            if self.get_token_expiry_date() and not self.token_has_expired():
+                return self.get_token_expiry_date()
+            else:
+                return self._default_token_expiry_date()
+        else:
+            return self._parse_token_expiration_date(expires_in)
```
bnchrch, line 338: 💅 Lets reduce the nesting / complexity
```suggestion
if expires_in is not None:
    return self._parse_token_expiration_date(expires_in)

# expires_in is None
expiry_date = self.get_token_expiry_date()
if expiry_date and not self.token_has_expired():
    return expiry_date

return self._default_token_expiry_date()
```

bnchrch (approved):
Only one nit (💅 )! Pre-Approving

PR #469, opened 2025-04-09
CI: Implement automatic dynamic versioning for pre-releases

aaronsteers:
Use Dunamai (https://dunamai.readthedocs.io) to calculate pre-release version when none is provided. 

Now, with this change, you can just leave the version number blank when executing a manual pre-release publish workflow. The CI job will auto-calculate a version number for you so there's no need to create a unique one yourself.

- Uses PEP 440 style for dynamic versioning to ensure compatibility with PyPI. Example: `6.45.0.post15.dev0+76a92a93`
- Version is guaranteed unique, as it includes the commit hash as well as the distance from the last-published release version.
- Version number is printed in the CI job within the job output.

To test locally, you can run this command:
```bash
$ poetry run dunamai from git --style pep440
6.45.0.post15.dev0+76a92a93
```

----

Sample CI execution here: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/14368872080/job/40287903080?pr=469

image

## Summary by CodeRabbit

- **Chores**
	- Streamlined the deployment process by enhancing release detection and reordering steps for clearer version validation.
	- Upgraded version management to adhere to PEP 440 standards, ensuring consistent and compatible versioning.
	- Updated workflow name from "Packaging and Publishing" to "CDK Publish" for improved clarity and organization.
	- Added steps for checking out the CDK repository and detecting prerelease versions.
	- Updated documentation to clarify instructions for publishing pre-release versions, including automatic version selection.

PR #471, opened 2025-04-10
ci(hotfix): update version detection format in workflow

aaronsteers:
Modify the version detection format in the workflow to include additional formatting options and comment out the tag trigger.


.github/workflows/pypi_publish.yml
```diff
@@ -9,8 +9,8 @@ name: CDK Publish
 
 on:
   push:
-    tags:
-      - "v*"
+    # tags:
+    #   - "v*"
```
aaronsteers, line 13:
```suggestion
    tags:
      - "v*"
```

aaronsteers:
Force-merging as hotfix to resolve format issue discovered after merging:

- #469

PR #472, opened 2025-04-10
feat(low-code): added key_transformation to DpathFlattenFields

darynaishchenko:
### What
neede for source-hubspot migration to low code. 
We already have `DpathFlattenFields` to extract object fields, but with current implementation we can't customize extracted object keys. 
[Current implementation in hubspot with RecordUnnester.](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py#L121)
### How
added `key_transformation` option to `DpathFlattenFields`.
Example:
```
      - type: DpathFlattenFields
        key_transformation:
          type: KeyTransformation
          prefix: merged_from_email_
        field_path:
          - merged_from_email
      - type: DpathFlattenFields
        key_transformation:
          type: KeyTransformation
          prefix: merged_to_email_
        field_path:
          - merged_to_email
```

## Summary by CodeRabbit

- **New Features**
	- Introduced a configurable key transformation option that allows dynamic renaming of keys during field removal and flatten operations, with support for prefixes, suffixes, and custom formats. If no transformation is provided, original keys remain unchanged.

- **Tests**
	- Expanded test coverage to ensure the new key transformation behavior works correctly under various scenarios, including new test cases for specific transformation formats.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2307,6 +2307,33 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
+  KeyTransformation:
+    title: Transformation to apply for extracted object keys by Dpath Flatten Fields
+    type: object
+    required:
+      - type
+    properties:
+      type:
+        type: string
+        enum: [ KeyTransformation ]
+      prefix:
+        title: Key Prefix
+        description: Prefix to add for object keys. If not provided original keys remain unchanged.
+        type: string
+        examples:
+          - flattened_
+      suffix:
+        title: Key Suffix
+        description: Suffix to add for object keys. If not provided original keys remain unchanged.
+        type: string
+        examples:
+          - _flattened
+      custom:
+        title: Custom Transformation for keys
+        description: Custom transformation. Can be used with {{ key }} as a original value for key name. If not provided original keys remain unchanged.
+        type: string
+        examples:
+          - flattened_{{ key }}
   DpathFlattenFields:
     title: Dpath Flatten Fields
     description: A transformation that flatten field values to the to top of the record.
@@ -2338,9 +2365,8 @@ definitions:
       key_transformation:
         title: Key transformation
         description: Transformation for object keys. If not provided, original key will be used.
-        type: string
-        examples:
-          - "flattened_{{ key }}"
+        type: object
+        "$ref": "#/definitions/KeyTransformation"
       $parameters:
         type: object
         additionalProperties: true
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2335,6 +2335,12 @@ definitions:
         title: Replace Origin Record
         description: Whether to replace the origin record or not. Default is False.
         type: boolean
+      key_transformation:
+        title: Key transformation
+        description: Transformation for object keys. If not provided, original key will be used.
+        type: string
+        examples:
+          - "flattened_{{ key }}"
       $parameters:
         type: object
         additionalProperties: true
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2328,12 +2328,6 @@ definitions:
         type: string
         examples:
           - _flattened
-      custom:
-        title: Custom Transformation for keys
-        description: Custom transformation. Can be used with {{ key }} as a original value for key name. If not provided original keys remain unchanged.
-        type: string
-        examples:
-          - flattened_{{ key }}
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -897,6 +897,14 @@ class DpathFlattenFields(BaseModel):
         description="Whether to replace the origin record or not. Default is False.",
         title="Replace Origin Record",
     )
+    key_transformation: Optional[str] = Field(
```
brianjlai, line 900: I have some concerns about using interpolation in this way for the `key_transformation`. I don't think we make it very discoverable that `key` is a valid interpolation value except if you read the documentation. 

Rather than having `key_transformation` be another jinja interpolation to manage from `DpathFlattenFields`, I think we should call this `key_transformation_prefix`, which would then take in `"flattened_`. And from there in the code below, we just append `flattened_` with the key (w/o creating a new InterpolatedString).

It's a little less flexible because if we need to allow for adding suffix, we need to add a new fields `key_transformation_suffix`, but this allows us to have less jinja magic and variables for the user to keep track of.

brianjlai (commented):
Mainly just one comment around how we can try to move away from some jinja interpolation magic that the user has to understand, but this is a good feature for the CDK and I image other connectors might use this so this makes sense to have here instead on a hubspot custom component

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
darynaishchenko: I got your point, it also makes sense to me. Maybe we can add both `key_prefix`, `key_suffix` and `key_transformation`?So we have user-friendly  `key_prefix`,  `key_suffix` and more flexible `key_transformation`. What do you think?

brianjlai: yep that sounds good to me!

airbyte_cdk/sources/declarative/transformations/dpath_flatten_fields.py
```diff
@@ -24,17 +32,43 @@ class DpathFlattenFields(RecordTransformation):
...
+    def _apply_key_transformation(self, extracted: Mapping[str, Any]) -> Mapping[str, Any]:
+        if self.key_transformation:
+            if self.key_transformation.prefix:
+                prefix = InterpolatedString.create(
+                    self.key_transformation.prefix, parameters=self._parameters
+                ).eval(config=self.config)
+                extracted = {f"{prefix}{key}": value for key, value in extracted.items()}
+
+            if self.key_transformation.suffix:
+                suffix = InterpolatedString.create(
```
brianjlai, line 57: Instead of instantiating the `InterpolatedString` for both the prefix and suffix every time we call `transform()` what I think we should do is use a `__post_init__()` method on the `KeyTransformation` class that takes in the prefix and suffix strings and instantiates InterpolatedStrings. And then from this method we should just need to call `self.key_transformation.prefix.eval()`.

Can we also add a check so that if `prefix.eval()` (+suffix) is an empty string or None, then we should not perform the transformation

brianjlai (commented):
A few final small comments, but the interface looks good!

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2307,6 +2307,33 @@ definitions:
...
+        type: string
+        examples:
+          - flattened_
+      suffix:
+        title: Key Suffix
+        description: Suffix to add for object keys. If not provided original keys remain unchanged.
+        type: string
+        examples:
+          - _flattened
+      custom:
```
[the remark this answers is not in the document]
darynaishchenko: removed

airbyte_cdk/sources/declarative/transformations/dpath_flatten_fields.py
darynaishchenko: added `__post_init__()` and updated checks in `_apply_key_transformation()`

airbyte_cdk/sources/declarative/transformations/dpath_flatten_fields.py
```diff
@@ -8,6 +8,24 @@
 from airbyte_cdk.sources.types import Config, StreamSlice, StreamState
 
 
+@dataclass
+class KeyTransformation:
+    config: Config
+    parameters: InitVar[Mapping[str, Any]]
+    prefix: Union[InterpolatedString, str, None] = None
```
brianjlai, line 15: Nit, but for `prefix` and `suffix`, since the YAML schema is defined to only accept strings, we know that this will only ever be a string or None, so we should be able to use `Optional[str]`. If mypy doesn't complain lets switch it to that.

brianjlai (approved):
small nit to address, but no need to block on approval. thanks for the changes and looks good to me!

PR #476, opened 2025-04-11
feat: hmac hashing jinja filter

natikgadzhi:
## What

This PR adds hmac hashing filter! We need this for TikTok Shops API because reasons /shrug

### DIFF
--- a/airbyte_cdk/sources/declarative/interpolation/filters.py
+++ b/airbyte_cdk/sources/declarative/interpolation/filters.py
@@ -4,9 +4,10 @@
 
 import base64
 import hashlib
+import hmac as hmac_lib
 import json
 import re
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
 
 def hash(value: Any, hash_type: str = "md5", salt: Optional[str] = None) -> str:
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/interpolation/filters.py
```diff
@@ -135,5 +136,51 @@ def regex_search(value: str, regex: str) -> str:
...
+    auth_headers:
+      $ref: "#/definitions/base_auth"
+      $parameters:
+        signature: "{{ 'message_to_sign' | hmac_hash('my_secret_key') }}"
+
+    :param value: The message to be authenticated
+    :param key: The secret key for the HMAC
+    :param hash_type: Hash algorithm to use (default: sha256)
+    :return: HMAC digest as a hexadecimal string
+    """
```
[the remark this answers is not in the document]
natikgadzhi: true

airbyte_cdk/sources/declarative/interpolation/filters.py
```diff
@@ -135,5 +136,51 @@ def regex_search(value: str, regex: str) -> str:
...
+
+    This filter creates a Hash-based Message Authentication Code (HMAC) using a cryptographic
+    hash function and a secret key. Currently only supports SHA-256, and returns hexdigest of the signature.
+
+    Example usage in a low code connector:
+
+    auth_headers:
+      $ref: "#/definitions/base_auth"
+      $parameters:
+        signature: "{{ 'message_to_sign' | hmac('my_secret_key') }}"
```
bnchrch, line 151: Ah gotcha, so to use this in many cases its up to the caller to create the list of query params by hand to sign.

Think thats a fine work around.

But it seems like the next win is
1. a jinja function to get all query params
2. a jinja function to transform it into a delimited string (after you optionally sort / filter / map the list)

bnchrch (approved):
No argument here. Its additive, I trust you that it works, and this is your last day so no need to go back and forth.

airbyte_cdk/sources/declarative/interpolation/filters.py
natikgadzhi: Yes, that would indeed make things easier.

PR #482, opened 2025-04-14
fix: Ensure page token 0 is being passed

maxi297:
## What

Following a couple of tickets, it seems like the injection on the first page stopped working. Here are some of the tickets:
* https://github.com/airbytehq/airbyte/issues/51505
* https://github.com/airbytehq/airbyte/issues/53140

## How

This was caused by the first token being `0` which gets evaluated as `False` and therefore is not injected. We now validate `None` explicitly instead

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of pagination tokens to correctly support cases where the initial token is zero.
- **Tests**
	- Added a test to ensure pagination works properly when the initial token is zero.


brianjlai (approved):
🚢

PR #484, opened 2025-04-15
fix(OffsetIncrement Pagination Strategy): Fix bug where streams that use record filtering do not paginate correctly

brianjlai:
## What

While migrating `source-hubspot` @tolik0 identified a bug in how we perform OffsetIncrement pagination. When we perform record filtering, we can potentially end up with fewer records than the amount returned in the API response. The paginator then falsely thinks the paging is complete because received records is smaller than the max page size.

## How

This PR fixes the said problem above by utilizing the existing `RecordExtractor` model defined by the developer and utilizes it in the `OffsetIncrement` pagination strategy. While determine if there are more pages to read in `next_page_token()` we have the extractor parse the `response` parameter (which will not perform filtering) to get the accurate count.

We opted for this solution because reusing an extractor is more accurate than recreating the dpath parsing behavior.

## Summary by CodeRabbit

- **New Features**
  - Pagination strategies now support extracting records from responses using an extractor, improving handling of paginated data.

- **Bug Fixes**
  - Pagination logic now correctly handles cases where the expected record path is missing in the response.

- **Tests**
  - Enhanced test coverage for pagination, including realistic response scenarios and validation of extractor integration.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2054,6 +2054,7 @@ class ModelToComponentFactory:
         config: Config,
         *,
         url_base: str,
+        extractor_model: Optional[Union[CustomRecordExtractorModel, DpathExtractorModel]] = None,
         decoder: Optional[Decoder] = None,
         cursor_used_for_stop_condition: Optional[DeclarativeCursor] = None,
     ) -> Union[DefaultPaginator, PaginatorTestReadDecorator]:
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2054,6 +2054,7 @@ def create_default_paginator(
         config: Config,
         *,
         url_base: str,
+        extractor_model: Optional[Union[CustomRecordExtractorModel, DpathExtractorModel]] = None,
```
brianjlai, line 2057: It may seem weird that this is an `Optional` component since for non-custom component manifests `RecordSelector` and `RecordExtractor` are required components. However, a custom retriever that instantiates a `DefaultPaginator` won't be able to easily pass in the extractor model so it is a little easier for this to be optional and default to the existing behavior.

I know it sort of sucks, but wanted to call it out and if anyone has a better solution, I'm happy to defer to that

tolik0 (approved):

PR #486, opened 2025-04-16
fix: (CDK) (Manifest) - Add `deprecations` support and handle `deprecation warnings`; deprecate `url_base`, `path`, `request_body_json` and `request_body_data` for `HttpRequester`

bazarnov:
## What
Resolves:
- https://github.com/airbytehq/airbyte-internal-issues/issues/12407
- https://github.com/airbytehq/airbyte-internal-issues/issues/12357

### Updated Component Models:
- Modified the `HttpRequester` to support the new `url` field and prioritize it.
- Updated schema definitions, validation logic, and deprecated fields in models.
- Added logic to handle warnings during component creation (e.g., get_model_deprecations).
- Ensure the backward compatibility with the deprecated `url_base` and `path` fields.

### Handling Deprecation Warnings:
- Added a new `deprecation_warnings` method to the `DeclarativeSource` and `ManifestDeclarativeSource` class to collect and process deprecation notices.
- Updated the `_categorise_groups()` to handle deprecation warnings alongside message groups.
- Added logic to collect these warnings in components using `BaseModelWithDeprecations`.
- The new `airbyte_cdk/sources/declarative/models/base_model_with_deprecations.py` module is added to cover the `deprecation` warnings for the fields marked as `deprecated`. This class is static (not autogenerated) and re-used when the `declarative_component_schema.py` is being generated. In this way, there is no need to set additional validations on top of the existing component implementations.

## User Impact
- No impact is expected, this is not a breaking change.

## Deprecation Warning examples
The deprecation rules applied to the `declarative_component_schema.yaml` are automatically picked up and propagated to the `declarative_component_schema.py` (auto-gen models), see the example bellow.

```yaml
# declarative component schema content

HttpRequester:
  required:
    - url
    # - url_base # removed from the required fields
  # .. the rest of the component content
  properties:
    url_base:
        deprecated: true
        deprecation_message: "Use `url` field instead."
        # .. the rest of the deprecated field content
    url:
       # .. the content of the deprecated field replacement
```

```python
class HttpRequester(BaseModelWithDeprecations):
    type: Literal["HttpRequester"]
    url_base: Optional[str] = Field(
        None,
        deprecated=True, # deprecation flag is set onto the field
        deprecation_message="Use `url` field instead.", # deprecation message is set onto the field, will be used as a part of the warning.
        description="....",
        examples=[],
        title="API Base URL",
    )
    url: Optional[str] = Field(
        None,
        description="....",
        examples=[],
        title="API URL",
    )
```

## Warning example
When `using` the deprecated field or `creating` an instance of the class with the deprecated field.

Python default behavior:

```
DeprecationWarning: Component type: `HttpRequester`. Field 'url_base' is deprecated. Use `url` field instead.
```
If multiple fields are marked as `deprecated` and used - more than 1 deprecation message is shown.
```
DeprecationWarning: Component type: `HttpRequester`. Field 'url_base' is deprecated. Use `url` field instead.

DeprecationWarning: Component type: `HttpRequester`. Field 'path' is deprecated. Use `url` field instead.
```

Connector Builder behavior:

Screenshot 2025-04-14 at 21 10 37

## Summary by CodeRabbit

- **New Features**
  - Deprecation warnings for deprecated fields are now integrated and displayed in the Connector Builder UI.
  - Added support for a new `url` field in HTTP requesters, replacing the deprecated `url_base` and `path` fields.
  - Introduced a new model base class to automatically emit deprecation warnings for deprecated fields.
  - Enhanced component factory to collect and expose deprecation warnings from models.
  - Added a method to retrieve deprecation warnings from sources for improved user notifications.
  - Added unified `request_body` support for HTTP requests, replacing deprecated `request_body_data` and `request_body_json`.
- **Bug Fixes**
  - Improved handling and aggregation of deprecation warnings to ensure clear user notifications.
- **Tests**
  - Added coverage for deprecation warning scenarios in source mocks.
  - Added tests for request body interpolation using the new `RequestBody` model abstraction.
- **Chores**
  - Enhanced code generation to detect and handle deprecated fields automatically in generated models.


airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1886,9 +1888,30 @@ definitions:
           - "{{ config['base_url'] or 'https://app.posthog.com'}}/api"
           - "https://connect.squareup.com/v2/quotes/{{ stream_partition['id'] }}/quote_line_groups"
           - "https://example.com/api/v1/resource/{{ next_page_token['id'] }}"
+      url:
+        sharable: true
```
lmossman, line 1892: This is supposed to be `linkable` now, right?

Also, I don't think the `url` field should actually be linkable, because we now expect it to also include the endpoint path, and we don't expect that to be common across streams.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1886,9 +1888,30 @@ definitions:
...
+        title: API URL
+        description: The URL of the API source. Do not put sensitive information (e.g. API tokens) into this field - Use the Authentication component for this.
```
lmossman, line 1894: I think the title and description should reflect that this is the URL of an API _endpoint_ specifically, not just the API

lmossman (commented):
Had a couple small thoughts on the declarative_component_schema - will leave python review to @maxi297 or someone else on the python side

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
bazarnov: That is a leftover from `un-merging` of the PoC PR,, fixed now.

airbyte_cdk/connector_builder/test_reader/reader.py
```diff
@@ -298,6 +308,17 @@ def _categorise_groups(self, message_groups: MESSAGE_GROUPS) -> GROUPED_MESSAGES
                 case _:
                     raise ValueError(f"Unknown message group type: {type(message_group)}")
 
+        # process deprecation warnings, if present
+        if deprecation_warnings is not None:
+            for deprecation in deprecation_warnings:
+                match deprecation:
+                    case AirbyteLogMessage():
```
maxi297, line 315: nit: Why is the typing here different than from line 117? It seems like we wouldn't have to worry about the type not matching here if we could rely on the type.

airbyte_cdk/connector_builder/test_reader/reader.py
```diff
@@ -238,7 +243,11 @@ def _check_record_limit(self, record_limit: Optional[int] = None) -> int:
 
         return record_limit
 
-    def _categorise_groups(self, message_groups: MESSAGE_GROUPS) -> GROUPED_MESSAGES:
+    def _categorise_groups(
+        self,
+        message_groups: MESSAGE_GROUPS,
+        deprecation_warnings: Optional[List[Any]] = None,
```
maxi297, line 249: nit: it seems like the deprecation warnings are not necessary for categorizing groups. Should we just do `log_messages.append(deprecation_warnings)` from `run_test_read`?

airbyte_cdk/sources/declarative/models/base_model_with_deprecations.py
```diff
@@ -0,0 +1,131 @@
...
+
+        message = f"Component type: `{self.__class__.__name__}`. Field '{field_name}' is deprecated. {message}"
+        # Emit a warning message for deprecated fields (to stdout) (Python Default behavior)
+        warnings.warn(message, DeprecationWarning)
+        # Create an Airbyte deprecation log message
+        deprecation_log_message = AirbyteLogMessage(level=Level.WARN, message=message)
+        # Add the deprecation message to the Airbyte log messages,
+        # this logs are displayed in the Connector Builder.
+        if deprecation_log_message not in self._deprecation_logs:
+            # Avoid duplicates in the deprecation logs
```
maxi297, line 130: Given this is instantiate once per model, I assume this will log multiple times even if it is the same error. Should we have `_deprecation_logs` as a private module variable instead of a private class field to avoid that? 

I see that the model_to_component_factory has a way to avoid duplication but is it handled internally by `warnings.warn` for the print to stdout?

airbyte_cdk/sources/declarative/models/base_model_with_deprecations.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        Show a warning message for deprecated fields (to stdout).
+        Args:
+            field_name (str): Name of the deprecated field.
+            message (str): Warning message to be displayed.
+        """
+
+        message = f"Component type: `{self.__class__.__name__}`. Field '{field_name}' is deprecated. {message}"
+        # Emit a warning message for deprecated fields (to stdout) (Python Default behavior)
+        warnings.warn(message, DeprecationWarning)
```
maxi297, line 124: I don't know about the `warnings` module. Will this be printed in stdout? If so, this does not align with the Airbyte protocol. Should we care about this?

maxi297 (commented):
I have a couple of questions to make sure I understand but overall, it looks good

maxi297 (commented):
One other concerns: I don't see any tests for that. Should we add some?

airbyte_cdk/sources/declarative/models/base_model_with_deprecations.py
bazarnov: It's printed once, as of the `default` behavior for Python and yes it goes to `stdout`. It's shouldn't imply the Airbyte expectations in any manner.

airbyte_cdk/sources/declarative/models/base_model_with_deprecations.py
bazarnov: Handled.

maxi297 (approved):
LGTM! Thanks for tackling the changes

airbyte_cdk/connector_builder/test_reader/reader.py
bazarnov: Re-worked. Use the direct type of `LogMessage` as `ConnectorBuilderLogMessage` and append directly. Thanks.

airbyte_cdk/connector_builder/test_reader/reader.py
bazarnov: yeap!

PR #487, opened 2025-04-16
fix(property chunking): Switch the ordering page iteration and property chunking process chunks first instead of pages first

brianjlai:
## What

While implementing the 3 Hubspot property history streams which require usage of property chunking, I noticed some strange behavior where there would be a very long pause between the start of the sync and the first record emitted whenever property chunking was enabled.

This change refactors the `SimpleRetriever` so we process each of the property chunks of the current page horizontally instead of vertically

## How

The way I had originally written the code where the property chunk is the outer loop and the inner loop is the pagination iteration leads to us taking the first property chunk and paginating all the way to the end. We then iterate to the second property chunk and paginate all the way to the end. And this continues for as many chunks. However, as you can see, this means that we effectively cannot emit any records until we reach the last property chunk.

This change reworks the order of how we process pages and chunks so that we start with the current page, fetch each property chunk for that page, then emit the complete merged record. We then continue to the next page. This is done by moving where we perform property chunking from `SimpleRetriever.read_records()` and into `SimpleRetriever._read_pages()`

Note: 
You'll see there are no changes to tests, I had already written test cases for each of the property chunking and no chunking scenarios and since this is a refactor the intent is for there to be no functional impact on what records are emitted.

## Summary by CodeRabbit

- **New Features**
  - Enhanced support for additional query property chunking to improve data retrieval flexibility.
- **Refactor**
  - Simplified record reading logic by centralizing property chunking and merging, improving efficiency and maintainability.
  - Adjusted property chunk size calculation to better handle character limits by accounting for delimiters.
- **Tests**
  - Updated tests to include nested dictionary and list structures for more comprehensive data handling validation.
  - Modified property chunking tests to reflect updated chunking behavior with character limits.


tolik0 (approved):

airbyte_cdk/sources/declarative/requesters/query_properties/property_chunking.py
```diff
@@ -52,8 +52,9 @@ def get_request_property_chunks(
         chunk_size = 0
         for property_field in property_fields:
             # If property_limit_type is not defined, we default to property_count which is just an incrementing count
+            # +2 for the comma ToDo: Add possibility to specify parameter representation and take it into account in property_field_size
             property_field_size = (
-                len(property_field)
+                len(property_field) + 2
```
brianjlai, line 57: nit, @tolik0 i think can actually just be one. Because the delimiter in https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/requesters/request_options/interpolated_request_options_provider.py#L105-L107 is actually just one `,` and not a space. So 1 is probably sufficient. 

If we still run into request URL issues during the migration, we can always make it smaller than 15000, just to buffer some space for the rest of the non property chunked URL like the `https://api.hubspot...etc`. I'll change this to +1 and adjust the comment and any tests.

PR #490, opened 2025-04-17
fix: Properly setup global substream cursor based on manifest

maxi297:
This PR allows for the support of global cursors when running in the concurrent engine

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved global cursor handling to support explicit overrides for enhanced data synchronization.
- **Tests**
  - Added a test verifying global cursor usage and accurate state management without per-partition states.


lazebnyi (approved):
lgtm

PR #501, opened 2025-04-22
fix: only check for file upload if stream is declarative

maxi297:
Given sources that are in transition from being declarative, we will have streams that are not in the manifest and therefore we will have a `KeyError: '<stream name>'` when grouping the streams

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type checking to prevent errors when determining file transfer support for streams.

### DIFF
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -211,8 +211,8 @@ class ConcurrentDeclarativeSource(ManifestDeclarativeSource, Generic[TState]):
             # so we need to treat them as synchronous
 
             supports_file_transfer = (
-                isinstance(declarative_stream, DeclarativeStream) and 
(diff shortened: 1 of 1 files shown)


aldogonzalez8 (approved):
APPROVED

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -211,7 +211,8 @@ def _group_streams(
             # so we need to treat them as synchronous
 
             supports_file_transfer = (
-                "file_uploader" in name_to_stream_mapping[declarative_stream.name]
+                isinstance(declarative_stream, DeclarativeStream)
+                and "file_uploader" in name_to_stream_mapping[declarative_stream.name]
```
aldogonzalez8, line 215: nit: this makes sense but what about use `.get()`?

maxi297: I think we expect this to be defined at this point so I'm fine with keeping the [] to communicate that

PR #503, opened 2025-04-23
feat(cdk): connector builder support for file uploader

aldogonzalez8:
## What
We want to implement support for the connector builder to communicate about the files that will be uploaded for the streams.

Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/12424

## How

**Current Implementation:**

*   **Runtime (`emit_connector_builder_messages=False`):**
    *   `FileUploader` is instantiated with a standard `FileWriter`, which performs the actual file writing.
*   **Connector Builder (`emit_connector_builder_messages=True`):**
    *   `FileUploader` is instantiated with a `NoopFileWriter`, preventing actual file writes.
    *   This `FileUploader` instance is then wrapped by `ConnectorBuilderFileUploader`.
    *   `ConnectorBuilderFileUploader.upload()` first calls the wrapped `FileUploader.upload()` (which populates `record.file_reference` but doesn't write the file) and then performs the Connector Builder-specific action of copying `file_reference` attributes into `record.data`.

## Summary by CodeRabbit

- **New Features**
  - Introduced a flexible file upload system with support for custom file writing strategies and enhanced file reference metadata in records.
  - Added new file writer options, including a no-operation writer and a local filesystem writer, to control file storage behavior.
  - Improved integration for connector builder workflows, enabling file reference fields to be included in record data.

- **Bug Fixes**
  - Enhanced test coverage to ensure correct file reference handling and metadata propagation in connector builder scenarios.


airbyte_cdk/sources/declarative/yaml_declarative_source.py
```diff
@@ -24,6 +24,7 @@ def __init__(
         catalog: Optional[ConfiguredAirbyteCatalog] = None,
         config: Optional[Mapping[str, Any]] = None,
         state: Optional[List[AirbyteStateMessage]] = None,
+        emit_connector_builder_messages: Optional[bool] = False,
```
maxi297, line 27: This is interesting. I guess this isn't needed yet (until we do https://github.com/airbytehq/airbyte-python-cdk/pull/460) but we wanted to preemptively set this up, is that correct?

airbyte_cdk/sources/declarative/retrievers/file_uploader/base_file_writer.py
```diff
@@ -0,0 +1,19 @@
+#
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+#
+
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+
+class BaseFileWriter(ABC):
```
maxi297, line 9: `Base` does not seem in our vocabulary for interfaces. Could we have something like this? 

* Parent: FileWriter
* LocalFileSystemFileWriter
* NoopFileWriter

airbyte_cdk/sources/declarative/retrievers/file_uploader/file_uploader.py
```diff
@@ -77,9 +87,7 @@ def upload(self, record: Record) -> None:
             full_path = files_directory / file_relative_path
```
maxi297, line 87: Comment unrelated to this line but didn't know where to put it: On line 75, it seems like we have a TODO. Can we at least improve the error message?

airbyte_cdk/sources/declarative/retrievers/file_uploader/base_file_uploader.py
```diff
@@ -0,0 +1,22 @@
...
+#
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+from airbyte_cdk.sources.declarative.types import Record
+
+
+@dataclass
+class BaseFileUploader(ABC):
```
maxi297, line 12: `Base` does not seem in our vocabulary for interfaces. Could we have something like this? 

* Parent: FileUploader
* DefaultFileUploader (this naming is not very good but we have history of naming things as `Default` when we don't know (like [here](https://github.com/airbytehq/airbyte-python-cdk/blob/2f29eff3f5d74862e1667f6274a0f23d1b3a6c0c/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py#L33))
* ConnectorBuilderFileUploader

airbyte_cdk/sources/declarative/retrievers/file_uploader/noop_file_writer.py
```diff
@@ -0,0 +1,16 @@
...
+-from .file_uploader import BaseFileWriter
++from .base_file_writer import BaseFileWriter
+
+
+class NoopFileWriter(BaseFileWriter):
+    def write(self, file_path: Path, content: bytes) -> int:
+        """
+        Noop file writer
+        """
+        return 0
```
maxi297, line 16: Does that mean that the size shown in the ConnectorBuilder will always be shown as 0? Should we set something more obvious that this is a Connector Builder test like `-1`?

maxi297 (commented):
I have a couple of comments but overall it looks like we are on track

airbyte_cdk/sources/declarative/retrievers/file_uploader/base_file_writer.py
aldogonzalez8: Ok, i have fixed names to match our vocabulary.

airbyte_cdk/sources/declarative/retrievers/file_uploader/noop_file_writer.py
aldogonzalez8: Ok, I have created a Constant with value -1 for this purpose.

airbyte_cdk/sources/declarative/retrievers/file_uploader/file_uploader.py
aldogonzalez8: Ok, I have improved the error message.

airbyte_cdk/sources/declarative/yaml_declarative_source.py
aldogonzalez8: I just needed a better patching, so there's no need to make any changes for now in this class.  Removed this.

airbyte_cdk/sources/declarative/retrievers/file_uploader/default_file_uploader.py
```diff
@@ -62,7 +72,9 @@ def upload(self, record: Record) -> None:
         )
 
         if self.content_extractor:
-            raise NotImplementedError("TODO")
+            raise NotImplementedError(
```
maxi297, line 75: I think I'm dumb. My proposal for this change should have beenn: Should we just remove `content_extractor` then?

maxi297 (approved):
LGTM! :shipit:

airbyte_cdk/sources/declarative/retrievers/file_uploader/default_file_uploader.py
aldogonzalez8: Ok, I have removed it, I think it doesn't break anything. I couldn't find it in the declarative_component_schema.yaml

maxi297: I think I initially added that in case the file content is actually returned as part of a field in a HTTP request like `VersionData` in Salesforce https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contentversion.htm

PR #507, opened 2025-04-25
feat(property chunking): Allow PropertiesFromEndpoint to be defined on HttpRequester

brianjlai:
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/12646

While implementing CRMSearch streams for Hubspot, we found a gap where we need to be able to reference the properties retrieved dynamically to inject them into other request inputs. Specifically JSON body for POST requests.

Within these other request inputs, property chunking itself doesn't really make sense because there aren't input limits unlike request parameters.

This implementation does a clever trick where under the hood we reuse the `QueryProperties` runtime component without any property chunking enabled. And by doing so, nearly all of the rest of the flow can be reused. And the interpolation context for the request inputs should still have access to the stream slice. For example: `{{ stream_partition.extra_fields['query_properties'] }}`

## Summary by CodeRabbit

- **New Features**
  - Added support for dynamically fetching properties from an API endpoint and injecting them into outbound requests.
- **Bug Fixes**
  - Improved validation to prevent conflicting definitions of dynamic property fetching in configurations.
- **Tests**
  - Enhanced test coverage for dynamic property fetching scenarios and validation of conflicting configurations.
- **Refactor**
  - Streamlined internal logic for detecting and handling dynamic property fetching configurations.
- **Style**
  - Cleaned up formatting in configuration schema for improved readability.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2967,6 +2967,18 @@ def create_simple_retriever(
                 model.requester.request_parameters = self._remove_query_properties(
                     model.requester.request_parameters
                 )
+        elif model.requester.fetch_properties_from_endpoint:
```
brianjlai, line 2970: This does apply a bit of an undocumented behavior where if a person were to define `requester.request_parameters.query_properties` and `requester.fetch_properties_from_endpoint` then we would default to the `request_parameters`

It might make sense to throw an error if both are defined, but right now allowing them to be defined in either places is a little easier and we don't have to go back and modify existing connectors

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2967,6 +2967,21 @@ def create_simple_retriever(
                 model.requester.request_parameters = self._remove_query_properties(
                     model.requester.request_parameters
                 )
+        elif (
+            hasattr(model.requester, "fetch_properties_from_endpoint")
+            and model.requester.fetch_properties_from_endpoint
```
brianjlai, line 2972: This does apply a bit of an undocumented behavior where if a person were to define requester.request_parameters.query_properties and requester.fetch_properties_from_endpoint then we would default to the request_parameters

It might make sense to throw an error if both are defined, but right now allowing them to be defined in either places is a little easier and we don't have to go back and modify existing connectors

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1484,6 +1484,9 @@ definitions:
             examples:
               - "{{ record.id }}/{{ record.file_name }}/"
               - "{{ record.id }}_{{ record.file_name }}/"
+          $parameters:
```
brianjlai, line 1487: This was rather interesting. I was seeing a failing mypy run locally and on CI where it was complaining in the model to component factory that the `FileUploaderModel` didn't have a `parameters` field. And that lines up with our `declarative_component_schema.py` file not having it either. So I have no idea how CI was passing before this, but I added it in to fix it here.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
tolik0: Why would throwing an error require modifying existing connectors?

brianjlai: sorry i should have made this more clear, the error message would not affect existing connectors, but if we were to remove allowing for`PropertiesFromEndpoint` to be specified from the `retriever.requester.request_properties` since it can now be defined in `retriever.requester`. That would be a breaking change.

I'll add an error message, but will allow it to be defined in two places for now

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1978,6 +1978,10 @@ definitions:
           - "$ref": "#/definitions/SelectiveAuthenticator"
           - "$ref": "#/definitions/CustomAuthenticator"
           - "$ref": "#/definitions/LegacySessionTokenAuthenticator"
+      fetch_properties_from_endpoint:
```
maxi297, line 1981: I know we were flagging things as experimental in the CDK until they were added to the Connector Builder and that with Lake's project, it is not necessarily possible anymore as they would be automatically added. Is there still a concern here of flagging this as a feature that is new?

maxi297 (approved):
LGTM! Thanks for enabling this for request body too

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
brianjlai: I think since its optional and non-breaking I'm fine leaving it a bit undocumented as new. I'm not super concerned for that reason. I don't expect many customer made connectors to need to enable this

Maybe there's a broader questions about how we socialize new features, but maybe we kick that can down the road.

brianjlai:
Anatolii tested the feature w/ `associations` stream. Appears to work so we should be set to merge

PR #508, opened 2025-04-26
fix: yield partitions for unique stream slices in StreamSlicerPartitionGenerator

lazebnyi:
Some parent streams may return duplicate record IDs in the response. As a result, we can end up with several identical partitions. After one partition is processed, the next one causes an error because it is already closed.  

For example, in the `source-stripe` connector, the `payout_balance_transactions` stream depends on `balance_transactions`, which uses the `events` endpoint to fetch data incrementally. This can lead to multiple events for the same parent with the same state but different values in other fields.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved partition generation to eliminate duplicate partitions when identical stream slices are encountered.

### DIFF
--- a/airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
+++ b/airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
@@ -99,7 +99,7 @@ class StreamSlicerPartitionGenerator(PartitionGenerator):
             yield self._partition_factory.create(stream_slice)
 
     @staticmethod
-    def _make_hashable(obj: Any) -> Hashable:
+    def _make_hashable(obj: Any) -> Any:
         if isinstance(obj, dict):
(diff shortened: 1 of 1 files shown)


maxi297 (approved):
Accepting right now to unblock oncall issue but I think this will requires some tests

maxi297:
Do we fear this set could increase in size and cause memory issues at some point?

lazebnyi:
> Do we fear this set could increase in size and cause memory issues at some point? Else, I thought about grouping partition routing but I fear it would mess the state

In theory, yes, we could have memory issues since we store each partition. So, maybe we should introduce a flag to avoid using duplicate validation in cases where we are sure that parent records are unique, or even consider switching off concurrency for them.

Alternatively, we could use a key-value storage for these cases, especially if it’s a tier0 or tier1 user.

maxi297:
Will be reverted as part of https://github.com/airbytehq/airbyte-python-cdk/pull/510 in favor of https://github.com/airbytehq/airbyte-python-cdk/pull/509

PR #510, opened 2025-04-26
chore: revert generate unique partitions added in #508

lazebnyi:
This revert changes added in https://github.com/airbytehq/airbyte-python-cdk/pull/508

PR #508 attempted to address an issue where some parent streams could return duplicate record IDs, resulting in multiple identical partitions. Once one partition was processed and closed, subsequent attempts to process a duplicate partition would cause errors.

## Summary by CodeRabbit

- **Refactor**
  - Updated partition generation to include all stream slices, removing previous duplicate filtering logic.


maxi297 (approved):
:shipit:

PR #512, opened 2025-04-26
feat(cdk): handle include_files from configured catalog

aldogonzalez8:
## What
There is a scenario where the user may not want to include files from a file stream, so they will only get metadata. We were not handling this.

## How

Read value from the stream configured catalog and check the include_files value, if false/none use NoopWriter, if true, it is fine to use the FileSystemWriter.

This only happens for read operations.

### DIFF
--- a/airbyte_cdk/sources/declarative/manifest_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/manifest_declarative_source.py
@@ -179,14 +179,16 @@ class ManifestDeclarativeSource(DeclarativeSource):
     def _get_include_files(
         stream_config: Dict[str, Any],
         catalog_with_streams_name: Mapping[str, ConfiguredAirbyteStream] | None,
-    ) -> Optional[bool]:
+    ) -> bool:
         """
         Returns the include_files for the stream if it exists in the catalog.
         """
         if catalog_with_streams_name:
             stream_name = stream_config.get("name")
-            configured_catalog_stream = catalog_with_streams_name.get(stream_name)
-            return configured_catalog_stream and configured_catalog_stream.include_files
+            configured_catalog_stream = (
+                catalog_with_streams_name.get(stream_name) if stream_name else None
+            )
+            return bool(configured_catalog_stream and configured_catalog_stream.include_files)
         return False
 
     @staticmethod
--- a/airbyte_cdk/sources/declarative/manifest_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/manifest_declarative_source.py
@@ -20,6 +20,7 @@ from airbyte_cdk.models import (
     AirbyteMessage,
     AirbyteStateMessage,
     ConfiguredAirbyteCatalog,
+    ConfiguredAirbyteStream,
     ConnectorSpecification,
     FailureType,
 )
@@ -141,7 +142,7 @@ class ManifestDeclarativeSource(DeclarativeSource):
                 f"Expected to generate a ConnectionChecker component, but received {check_stream.__class__}"
             )
 
-    def streams(self, config: Mapping[str, Any]) -> List[Stream]:
+    def streams(self, config: Mapping[str, Any], catalog: Optional[ConfiguredAirbyteCatalog] = None) -> List[Stream]:
         self._emit_manifest_debug_message(
             extra_args={"source_name": self.name, "parsed_config": json.dumps(self._source_config)}
         )
@@ -154,6 +155,7 @@ class ManifestDeclarativeSource(DeclarativeSource):
         if api_budget_model:
             self._constructor.set_api_budget(api_budget_model, config)
 
+        catalog_with_streams_name = self._catalog_with_streams_name(catalog)
         source_streams = [
             self._constructor.create_component(
                 StateDelegatingStreamModel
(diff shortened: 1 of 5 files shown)


airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -229,7 +230,9 @@ def connection_checker(self) -> ConnectionChecker:
                 f"Expected to generate a ConnectionChecker component, but received {check_stream.__class__}"
             )
 
-    def streams(self, config: Mapping[str, Any]) -> List[Stream]:
+    def streams(
+        self, config: Mapping[str, Any], catalog: Optional[ConfiguredAirbyteCatalog] = None
```
maxi297, line 234: I'm a bit ambivalent on modifying this interface because now, it does not align with [the parent interface](https://github.com/airbytehq/airbyte-python-cdk/blob/4a96a2adede7b5f27c2585399e223f4cf0344bf5/airbyte_cdk/sources/abstract_source.py#L73-L79) so it seems like we are breaking compatibility and it'll be hard to maintain with doing `isinstance` before calling which means that we will lose most of the benefits of inheritance.

Could we pass the catalog at `ModelToComponentFactory.__init__` as a mandatory field? It seems like we are already passing the information [here](https://github.com/airbytehq/airbyte-python-cdk/blob/4a96a2adede7b5f27c2585399e223f4cf0344bf5/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L70). This would avoid the interface problem, the problem of managing an optional catalog and the problem of passing `include_files` through kwargs. We might want to do it [here](https://github.com/airbytehq/airbyte-python-cdk/blob/4a96a2adede7b5f27c2585399e223f4cf0344bf5/airbyte_cdk/sources/declarative/manifest_declarative_source.py#L118-L121) too if it's not too much trouble.

maxi297 (commented):
Functionally, it seems good. I have one design concern. Let me know what you think

airbyte_cdk/sources/declarative/manifest_declarative_source.py
aldogonzalez8: I think that makes sense in order to keep the signature of the interfaces.

aldogonzalez8: I have applied changes to pass it through ModelToComponentFactory.__init__

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -583,6 +583,7 @@ def __init__(
         disable_retries: bool = False,
         disable_cache: bool = False,
         disable_resumable_full_refresh: bool = False,
+        catalog: ConfiguredAirbyteCatalog = ConfiguredAirbyteCatalog(streams=[]),
```
maxi297, line 586: Do we need the default value here? Is there usage of `ModelToComponent.__init__` that are public facing?

maxi297 (approved):
LGTM! Just a small nit

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: I think we don't, I set it as None.

maxi297: Or should we make in mandatory?

aldogonzalez8: When you say mandatory you mean something like:

```
class ModelToComponentFactory:
    EPOCH_DATETIME_FORMAT = "%s"

    def __init__(
        self,
        catalog: ConfiguredAirbyteCatalog, 
        ...
```

?

I think yes is possible, it will  require us to update in different places like [here](https://github.com/airbytehq/airbyte-python-cdk/blob/4d04ed5e1b6da700ed79eaf396a61240d82edee9/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3483C9-L3483C52) for which probably we just need to pass a self.catalog that we keep.

Also, some need to update unit test that do things like `something = ModelToComponentFactory()`, 

What are the advantages we get from doing it mandatory? Or have we observed downsides of the default approach we've doing for other fields?

maxi297: Whenever we reference `catalog` internally, we know it is available and properly populated so easier code to maintain and less cognitive load for the devs.

I agree unit tests are annoying to update. How widespread would be the impact?

aldogonzalez8: less cognitive load for the devs.

aldogonzalez8: > less cognitive load for the devs.

Yes, we get confused with large context windows.

> How widespread would be the impact?

Ok, probably a few places for UT
 
image

aldogonzalez8: For what I see we would also need an update [here](https://github.com/airbytehq/airbyte-python-cdk/blob/4d04ed5e1b6da700ed79eaf396a61240d82edee9/airbyte_cdk/connector_builder/connector_builder_handler.py#L78-L93) that is called [here](https://github.com/airbytehq/airbyte-python-cdk/blob/4d04ed5e1b6da700ed79eaf396a61240d82edee9/airbyte_cdk/connector_builder/main.py#L91-L99) where the catalog from `get_config_and_catalog_from_args` is optional.

The above has its own test, which will increase the impact.

maxi297: Ok! As you wish. If you think this is worth, we can fix the test and move forward. Else, I'm fine with the current solution

PR #515, opened 2025-04-28
fix(low-code): set transform_before_filtering from manifest if client_side_incremental_sync=true

darynaishchenko:
For client side incremental sync use `transform_before_filtering` from manifest, default True if not provided 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of the `transform_before_filtering` setting to ensure consistent behavior based on configuration without altering internal state.

- **Tests**
  - Added tests to verify the correct behavior of the `transform_before_filtering` option with client-side incremental syncs.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2827,13 +2827,9 @@ class ModelToComponentFactory:
             else None
         )
 
-        if model.transform_before_filtering is None:
-            # default to False if not set
-            model.transform_before_filtering = False
-
-        assert model.transform_before_filtering is not None  # for mypy
-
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2827,13 +2827,9 @@ def create_record_selector(
             else None
         )
 
-        if model.transform_before_filtering is None:
-            # default to False if not set
-            model.transform_before_filtering = False
-
-        assert model.transform_before_filtering is not None  # for mypy
```
brianjlai, line 2834: This whole structure of an assertion in the existing code is very weird. nice to get rid of an unnecessary assert statement

brianjlai (approved):
🚢

tolik0 (approved):

PR #522, opened 2025-04-30
feat(cli): add support for building images for java connectors

aaronsteers:
This add the ability to build Java connector images.

```bash
# Test building from the dev version of the CDK:
poetry run airbyte-cdk image build --connector-directory=../airbyte/airbyte-integrations/connectors/source-mysql

# Future syntax with the prefix (after release):
cd /airbyte/airbyte-integrations/connectors/source-mysql
airbyte-cdk image build

# Prove the image works:
docker run --rm airbyte/source-mysql:dev spec
```

This _does not_ fully harden the image nor does it test every scenario needed to allow this to (yet) replace the `airbyte-ci` build and publish process. In future iterations, we would hopefully gain increasing confidence over time until we're ready to turn off the current Dagger-based processes.

Combined with this PR, we'd be able to run FAST tests on _any_ connector via docker, regardless of language:

- https://github.com/airbytehq/airbyte-python-cdk/pull/524

## Summary by CodeRabbit

- **New Features**
	- Added support for building and running Java connectors within the Docker build process.
- **Improvements**
	- Enhanced verification of connector images by improving JSON output parsing for increased robustness.
	- Updated Dockerfile and ignore rules to optimize Java connector packaging.

### DIFF
--- a/airbyte_cdk/utils/docker_image_templates.py
+++ b/airbyte_cdk/utils/docker_image_templates.py
@@ -107,7 +107,6 @@ ENTRYPOINT ["python", "./main.py"]
 
 JAVA_CONNECTOR_DOCKERFILE_TEMPLATE = """
 # Java connector Dockerfile for Airbyte
-# This Dockerfile replicates the dagger-based build process for Java connectors
 
 # Build arguments
 ARG BASE_IMAGE
--- a/airbyte_cdk/utils/docker_image_templates.py
+++ b/airbyte_cdk/utils/docker_image_templates.py
@@ -14,8 +14,9 @@ These templates are used to generate connector images.
 DOCKERIGNORE_TEMPLATE: str = "\n".join(
     [
         "# This file is auto-generated. Do not edit.",
-        # "*,"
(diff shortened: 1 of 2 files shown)


airbyte_cdk/utils/docker_image_templates.py
```diff
@@ -99,3 +100,38 @@
...
 ENTRYPOINT ["python", "./main.py"]
 """
+
+#########################
+# JAVA CONNECTOR IMAGE ##
+#########################
+
+JAVA_CONNECTOR_DOCKERFILE_TEMPLATE = """
+# Java connector Dockerfile for Airbyte
+# This Dockerfile replicates the dagger-based build process for Java connectors
```
aaronsteers, line 110:
```suggestion
```

edgao (approved):
seems legit! I didn't review the dockerfile, but the python diff made sense. (I wouldn't be surprised if the dockerignore stuff causes some random weird connector to blow up, but we can deal that if/when it happens)

probably worth also tagging someone from db sources as fyi

PR #527, opened 2025-05-01
[WIP] destination discover PoC

maxi297:


airbyte_cdk/test/mock_http/request.py
```diff
@@ -72,7 +72,11 @@ def _to_mapping(
         elif isinstance(body, bytes):
             return json.loads(body.decode())  # type: ignore  # assumes return type of Mapping[str, Any]
         elif isinstance(body, str):
-            return json.loads(body)  # type: ignore  # assumes return type of Mapping[str, Any]
+            try:
```
maxi297, line 75: Without this addition, depending on the order of evaluation, test `test_given_on_match_is_mapping_but_not_input_when_matches_then_return_false` would fail

airbyte_cdk/destinations/destination.py
```diff
@@ -68,52 +135,9 @@ def _run_write(
...
-            raise Exception("No command entered. ")
-        elif cmd not in ["spec", "check", "write"]:
-            # This is technically dead code since parse_args() would fail if this was the case
-            # But it's non-obvious enough to warrant placing it here anyways
-            raise Exception(f"Unknown command entered: {cmd}")
-
-        return parsed_args
+    @staticmethod
+    def parse_args(args: List[str]) -> argparse.Namespace:
+        return parse_args(args)
```
aaronsteers, line 140: In terms of patterns, I think my preference here and proposed best best practice (eventually) is that we make this a class method on the base connector class. I would call it "launch" and then every connector class could be able to invoke itself. In theory, all sources and destinations could share the same implementation across connectors of the same type, and unless the connector needs something special (a code smell anyway), the "cls" input arg should be all you need in order to instantiate a connector using CLI args 

I don't think we need to tackle all the scope here in this PR, but your implementation is already very close to what I think is the ideal state, so I'll mention it here as a non-blocking proposal.

maxi297: That makes sense to me. I started diverging a bit from the current solution because the `run_cmd` it is not static so we need to instantiate the object before calling it but the way we instantiate the destination depends on how the method of the protocal that is called (see [this piece of code](https://github.com/airbytehq/airbyte-enterprise/blob/master/airbyte-integrations/connectors/destination-cobra/main.py#L64-L68) for an example). 

So do we agree that the `launch` method would be static? If so, I think we are moving the right direction. I can start adding this in later changes. If not, I'll need more information about what you had in mind.

aaronsteers: Different b/w static method and class method for this use case is just the a class method knows what class it is, and the implementation can be inherited by subclasses. So, yes, agreed method is static in a sense that it doesn't need the object, but in order to not need to implement on each class, I think making a class method is slightly cleaner in the long run. Sorry if I'm not good at explaining this, and please don't block on my comment - either way, it's a step in the right direction, I think, if the class knows how to instantiate itself, without needing an external class or function in order to be instantiated. 👍

airbyte_cdk/sources/declarative/schema/dynamic_schema_loader.py
```diff
@@ -149,22 +182,26 @@ def get_json_schema(self) -> Mapping[str, Any]:
...
+            )
             properties[key] = value
 
-        transformed_properties = self._transform(properties, {})
+        transformed_properties = self._transform(properties)
 
         return {
             "$schema": "https://json-schema.org/draft-07/schema#",
             "type": "object",
-            "additionalProperties": True,
```
aaronsteers, line 159: 💎 Nice. Appreciate that we're following JSON Schema standards for communicating this. 👍

airbyte_cdk/destinations/destination.py
aaronsteers: @maxi297 -  Here's the early implementation I wrote for S3 a while back...

```
    @classmethod
    def launch(cls, args: list[str] | None = None) -> None:
        """Launch the source using the provided CLI args.

        If no args are provided, the launch args will be inferred automatically.

        In the future, we should consider moving this method to the Connector base class,
        so that all sources and destinations can launch themselves and so none of this
        code needs to live in the connector itself.
        """
        args = args or sys.argv[1:]
        catalog_path = AirbyteEntrypoint.extract_catalog(args)
        # TODO: Delete if not needed:
        # config_path = AirbyteEntrypoint.extract_config(args)
        # state_path = AirbyteEntrypoint.extract_state(args)

        source = cls.create(
            configured_catalog_path=Path(catalog_path) if catalog_path else None,
        )
        # The following function will wrap the execution in proper error handling.
        # Failures prior to here may not emit proper Airbyte TRACE or CONNECTION_STATUS messages.
        launch(
            source=source,
            args=args,
        )
```

This wasn't a very _good_ implementation, but you can see the generic `cls.create()` ref.

Where `cls.create()` was defined as:

```
    @classmethod
    def create(
        cls,
        *,
        configured_catalog_path: Path | str | None = None,
    ) -> SourceS3:
        """Create a new instance of the source.
        ...
        # A bunch of hacky stuff here.
        ...
        return cls(
            # These are the defaults for the source. No need for a caller to change them:
            stream_reader=SourceS3StreamReader(),
            spec_class=Config,
            cursor_cls=Cursor,
            # This is needed early. (We also will provide it again later.)
            catalog=configured_catalog,
            # These will be provided later, after we have wrapped proper error handling.
            config=None,
            state=None,
        )

```

aaronsteers: Another problem with our current design is that we ask for config, catalog, etc in the constructor, but then pass it again during invocation of 'check', 'discover', etc. I think it would be better to not need these in the constructor at all, in which case we greatly simplify the process of creating a connector class, and we put all code that can fail in a code path that can properly message about any failures.

Again - hopefully this is helpful for long-term thinking, but I don't think it needs to block the current PR.

aaronsteers (approved):
@maxi297 - I have reviewed all the code here and it looks good. 👍 

Thanks for responding to my comments - and again, nothing blocking there.

What I didn't find was a definition of `DestinationSyncMode` or any changes to sync modes. I did want to review those - can you point me to them?

airbyte_cdk/destinations/destination.py
maxi297: classmethod makes sense, yes!

And I would prefer to have these as part of the constructor personally as if we don't, each method will need to instantiate the object it needs every time is called or have some kind on intelligent accessor which would check if the field has already been instantiated and if not instantiate it. It feels complex instead of just instantiating the connector properly in `launch`. WDYT?

PR #530, opened 2025-05-02
feat(cli): use central dockerfile definitions, add capability to find airbyte-repo-root

aaronsteers:
This PR does a few things:

1. Add the ability to locate the airbyte repo root.
   - This works when cwd is anywhere inside the airbyte repo.
   - This also works when cwd is inside the `airbyte-enterprise` repo.
    - This also works in any repo checked out that is adjacent (sibling) to the airbyte repo. For instance, this works within the airbyte-python-cdk repo if `airbyte-python-cdk` and `airbyte` are sibling directories.
2. Modify the CLI so that `--connector-name` and `--connector-directory` can be ommitted from the first CLI arg.
   - The CLI will infer if a connector name or directory is passed, based on the presence of a path delimiter (normally "/", or "\" on windows).
   - As before, if no arg is set, we'll try to get connector dir and name from the current cdw. (Let's you run directly within the connector directory with no args.)
   - Now this works from the CDK repo: `poetry run airbyte-cdk image build source-s3`. The CDK CLI will find the airbyte repo and find the connector directory automatically.


dbgold17 (approved):
Approving, with a discussion question that we can always figure out later.

Did you consider having the Dockerfiles live within the CDK and imported into the `airbyte` repo as needed instead of now having all of this logic to find the airbyte repo locally? The file path traversing and directory structure logic feels a bit magic / fragile, but is is also nice if it just works. Another alternative that I don't love is that we could have the cdk cli by default pull from Github unless the user specifies a path to the airbyte repo locally.

airbyte_cdk/cli/airbyte_cdk/_connector.py
```diff
@@ -115,15 +114,17 @@ def connector_cli_group() -> None:
     help="Only collect tests, do not run them.",
 )
 def test(
-    connector_name: str | None = None,
-    connector_directory: Path | None = None,
+    connector: str | Path | None = None,
```
dbgold17, line 117: nit: why is connector defined here as `str | Path | None` and below as just `str | None`?

airbyte_cdk/cli/airbyte_cdk/_connector.py
```diff
@@ -115,15 +114,17 @@ def connector_cli_group() -> None:
...
     *,
     collect_only: bool = False,
 ) -> None:
     """Run connector tests.
 
     This command runs the standard connector tests for a specific connector.
 
+    [CONNECTOR] can be a connector name (e.g. 'source-pokeapi'), a path to a connector directory, or omitted to use the current working directory.
```
dbgold17, line 125: I'm not super familiar with click, but is there a way to define the connector argument once and reuse it in multiple commands?

aaronsteers: @dbgold17 - The issue here is that Click doesn't allow us to add help text for positional args. So, we have to repeat the text in the function docstring, which will render as `--help` text for the command.

airbyte_cdk/cli/airbyte_cdk/_connector.py
aaronsteers: This is because we are collecting a positional `CONNECTOR` arg from the user input something that could be a connector name, path, or nothing (omitted). And below, we detect the type and assign to stronger-typed variables. (`connector_name` will strictly be a typed as a string, for instance, and `connector_directory` will be typed as a path.)

aaronsteers (commented):
@dbgold17 - Thanks again for your review here. Responded inline to your questions and we can keep these discussions going after the merge.

Regarding your question about where to house the Dockerfiles, I feel more strongly that putting them in the main repo is at least a _good_ path, and as discussed the past couple days, it allows all teams to have a sense of ownership in those build processes. 

I think it's okay for the CDK CLI to get files from that repo - and I don't mind much if we get from local refs vs defaulting to downloading from github. The only reason I didn't default to downloading from github in this iteration is that I wanted our default mode to not need network access. So, preferring local files, then falling back to a remote download seemed like the best first implementation for now, at least.

Thanks again for all your feedback and the very thoughtful comments!

PR #531, opened 2025-05-02
tests: handle case of `acceptance-test-config.yml` with no valid test cases

aaronsteers:
This PR handles the case of missing config in `acceptance-test-config.yml`. Specifically, it now runs successfully on `source-7shift`, which has the following definition:

```yml
# See [Connector Acceptance Tests](https://docs.airbyte.com/connector-development/testing-connectors/connector-acceptance-tests-reference)
# for more information about how to configure these tests
connector_image: airbyte/source-7shifts:dev
acceptance_tests:
  spec:
    tests:
      - spec_path: "manifest.yaml"
  connection:
    bypass_reason: "This is a builder contribution, and we do not have secrets at this time"
  discovery:
    bypass_reason: "This is a builder contribution, and we do not have secrets at this time"
  basic_read:
    bypass_reason: "This is a builder contribution, and we do not have secrets at this time"
  incremental:
    bypass_reason: "This is a builder contribution, and we do not have secrets at this time"
  full_refresh:
    bypass_reason: "This is a builder contribution, and we do not have secrets at this time"
```

This test config file is especially useless because `spec` is not the same as `manifest.yml` and so it's not really helpful at all in defining test cases.

Nevertheless, we test what we can and the tests succeed now:

image

## Summary by CodeRabbit

- **Tests**
  - Improved handling of missing test scenarios in acceptance test configuration, returning an empty list instead of raising an error when no tests are found for a category.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new standard test for validating connector specification output.
  - Added a property to easily access specification messages in test results.

- **Bug Fixes**
  - Improved handling of test scenarios to support cases where no scenario is provided.
  - Enhanced test scenario aggregation to include multiple categories and handle missing data more gracefully.
  - Expanded test job support to include the "spec" verb and optional test scenarios.

### DIFF
--- a/airbyte_cdk/test/standard_tests/_job_runner.py
+++ b/airbyte_cdk/test/standard_tests/_job_runner.py
@@ -56,9 +56,9 @@ class IConnector(Protocol):
 
 def run_test_job(
     connector: IConnector | type[IConnector] | Callable[[], IConnector],
-    verb: Literal["read", "check", "discover"],
-    test_scenario: ConnectorTestScenario,
+    verb: Literal["spec", "read", "check", "discover"],
     *,
+    test_scenario: ConnectorTestScenario | None = None,
     catalog: ConfiguredAirbyteCatalog | dict[str, Any] | None = None,
 ) -> entrypoint_wrapper.EntrypointOutput:
     """Run a test scenario from provided CLI args and return the result."""
@@ -81,9 +81,9 @@ def run_test_job(
         )
 
     args: list[str] = [verb]
-    if test_scenario.config_path:
+    if test_scenario and test_scenario.config_path:
         args += ["--config", str(test_scenario.config_path)]
-    elif test_scenario.config_dict:
+    elif test_scenario and test_scenario.config_dict:
         config_path = (
             Path(tempfile.gettempdir()) / "airbyte-test" / f"temp_config_{uuid.uuid4().hex}.json"
         )
@@ -103,7 +103,7 @@ def run_test_job(
             )
             catalog_path.parent.mkdir(parents=True, exist_ok=True)
             catalog_path.write_text(orjson.dumps(catalog).decode())
-        elif test_scenario.configured_catalog_path:
+        elif test_scenario and test_scenario.configured_catalog_path:
             catalog_path = Path(test_scenario.configured_catalog_path)
 
         if catalog_path:
@@ -112,12 +112,18 @@ def run_test_job(
     # This is a bit of a hack because the source needs the catalog early.
     # Because it *also* can fail, we have to redundantly wrap it in a try/except block.
 
+    expect_exception = False
+    if test_scenario and test_scenario.expect_exception:
+        # If the test scenario expects an exception, we need to set the
+        # `expect_exception` flag to True.
+        expect_exception = True
+
     result: entrypoint_wrapper.EntrypointOutput = entrypoint_wrapper._run_command(  # noqa: SLF001  # Non-public API
         source=connector_obj,  # type: ignore [arg-type]
         args=args,
-        expecting_exception=test_scenario.expect_exception,
+        expecting_exception=False if not test_scenario else expect_exception,
     )
-    if result.errors and not test_scenario.expect_exception:
+    if result.errors and not expect_exception:
         raise AssertionError(
             f"Expected no errors but got {len(result.errors)}: \n" + _errors_to_str(result)
         )
@@ -132,7 +138,7 @@ def run_test_job(
             + "\n".join([str(msg) for msg in result.connection_status_messages])
             + _errors_to_str(result)
         )
-        if test_scenario.expect_exception:
+        if expect_exception:
             conn_status = result.connection_status_messages[0].connectionStatus
             assert conn_status, (
                 "Expected CONNECTION_STATUS message to be present. Got: \n"
@@ -146,7 +152,7 @@ def run_test_job(
         return result
 
     # For all other verbs, we assert check that an exception is raised (or not).
-    if test_scenario.expect_exception:
+    if expect_exception:
         if not result.errors:
             raise AssertionError("Expected exception but got none.")
 
--- a/airbyte_cdk/test/standard_tests/_job_runner.py
+++ b/airbyte_cdk/test/standard_tests/_job_runner.py
@@ -62,6 +62,9 @@ def run_test_job(
     catalog: ConfiguredAirbyteCatalog | dict[str, Any] | None = None,
 ) -> entrypoint_wrapper.EntrypointOutput:
     """Run a test scenario from provided CLI args and return the result."""
+    # Use default (empty) scenario if not provided:
+    test_scenario = test_scenario or ConnectorTestScenario()
+
     if not connector:
         raise ValueError("Connector is required")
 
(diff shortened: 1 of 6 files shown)


airbyte_cdk/test/standard_tests/_job_runner.py
```diff
@@ -56,9 +56,9 @@ def spec(self, logger: logging.Logger) -> Any:
 
 def run_test_job(
     connector: IConnector | type[IConnector] | Callable[[], IConnector],
-    verb: Literal["read", "check", "discover"],
-    test_scenario: ConnectorTestScenario,
+    verb: Literal["spec", "read", "check", "discover"],
     *,
+    test_scenario: ConnectorTestScenario | None = None,
```
aaronsteers, line 61: This now handles a "None" scenario - especially for the `spec` use case. If test_scenario is omitted, we presumably won't have a `config` but this is not needed for `spec` anyway, and there may be additional cases where it is not needed which we haven't considered as of yet.

airbyte_cdk/test/standard_tests/_job_runner.py
```diff
@@ -81,9 +81,9 @@ def run_test_job(
         )
 
     args: list[str] = [verb]
-    if test_scenario.config_path:
+    if test_scenario and test_scenario.config_path:
         args += ["--config", str(test_scenario.config_path)]
-    elif test_scenario.config_dict:
+    elif test_scenario and test_scenario.config_dict:
```
aaronsteers, line 86: Now we check if scenario exists before pulling 'config' from it.

airbyte_cdk/test/standard_tests/_job_runner.py
```diff
@@ -112,12 +112,18 @@ def run_test_job(
...
+    if test_scenario and test_scenario.expect_exception:
+        # If the test scenario expects an exception, we need to set the
+        # `expect_exception` flag to True.
+        expect_exception = True
+
     result: entrypoint_wrapper.EntrypointOutput = entrypoint_wrapper._run_command(  # noqa: SLF001  # Non-public API
         source=connector_obj,  # type: ignore [arg-type]
         args=args,
-        expecting_exception=test_scenario.expect_exception,
+        expecting_exception=False if not test_scenario else expect_exception,
```
aaronsteers, line 124: I forgot to update this after adding the dedicated variable. Fixes:

```suggestion
        expecting_exception=expect_exception,
```

airbyte_cdk/test/standard_tests/source_base.py
```diff
@@ -64,6 +64,26 @@ def test_discover(
             test_scenario=scenario,
         )
 
+    def test_spec(
+        self,
+        # scenario: ConnectorTestScenario,  # No inputs needed for spec test
+    ) -> None:
+        """Standard test for `spec`."""
```
aaronsteers, line 71:
```suggestion
    def test_spec(
        self,
        # scenario: ConnectorTestScenario,  # No inputs needed for spec test
    ) -> None:
        """Standard test for `spec`.

        This test does not require a `scenario` input, since `spec`
        does not require any inputs.
        
        We assume `spec` should always succeed and it should always generate 
        a valid `SPEC` message.

        Note: the parsing of messages by type also implicitly validates that
        the generated `SPEC` message is valid JSON.
        """
```

airbyte_cdk/test/standard_tests/declarative_sources.py
```diff
@@ -64,7 +64,7 @@ def components_py_path(cls) -> Path | None:
     @classmethod
     def create_connector(
         cls,
-        scenario: ConnectorTestScenario,
+        scenario: ConnectorTestScenario | None,
```
aaronsteers, line 67: The scenario can now accept `None` but it does not default to `None`. For now at least, this prevents us from forgetting to provide `scenario`, while still allowing us to do so explicitly when needed.

airbyte_cdk/test/standard_tests/_job_runner.py
```diff
@@ -112,12 +112,18 @@ def run_test_job(
     # This is a bit of a hack because the source needs the catalog early.
     # Because it *also* can fail, we have to redundantly wrap it in a try/except block.
 
+    expect_exception = False
+    if test_scenario and test_scenario.expect_exception:
```
dbgold17, line 116: could probably restructure this logic a bit so you aren't repeatedly checking if the scenario isn't `None`

airbyte_cdk/test/standard_tests/connector_base.py
```diff
@@ -89,7 +89,7 @@ def get_test_class_dir(cls) -> Path:
     @classmethod
     def create_connector(
         cls,
-        scenario: ConnectorTestScenario,
+        scenario: ConnectorTestScenario | None,
```
dbgold17, line 92: why do we need a scenario at all for this method?

airbyte_cdk/test/standard_tests/source_base.py
```diff
@@ -64,6 +64,33 @@ def test_discover(
...
+
+        Note: the parsing of messages by type also implicitly validates that
+        the generated `SPEC` message is valid JSON.
+        """
+        result = run_test_job(
+            verb="spec",
+            test_scenario=None,
+            connector=self.create_connector(scenario=None),
+        )
+        if result.errors:
```
dbgold17, line 84: This is outside the scope of this PR, but I think I need to better understand the structure of the standard test code. At first glance it seems somewhat duplicative to be checking for errors here but also have some logic in the `run_test_job` method to check for whether exceptions were thrown

dbgold17 (approved):
approving with some comments

airbyte_cdk/test/standard_tests/connector_base.py
aaronsteers: Great question. In the long run, I would like to not need it. But for legacy reasons, many (_most?_) connectors cannot properly initialize themselves without having the config, catalog, and/or state inputs. Ideally, these inputs would only be passed when invoking the verbs (`read()`, `discover()`, etc.) but for now, they are part of the class constructor - and connectors (esp. S3, which I have experience with) may act unpredictably if the inputs are not passed to the constructor.

I have another thread on this topic from earlier today with @maxi297 here:

- https://github.com/airbytehq/airbyte-python-cdk/pull/527#discussion_r2075956954

airbyte_cdk/test/standard_tests/_job_runner.py
aaronsteers: @dbgold17 - Per your feedback, I've updated the implementation so that scenarios can now have an empty config. This reduces the not-null checks we otherwise would have to do in this implementation.

airbyte_cdk/test/standard_tests/source_base.py
aaronsteers: You are right to raise it. In fact, as long as the `run_test_job` doesn't get a scenario that expects failure, it will raise the error and message about it for us. So, I've updated the implementation to simply have a code-comment that the hard-fail scenario is already handled. Then, all that is left to do is check that we have 1 and exactly 1 `SPEC` output.

Here is the part in `run_test_job` which does the check already:

- https://github.com/airbytehq/airbyte-python-cdk/blob/72e73ecc3e04aae6513c47ef5f6f56675266b803/airbyte_cdk/test/standard_tests/_job_runner.py#L123-L127

airbyte_cdk/test/standard_tests/models/scenario.py
```diff
@@ -13,6 +13,7 @@
 from typing import Any, Literal, cast
 
 import yaml
+from numpy import empty
```
aaronsteers, line 16:
```suggestion
```

PR #533, opened 2025-05-06
feat: implement validators, validation strategies, and RemapField transformer

pnilan:
## What
- Implements runtime components to support migrating config migrations to low-code
- Closes #https://github.com/airbytehq/airbyte-internal-issues/issues/12696

## How
- New `Validator` base class
- New `DpathValidator` and `PredicateValidator` validator classes
- New `ValidationStrategy` base class
- New `ValidateAdheresToSchema` validation strategy class
- New `ConfigTransformation` base class
- New `RemapField` config transformation class

## Recommended Review Order
- `validator.py`
- `dpath_validator.py` and unit tests
- `predicate_validator.py` and unit tests
- `validation_strategy.py`
- `validate_adheres_to_schema` and unit tests
- `config_transformation.py`
- `remap_field.py` and unit tests

## Summary by CodeRabbit

- **New Features**
  - Introduced transformations for configuration dictionaries: adding fields with nested and conditional support, removing fields with conditional logic and list-aware handling, and remapping field values based on static mappings.
  - Added a suite of validators for configuration data, including path-based extraction validation, predicate validation, and JSON schema adherence validation strategies.
- **Tests**
  - Added comprehensive unit tests for all new transformation and validation features, covering typical, edge, and error cases to ensure robustness and correctness.


airbyte_cdk/sources/declarative/validators/dpath_validator.py
```diff
@@ -0,0 +1,65 @@
...
+        if "*" in path:
+            try:
+                values = dpath.values(input_data, path)
+            except KeyError as e:
+                raise KeyError(f"Error validating path '{self.field_path}': {e}")
+            for value in values:
+                try:
+                    self.strategy.validate(value)
+                except Exception as e:
+                    raise ValueError(f"Error validating value '{value}': {e}")
```
pnilan, line 56: Maybe catching and re-raising here is unnecessary...

brianjlai (approved):
Looks good and everything looks good. We talked about adding the `ConfigTransformation` implementations for AddFields and RemoveFields. But the rest we can add as needed during migrations. And the comment about the mapping. Will approve so I don't block you, but happy to give the final look if you want to once you add the other transformations

airbyte_cdk/sources/declarative/transformations/config_transformations/config_transformation.py
```diff
@@ -0,0 +1,23 @@
...
+
+class ConfigTransformation(ABC):
+    """
+    Implementations of this class define transformations that can be applied to source configurations.
+    """
+
+    @abstractmethod
+    def transform(
+        self,
+        config: Dict[str, Any],
```
brianjlai, line 17: It makes sense that we can't use `Config` since its immutable. But for consistency, should this be a `MutableMapping` like the underlying implementations of each config transformation?

`Dict` would be fine too, but i think we tend to use `MutableMapping`

airbyte_cdk/sources/declarative/transformations/config_transformations/remap_field.py
```diff
@@ -0,0 +1,64 @@
...
+            raise Exception("field_path cannot be empty.")
+        self._field_path = [
+            InterpolatedString.create(path, parameters={}) for path in self.field_path
+        ]
+        for path_index in range(len(self.field_path)):
+            if isinstance(self.field_path[path_index], str):
+                self._field_path[path_index] = InterpolatedString.create(
+                    self.field_path[path_index], parameters={}
+                )
+        self._map = InterpolatedMapping(self.map, parameters={}).eval(config=self.config)
```
brianjlai, line 37: We should instantiate the `InterpolatedMapping` as you have done. However, we shouldn't call `self._map.eval()` until we are in the `transform()` method.

It is a subtle different, but we try to enforce a general rule of thumb that we perform all interpolation and evaluation of the template at runtime. The way you have it, we're going to resolve all the mapping values in the resulting `self._map`

For this case specifically, it doesn't make much of a difference. However, in the cases of other interpolation of `next_page_token`, `stream_slice`, `response`, etc, those must be evaluated in their respective runtime methods and we just try to do that across our entire codebase for consistency.

brianjlai (approved):
just a few small things, but this looks ready to go. nice work! ✅ 

let me know if my explanation for the runtime vs. parse time eval of the mapping makes sense too since its a somewhat subtle way that we think about interpolation processing

airbyte_cdk/sources/declarative/transformations/config_transformations/config_transformation.py
pnilan: Good call. Should be MutableMapping

airbyte_cdk/sources/declarative/transformations/config_transformations/remap_field.py
pnilan: Makes sense! Good call out.

airbyte_cdk/sources/declarative/transformations/config_transformations/remap_field.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+
+@dataclass
+class RemapField(ConfigTransformation):
+    """
+    Transformation that remaps a field's value to another value based on a static map.
+    """
+
+    map: Mapping[str, Any]
```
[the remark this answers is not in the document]
pnilan: Agreed! Updated to add InterpolatedMapping/

PR #534, opened 2025-05-07
fix(connector-builder): fix dynamic stream read with dynamic schema

lazebnyi:
## What

The Connector Builder CDK, during execution of the `read_stream` command, listens to all logs and captures all outgoing requests, serializing them as part of the read stream process. As a result, dynamic schema loader calls (e.g., for `dynamic_properties`) are currently treated the same as regular stream reads. We need to update the `log_formatter` so that calls to `dynamic_properties` are marked as auxiliary.

## How

Updated `_fetch_next_page` in `SimpleRetrieverTestReadDecorator` to set the `is_auxiliary` flag to `True` for streams named `dynamic_properties`.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to enable or disable HTTP response logging for data retrieval operations.
  - Introduced customizable formatting for HTTP response logs.

- **Refactor**
  - Improved the way log formatting is handled, allowing for greater flexibility and consistency in log output.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2382,7 +2382,6 @@ def create_dynamic_schema_loader(
             primary_key=None,
             stream_slicer=combined_slicers,
             transformations=[],
-            use_cache=True,
```
maxi297, line 2385: Why did we remove that? It seems to make sense that since the schema doesn't change over the time of the execution, we should use the cache here. Although, I don't know if we perform the same requests multiple times.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -670,6 +670,7 @@ def _fetch_next_page(
         stream_slice: StreamSlice,
         next_page_token: Optional[Mapping[str, Any]] = None,
     ) -> Optional[requests.Response]:
+        is_dynamic_schema_call = self.name == "dynamic_properties"
```
maxi297, line 673: I'm not super in love with this solution because someone could definitely change the name of the retriever without thinking it would break something in the connector builder. It also feels like the log formatter is still wrong as it would return Could we `f"Request performed in order to extract records for stream '{self.name}'"` as a description. Could we make the log_formatter a parameter of the SimpleRetriever as a nullable parameter and remove the override from `SimpleRetrieverTestReadDecorator`?

maxi297 (commented):
Questions so that I understand a bit more the context

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
lazebnyi: For some connectors, we have a large number of schemas (e.g. service-now), and we treat them as async calls, which triggers an error with SQLite and prevents fetching the session. This change allows us to disable/enable caching when needed.

maxi297: Can you share the errors here for documentation purposes?

lazebnyi: Hmm... actually, I can't reproduce it now.

Rolling back the change.

maxi297 (commented):
Do we need to instantiate the log_formatter for the DynamicSchemaResolver in order to pass the right format?

lazebnyi:
> Do we need to instantiate the log_formatter for the DynamicSchemaResolver in order to pass the right format?
I'm not sure if we use this information anywhere, but I'm okay with adding it.

maxi297 (approved):
LGTM!

PR #539, opened 2025-05-09
fix(cdk): fix state emit for ConcurrentPerPartitionCursor

lazebnyi:
## What
If the parent stream of a per-partition stream commits states, we may emit the state more than once at the end of the sync for a global cursor.

## How
Added a condition in `ConcurrentPerPartitionCursor` to avoid emitting the state when the global cursor is enabled, there's no `parent_stream` in the state, and throttling is applied.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved incremental sync behavior to prevent unnecessary state message emissions when using a global cursor with no parent state and throttling enabled.

- **Tests**
  - Added new test scenarios to validate incremental sync with a global cursor and no incremental dependency.
  - Enhanced test utilities to verify the number of emitted state messages during sync.

- **Style**
  - Removed an extraneous blank line to improve code readability.

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -242,6 +242,10 @@ class ConcurrentPerPartitionCursor(Cursor):
             if current_time is None:
                 return
             self._last_emission_time = current_time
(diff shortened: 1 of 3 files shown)


maxi297 (approved):
LGTM! :shipit:

PR #543, opened 2025-05-13
feat(low-code): pass top level params to PropertiesFromEndpoint requester

darynaishchenko:
## Summary by CodeRabbit

- **Refactor**
  - Enhanced parameter propagation to support nested components without explicit types, improving handling of complex nested structures.
  - Extended recognition of JSON schema object types for more accurate processing.

- **Tests**
  - Added a test to verify correct deep propagation of parameters across nested declarative components, including property chunking configurations.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
+++ b/airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
@@ -120,14 +120,6 @@ class ManifestComponentTransformer:
             if found_type:
                 propagated_component["type"] = found_type
 
-        # When there is no resolved type, we're not processing a component (likely a regular object) and don't need to propagate parameters
-        # When the type refers to a json schema, we're not processing a component as well. This check is currently imperfect as there could
-        # be json_schema are not objects but we believe this is not likely in our case because:
-        # * records are Mapping so objects hence SchemaLoader root should be an object
-        # * connection_specification is a Mapping
-        if "type" not in propagated_component or self._is_json_schema_object(propagated_component):
-            return propagated_component
-
         # Combines parameters defined at the current level with parameters from parent components. Parameters at the current
         # level take precedence
         current_parameters = dict(copy.deepcopy(parent_parameters))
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
```diff
@@ -181,4 +208,34 @@ def propagate_types_and_parameters(
 
     @staticmethod
     def _is_json_schema_object(propagated_component: Mapping[str, Any]) -> bool:
-        return propagated_component.get("type") == "object"
+        return propagated_component.get("type") == "object" or propagated_component.get("type") == [
+            "null",
+            "object",
```
darynaishchenko, line 213: this a a fix for a case when schema type is `["null", "object"]` and `$parameters` were not ignored and and were added to json schema

airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
```diff
@@ -181,4 +194,34 @@ def propagate_types_and_parameters(
...
+        ]
+
+    @staticmethod
+    def _has_nested_components(propagated_component: Mapping[str, Any]) -> bool:
+        has_nested_components = False
+        for k, v in propagated_component.items():
+            if isinstance(v, dict) and v.get("type"):
+                has_nested_components = True
```
brianjlai, line 207: This method can be simplified because once we see any dictionary that contains `type`, we can stop looping and return True immediately.

```
def _has_nested_components(propagated_component: Mapping[str, Any]) -> bool:
  for k, v in propagated_component.items():
    if isinstance(v, dict) and v.get("type"):
      return True
  return False
```

airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
```diff
@@ -181,4 +194,34 @@ def propagate_types_and_parameters(
...
+        return has_nested_components
+
+    def _process_nested_components(
+        self,
+        propagated_component: Dict[str, Any],
+        parent_field_identifier: str,
+        current_parameters: Mapping[str, Any],
+        use_parent_parameters: Optional[bool] = None,
+    ) -> Dict[str, Any]:
+        for k, v in propagated_component.items():
```
brianjlai, line 217: nit: for convention we usually try to avoid single letter variables and describe them. It's probably more descriptive to make this `field_name` and `value`.

brianjlai (approved):
a few nits and suggestions, but non-blocking and looks good to merge!

PR #544, opened 2025-05-14
fix(cdk): determine state from manager if not received a state in per partition router

aldogonzalez8:
We found that indeed it seems we have an issue where state is not passed downstream for [this flow here](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1996-L2005) when:
1. Stream uses async retriever
2. Asyc Retriever has a SubstreamPartitionRouter
3. Is incremental stream

I was able to reproduce for:

- source-bing-ads manifest [migration](https://github.com/airbytehq/airbyte/pull/59750/files#diff-e940c9c1127f2dfa6d01ef1e93b0ff7338d71d46fa225de7bd0d0aaf7cf2940aR125-R228).

   - I have a walkthrough Loom video debugging [here](https://www.loom.com/share/dbc07fbf95b947989ed363fa44685e92?sid=474b0c63-170d-4338-a572-e333eb9a9162).
   - I also added an integration test that reproduces the issue [here](https://github.com/airbytehq/airbyte/pull/59750/files#diff-1c231116365a62bb9bb5c81cc7156b7ad5d9d56b60cf875fdf24167b2135eb5fR156-R217).
- Current implementation of source-amazon-Ads, e.g., [sponsored_display_campaigns_report_stream_daily](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/manifest.yaml#L1404-L1413).
   - I have another walkthrough loom video debugging [here](https://www.loom.com/share/164ba10d5bc049dfbbb63aeff8475406?sid=5f3f6f19-c4a3-4643-a9b1-7e7b8fa242e8).
   - A sample connection [here](https://cloud.airbyte.com/workspaces/3fe0e748-014b-4f17-8c85-911da22f9b85/connections/b2ef0722-1e13-4194-9bd5-3398a7a1b428/timeline).

Downstream, the state is never passed to [ConcurrentPerPartitionCursor._set_initial_state(stream_state) ](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py#L116)then we don't have a self._cursor_per_partition [here](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py#L419-L420). There is an [alternative flow ](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py#L279-L283)when no cursor is present, which is also followed in the looms; neither seems to retrieve a state that can be used in the [ConcurrentCursor](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/streams/concurrent/cursor.py#L158).

## Summary by CodeRabbit

- **New Features**
  - Added support for asynchronous data retrieval with partitioned request handling in declarative streams.
  - Introduced incremental sync capabilities with datetime-based cursors and dynamic pagination.
  - Enabled session token authentication with automatic token refresh and API key injection.
- **Bug Fixes**
  - Improved handling of stream state to ensure a valid state is always used when creating concurrent cursors, reducing the risk of errors when no stream state is provided.
  - Enhanced stream state migration application to maintain accurate and up-to-date stream states during incremental syncs.
- **Tests**
  - Added comprehensive tests verifying asynchronous retriever behavior with partitioned cursors and incremental sync functionality.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -1485,7 +1485,6 @@ class ModelToComponentFactory:
             )
         )
 
-
         stream_state = self.apply_stream_state_migrations(stream_state_migrations, stream_state)
         # Per-partition state doesn't make sense for GroupingPartitionRouter, so force the global state
         use_global_cursor = isinstance(
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -1491,7 +1491,9 @@ class ModelToComponentFactory:
         ) or component_definition.get("global_substream_cursor", False)
 
         if not stream_state:
-            stream_state = self._connector_state_manager.get_stream_state(stream_name, stream_namespace)
+            stream_state = self._connector_state_manager.get_stream_state(
+                stream_name, stream_namespace
+            )
 
         # Return the concurrent cursor and state converter
         return ConcurrentPerPartitionCursor(
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -1484,16 +1484,17 @@ class ModelToComponentFactory:
                 stream_state_migrations=stream_state_migrations,
             )
         )
-        stream_state = self.apply_stream_state_migrations(stream_state_migrations, stream_state)
-        # Per-partition state doesn't make sense for GroupingPartitionRouter, so force the global state
-        use_global_cursor = isinstance(
-            partition_router, GroupingPartitionRouter
-        ) or component_definition.get("global_substream_cursor", False)
 
         if not stream_state:
             stream_state = self._connector_state_manager.get_stream_state(
                 stream_name, stream_namespace
             )
+
+        stream_state = self.apply_stream_state_migrations(stream_state_migrations, stream_state)
+        # Per-partition state doesn't make sense for GroupingPartitionRouter, so force the global state
+        use_global_cursor = isinstance(
+            partition_router, GroupingPartitionRouter
+        ) or component_definition.get("global_substream_cursor", False)
 
         # Return the concurrent cursor and state converter
         return ConcurrentPerPartitionCursor(
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -1485,10 +1485,6 @@ class ModelToComponentFactory:
             )
         )
 
-        if not stream_state:
-            stream_state = self._connector_state_manager.get_stream_state(
-                stream_name, stream_namespace
-            )
 
         stream_state = self.apply_stream_state_migrations(stream_state_migrations, stream_state)
         # Per-partition state doesn't make sense for GroupingPartitionRouter, so force the global state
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1484,6 +1484,12 @@ def create_concurrent_cursor_from_perpartition_cursor(
                 stream_state_migrations=stream_state_migrations,
             )
         )
+
+        if not stream_state:
+            stream_state = self._connector_state_manager.get_stream_state(
+                stream_name, stream_namespace
+            )
+
```
aldogonzalez8, line 1492: This change seems to indeed fix the problem with state for async jobs.

image

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -1484,6 +1484,12 @@ def create_concurrent_cursor_from_perpartition_cursor(
                 stream_state_migrations=stream_state_migrations,
             )
         )
+
+        if not stream_state:
```
brianjlai, line 1488: While not a blocker, why did we decide to fix this here? The alternative place that would also work would be in:
https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1996

We should have all the information needed to resolve it there as well.

I'm not that strongly opinionated on either place and both should work, but I mainly noticed that this wasn't a problem for other flows because over here, we perform the state retrieval right before we invoke `create_concurrent_cursor_from_perpartition_cursor()`
https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L380-L382

brianjlai (approved):
thanks for digging deep into this! just one thing I wanted to mention, but whichever approach we go with i'm fine. 

How easy/hard would this be to unit test?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: Makes sense, I just moved it. Let me check about the unit test.

tolik0 (approved):

unit_tests/sources/declarative/parsers/resources/stream_with_incremental_and_aync_retriever_with_partition_router.yaml
```diff
@@ -0,0 +1,140 @@
...
+    creation_requester:
+      type: HttpRequester
+      url_base: https://reporting.api.bingads.microsoft.com/
+      path: Reporting/v13/GenerateReport/Submit
+      http_method: POST
+      request_headers:
+        Content-Type: application/json
+        DeveloperToken: "{{ config['developer_token'] }}"
+        CustomerId: "'{{ stream_partition['customer_id'] }}'"
+        CustomerAccountId: "'{{ stream_partition['account_id'] }}'"
```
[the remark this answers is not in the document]
aldogonzalez8: Not a concern at the moment.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: > While not a blocker, why did we decide to fix this here? The alternative place that would also work would be in:
> https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L1996

Change the state initialization to the proposed place, and added a unit test

PR #545, opened 2025-05-15
feat(cli): accept dynamic GCP project ID from env var during secrets fetch

aaronsteers:
This does a few things:

First, this sets us up to support community-provided "BYO" creds as related to here:
   - https://github.com/airbytehq/airbyte/pull/60289

By setting an env var for `GCP_PROJECT_ID` (non-sensitive) along with `GCP_GSM_CREDENTIALS` (secret), partners and community members can enable integration tests to run in their own forks, without intervention from Airbyte and without needing Airbyte approval for their workflows to run.

Also included: this drive-by fix:
- Resolve a missed update to the `secrets list` CLI, which is to accept a CONNECTOR positional arg, rather than requiring `--connector-name` or `--connector-directory` to be set. This brings this CLI command into parity with the other recent updates to `secrets fetch`, `connector test`, and others.

## Summary by CodeRabbit

- **New Features**
  - The CLI now allows specifying the GCP project ID via an environment variable, providing more flexibility in configuration.

- **Improvements**
  - The `list` and `fetch` commands now accept a single optional positional argument to specify the connector by name or path, simplifying usage.
  - Updated help texts and documentation for CLI commands to clarify new behaviors and argument options.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte_cdk/cli/airbyte_cdk/_secrets.py
+++ b/airbyte_cdk/cli/airbyte_cdk/_secrets.py
@@ -49,10 +49,12 @@ from airbyte_cdk.utils.connector_paths import (
     resolve_connector_name_and_directory,
 )
 
-AIRBYTE_GCP_PROJECT_ID = os.environ.get(
-    "AIRBYTE_GCP_PROJECT_ID",
+GCP_PROJECT_ID: str = os.environ.get(
+    "GCP_PROJECT_ID",
     "dataline-integration-testing",
 )
+"""The GCP project ID to use for fetching integration test secrets."""
+
 CONNECTOR_LABEL = "connector"
 GLOBAL_MASK_KEYS_URL = "https://connectors.airbyte.com/files/registries/v0/specs_secrets_mask.yaml"
 
@@ -86,8 +88,11 @@ def secrets_cli_group() -> None:
 @click.option(
     "--gcp-project-id",
     type=str,
-    default=AIRBYTE_GCP_PROJECT_ID,
-    help=f"GCP project ID. Defaults to '{AIRBYTE_GCP_PROJECT_ID}'.",
+    default=GCP_PROJECT_ID,
+    help=(
+        "GCP project ID for retrieving integration tests credentials. "
+        "Defaults to the value of the `GCP_PROJECT_ID` environment variable, if set."
+    ),
 )
 @click.option(
     "--print-ci-secrets-masks",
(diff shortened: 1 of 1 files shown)


aaronsteers:
Logging here:

I successfully tested this locally using the following method.

1. `poetry run which airbyte-cdk` to get an absolute path to the CDK CLI within the Poetry venv.
2. `cd ../airbyte/airbyte-integrations/connectors/destination-duckdb` so I'm in a connector directory.
4. `.../airbyte-cdk secrets list` to confirm secrets list works as expected still after this change.
5. `export GCP_PROJECT_ID=foobar` and then re-run the above to confirm that it fails.
5. `export GCP_PROJECT_ID=dataline-integration-testing` and then re-run the above to confirm that it succeeds.

airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -49,7 +49,9 @@
     resolve_connector_name_and_directory,
 )
 
-AIRBYTE_INTERNAL_GCP_PROJECT = "dataline-integration-testing"
+GCP_PROJECT_ID: str = os.environ.get("GCP_PROJECT_ID", "") or "dataline-integration-testing"
```
dbgold17, line 52: remind me the reason why `os.environ.get("GCP_PROJECT_ID", "dataline-integration-testing")` didn't work. worth adding a comment?

dbgold17 (approved):
LGTM, one comment

airbyte_cdk/cli/airbyte_cdk/_secrets.py
aaronsteers: Yes, thanks for the suggestion. Let me write it out inline and then apply:

```suggestion
GCP_PROJECT_ID: str = os.environ.get("GCP_PROJECT_ID", "") or "dataline-integration-testing"
# We put the `or` outside the `get()` because we want the `GCP_PROJECT_ID`
# value to be ignored if it contains an empty string, such as in CI where the
# workflow might set it to a value that is itself actually missing or unset.
```

airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -49,7 +49,12 @@
...
+# We put the `or` outside the `get()` because we want the `GCP_PROJECT_ID`
+# value to be ignored if it contains an empty string, such as in CI where the
```
aaronsteers, line 54:
```suggestion
# env var to be ignored if it contains an empty string, such as in CI where the
```

PR #546, opened 2025-05-15
fix(cdk): pass transformations to concurrent cursor and fix transformation

aldogonzalez8:
## Initial scope...
Pass state transformations to concurrent cursor flow creation for async retriever with stream slicer flow.

Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/13019

## But Also!
If we try to apply a transformation to the stream_state that already represents the cursor for the partition we can end up with things like happening [here](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/migrations/legacy_to_per_partition_state_migration.py#L93-L98):

From
````
{'TimePeriod': '2025-05-12T08:00:00+00:00'}
````

To:
```

{
"states": [
      {
         'cursor': '2025-05-12T08:00:00+00:00', 
         'partition': {'account_id': 'TimePeriod'}
      }
   ]
}
```

Then I'm falling False if the internal stream_state items value is not a dict, according to the docstrings:

```
        The expected state format is
        "<parent_key_id>" : {
          "<cursor_field>" : "<cursor_value>"
        }
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for state migration to ensure only correctly formatted partitioned states are processed.

- **New Features**
  - Added support for applying state migrations to streams using async retrievers in addition to simple retrievers.
  - Enhanced configuration options for partition routing by supporting hierarchical (parent-child) stream relationships.

- **Tests**
  - Expanded test coverage for state migration logic, including new scenarios for both legacy and newest state formats, and updated error message assertions for retriever compatibility.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -928,9 +928,9 @@ class ModelToComponentFactory:
         declarative_stream: DeclarativeStreamModel,
     ) -> LegacyToPerPartitionStateMigration:
         retriever = declarative_stream.retriever
-        if not isinstance(retriever, SimpleRetrieverModel):
+        if not isinstance(retriever, (SimpleRetrieverModel, AsyncRetrieverModel)):
             raise ValueError(
-                f"LegacyToPerPartitionStateMigrations can only be applied on a DeclarativeStream with a SimpleRetriever. Got {type(retriever)}"
+                f"LegacyToPerPartitionStateMigrations can only be applied on a DeclarativeStream with a SimpleRetriever or AsyncRetriever. Got {type(retriever)}"
             )
         partition_router = retriever.partition_router
         if not isinstance(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py [hunk 1]
```diff
@@ -928,9 +928,9 @@ def create_legacy_to_per_partition_state_migration(
         declarative_stream: DeclarativeStreamModel,
     ) -> LegacyToPerPartitionStateMigration:
         retriever = declarative_stream.retriever
-        if not isinstance(retriever, SimpleRetrieverModel):
+        if not isinstance(retriever, (SimpleRetrieverModel, AsyncRetrieverModel)):
             raise ValueError(
-                f"LegacyToPerPartitionStateMigrations can only be applied on a DeclarativeStream with a SimpleRetriever. Got {type(retriever)}"
+                f"LegacyToPerPartitionStateMigrations can only be applied on a DeclarativeStream with a SimpleRetriever or AsyncRetriever. Got {type(retriever)}"
```
aldogonzalez8, line 933: I'm assuming it's because we didn't have AsyncRetriever before (?).

brianjlai: that seems likely

brianjlai (approved):
:shipit:

PR #550, opened 2025-05-17
feat(cdk): add schema filter to DynamicSchemaLoader

lazebnyi:
## What
Resolved: https://github.com/airbytehq/airbyte-internal-issues/issues/13045

In some cases, when using the dynamic schema loader endpoint, the provider may return **all possible fields** that could be present. To keep the schema up to date and ensure that the `properties` section only includes the fields that were actually requested, we need to filter the returned values accordingly.

## How
Added filtering behavior to the schema loader using the `schema_filter`.

## Summary by CodeRabbit

- **New Features**
  - Added support for filtering fields in dynamic schema loading, allowing selective inclusion of fields in the generated schema using configurable filters.

- **Documentation**
  - Updated descriptions to clarify schema loader behavior and the merging of multiple schema loaders.

- **Tests**
  - Enhanced tests to verify the new schema filtering functionality.


brianjlai (approved):
🚢

PR #551, opened 2025-05-17
feat(cdk): add CombinedExtractor

lazebnyi:
## How


brianjlai (commented):
Looking over how this extractor is used, I think this feels better suited as a custom component used by `source-google-analytics`. The main part to me that feels particularly specific to one connector is the merging of all the combined extractor records into a single record instance and emitting just one record.

Unless we have other examples where this extractor would be used by other connectors, I propose that we simplify things by defining this implementation in within `source-google-analytics-data-api`. And if we see a need to promote this back to the CDK at a later time we can. 

Since the language supports `CustomRecordExtractor` it seems like we don't have any blocker on just defining this in `components.py`

PR #552, opened 2025-05-17
feat(cdk): add KeyValueExtractor

lazebnyi:
## How


airbyte_cdk/sources/declarative/extractors/key_value_extractor.py
```diff
@@ -0,0 +1,42 @@
...
+    values_extractor: RecordExtractor
+
+    def extract_records(self, response: requests.Response) -> Iterable[MutableMapping[Any, Any]]:
+        keys = list(self.keys_extractor.extract_records(response))
+        values = self.values_extractor.extract_records(response)
+
+        while True:
+            chunk = list(islice(values, len(keys)))
+            if not chunk:
+                break
```
brianjlai, line 41: It feels a little more idiomatic to avoid using a permanent `True` + break statement. Can we instead use a `has_more_chunks` variable to manage when to stop iterating.
```
has_more_chunks = True
while has_more_chunks:
  ...
  if len(chunk) == 0: # As far as I understand reading the code we break if there are no values in the current chunk
    has_more_chunks = True
  else:
    yield dict(zip(keys, chunk))
```

airbyte_cdk/sources/declarative/extractors/key_value_extractor.py
```diff
@@ -0,0 +1,42 @@
...
+      values_extractor -> yields: ["Alice", 30]
+      result: { "name": "Alice", "age": 30 }
+    """
+
+    keys_extractor: RecordExtractor
+    values_extractor: RecordExtractor
+
+    def extract_records(self, response: requests.Response) -> Iterable[MutableMapping[Any, Any]]:
+        keys = list(self.keys_extractor.extract_records(response))
+        values = self.values_extractor.extract_records(response)
```
brianjlai, line 36: What is still a little unclear to me is what the expected behavior for this should be beyond the simplest case of there are 2 field keys and 2 values in the list.

What happens in the following:

- Values has a length of 3 like `["Alice", 30, "what about this"]`? Is this an error state, because right now, this would yield a second record of `{"name": "what about this"}` without the second property
- Are we always expecting `values` to always be a single continuous list `["Alice", 30, "Thomas", 40, "Alex", 35]`? Or do we expect it to be grouped into multiple lists of lists: `[["Alice", 30], ["Thomas", 40], ["Alex", 35]]`?

unit_tests/sources/declarative/extractors/test_key_value_extractor.py
```diff
@@ -0,0 +1,137 @@
...
+                    },
+                },
+                "paginator": {"type": "NoPagination"},
+            },
+        }
+    ],
+}
+
+
+def test_key_value_extractor():
```
brianjlai, line 102: Let's add some additional scenarios beyond just the single element case to make sure this is working and demonstrate what the intended behavior should be for multiple elements.

brianjlai (commented):
Some general comments and little things. I think the main thing I want to understand better before approving is how this is supposed to work when we have multiple elements to emit depending on what `values` is.

Also, just as a general thought, my worry is that parsing over the `values` array seems pretty unique so I was not sure if there was a common pattern shared by multiple connectors and if this was more suitable as a custom component.

airbyte_cdk/sources/declarative/extractors/key_value_extractor.py
```diff
@@ -0,0 +1,42 @@
...
+    and each value from `values_extractor` is the value for that key.
+
+    Example:
+      keys_extractor -> yields: ["name", "age"]
+      values_extractor -> yields: ["Alice", 30]
+      result: { "name": "Alice", "age": 30 }
+    """
+
+    keys_extractor: RecordExtractor
+    values_extractor: RecordExtractor
```
brianjlai, line 32: for consistency, we should probably also take in the `config` and `parameter` objects since right now we are also passing them in via the factory in `model_to_component_factory.py.create_key_value_extractor()`

lazebnyi:
A custom component will be used for the migration instead of this implementation.

PR #553, opened 2025-05-17
feat(cdk): update ConfigComponentsResolver to support list of stream_config

lazebnyi:
## How


lazebnyi:
Status: MyPy fix in progress

airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
```diff
@@ -130,7 +172,21 @@ def resolve_components(
...
+                if parsed_value and not updated and resolved_component.create_or_update:
+                    dpath.new(updated_config, path, parsed_value)
 
             yield updated_config
+
+    @staticmethod
+    def _parse_yaml_if_possible(value: Any) -> Any:
+        if isinstance(value, str):
+            try:
+                import yaml
```
brianjlai, line 187: Import this at the top level instead of inline

airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
```diff
@@ -130,7 +172,21 @@ def resolve_components(
...
+
+                return yaml.safe_load(value)
+            except Exception:
+                return value
```
brianjlai, line 191: When is it not the case that this isn't parsable in yaml? It seems like in most scenario this should always be parsable and if it wasn't we throw an error.

Would it be because the spec/config might be JSON instead of yaml?

Or at best we may want to have stricter error handling against a specific error type instead of all.

airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
```diff
@@ -91,17 +94,56 @@ def __post_init__(self, parameters: Mapping[str, Any]) -> None:
...
+            result = {}
+            for config_index, (elem_index, elem) in enumerate(combo):
+                if isinstance(elem, dict):
+                    result.update(elem)
+                else:
+                    result.setdefault(f"source_config_{config_index}", (elem_index, elem))
+            return result
 
-        return stream_config
+        all_indexed_streams = list(prepare_streams())
```
brianjlai, line 145: The result of `prepare_streams()` is an iterable, but we always iterate over it to turn it back into a list. We should probably not perform this `list()` conversion unless it must be turned into a list to process on the next line.

And if we were to do that, then we may want to just make `prepare_streams()` return a list for simplicity

brianjlai (commented):
the idea of multiple configs and defaults makes sense, I just had some areas around how we handle errors and structure some of the methods and imports to be a little more idiomatic

airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
```diff
@@ -91,17 +94,56 @@ def __post_init__(self, parameters: Mapping[str, Any]) -> None:
...
-        stream_config = dpath.get(dict(self.config), path, default=[])
+    def _stream_config(self) -> List[Dict[str, Any]]:
+        """
+        Builds a list of stream configuration dictionaries.
+
+        Each resulting config represents a unique combination of configurations
+        defined by the `configs_pointer` and any provided `default_values`.
+        """
+
+        def resolve_path(pointer: List[Union["InterpolatedString", str]]) -> List[str]:
```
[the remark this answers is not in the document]
lazebnyi: updated

airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
lazebnyi: Remove prepare_streams() and move implementation to _stream_config

PR #561, opened 2025-05-20
feat: update `ManifestDeclarativeSource` to invoke declarative config migrations/transformations/validations

pnilan:
## What
- Adds oprtional `config_path` param to `ManifestDeclarativeSource`, then builds the spec component in the source's `__init__` and then migrates config, if needed.
- Also invokes the spec's config transformations
- Invokes config validations in `streams()` method
- Closes https://github.com/airbytehq/airbyte-internal-issues/issues/12698

## Summary by CodeRabbit

- **New Features**
  - Added support for automatic migration and transformation of configuration files in declarative sources when a configuration path is provided.
  - Configuration changes are now persisted and control messages are emitted when migrations occur.

- **Bug Fixes**
  - Improved validation of configuration files before creating streams, ensuring errors are caught early.

- **Tests**
  - Introduced comprehensive tests for configuration migration, transformation, and validation behaviors in declarative sources.
  - Removed outdated migration-related tests from spec tests.

- **Chores**
  - Updated constructors for several classes to accept an optional configuration file path for enhanced flexibility.


airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -139,6 +144,35 @@ def __init__(
...
+                    if mutable_config != config:
+                        with open(config_path, "w") as f:
+                            json.dump(mutable_config, f)
+                        self.message_repository.emit_message(
+                            create_connector_config_control_message(mutable_config)
+                        )
+                        # We have no mechanism for consuming the queue, so we print the messages to stdout
+                        for message in self.message_repository.consume_queue():
+                            print(orjson.dumps(AirbyteMessageSerializer.dump(message)).decode())
+                except Exception as e:
```
brianjlai, line 168: I'm curious what types of errors we get here. I presume we can always have a chance of file not existing among others, but usually it's better to only catch the ones we expect and handle accordingly so that any unexpected errors are properly made loud

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -255,6 +289,9 @@ def connection_checker(self) -> ConnectionChecker:
             )
 
     def streams(self, config: Mapping[str, Any]) -> List[Stream]:
+        if self._spec_component:
+            self._spec_component.validate_config(config)
```
brianjlai, line 293: i feel like i want to document somewhere that the `config` that arrives in every connector came from reading the same config that was rewritten earlier in `__init__`.

And I think just as a whole in our ideal world, we should not even take in the `config` from the method at all since the official source of truth is the `ManifestDeclarativeSource._config` value.

But I digress. No action here because I don't think this PR should concern itself with the removal of incoming `config` from read/check/discover. But in theory after this change, it should be unneeded

One day we'll eventually get to this!
https://github.com/airbytehq/airbyte-internal-issues/issues/13101

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -139,6 +144,35 @@ def __init__(
...
         self._post_process_manifest()
 
+        self._config: Mapping[str, Any]
+        self._spec_component: Optional[Spec] = None
+        spec = self._source_config.get("spec")
+        if spec:
+            if "type" not in spec:
+                spec["type"] = "Spec"
+            self._spec_component = self._constructor.create_component(SpecModel, spec, dict())
+            mutable_config = dict(config) if config else {}
```
brianjlai, line 154: I think we should move this functionality into it's own private method just to organize the code a little. I envision it to roughly look like:
```
def _migrate_config(self, config: Mapping[str, Any]) -> Mapping[str, Any]:
  # convert config into mutable mapping
  # migrate config
  # write config back to file system
  # transform config
  # return config

self._config = self._migrate_config(self._config)
```

something like that. Just so we keep some of the gory logic in its own method instead of the constructor

brianjlai (commented):
Some stylistic suggestions, but I think it's pretty close. nice work!

airbyte_cdk/sources/declarative/manifest_declarative_source.py
pnilan: Yeah, I agree with you. Migrations *should* work or else fail loudly.

airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -199,6 +216,25 @@ def _normalize_manifest(self) -> None:
             normalizer = ManifestNormalizer(self._source_config, self._declarative_component_schema)
             self._source_config = normalizer.normalize()
 
+    def _migrate_config(
+        self,
+        config_path: Optional[str],
+        mutable_config: MutableMapping[str, Any],
```
brianjlai: we talked about the directly, Instead of taking in both `mutable_config` and `config`, its more straightforward to take in just `config` and perform the duplication into a mutable mapping inside this methodso it can be used by the `spec_component`'s migration and transformation methods. Then return the final config.

brianjlai (approved):
a few discussion points as mentioned, but doesn't need to be blocking. 🚢 

+ the coderabbit suggestion to properly catch error handling on the file write

PR #562, opened 2025-05-21
refactor: connectors can launch themselves; deprecate: `AirbyteEntrypoint` and `launch()`

aaronsteers:
Request: Early Feedback.

Goal here is to detangle the tightly-coupled interdependencies:
- `AirbyteEntrypoint` class
- `launch()` method
- `Connector` and `Source` classes themselves.


airbyte_cdk/__init__.py
```diff
@@ -52,7 +52,7 @@
     create_connector_config_control_message,
     emit_configuration_as_airbyte_control_message,
 )
-from .connector import BaseConnector, Connector
+from .connector import BaseConnector
```
aaronsteers, line 55:
```suggestion
from .connector import BaseConnector, Connector
```

PR #563, opened 2025-05-21
feat: adds config_path to source instantiation for sdm, enabling config migrations

pnilan:
## What
- Updates source-declarative-manifest to inject the config_path to the source, enabling config migrations as part of this epic: https://github.com/airbytehq/airbyte-internal-issues/issues/12530

## Summary by CodeRabbit

- **New Features**
	- Added support for specifying an optional configuration file path when running sources from local YAML files.


brianjlai (approved):
The change itself seems sane, I didn't realize that we had actually moved SDM directly into the repo vs the original `airbyte` repo, so that was new to me ✅

PR #567, opened 2025-05-26
fix(connector-builder): fix property chunking in connector builder

aldogonzalez8:
# WHAT

`SimpleRetrieverTestReadDecorator` when `self._emit_connector_builder_message` is True is missing `additional_query_properties` param,  causing:

```
QueryProperties component is defined but stream_partition does not contain query_properties. Please contact Airbyte Support
```

image

# HOW
We have replaced Retriever decorator by stream_slicer decorator in order to not have to maintain properties in two retrievers.

## Summary by CodeRabbit

- **Tests**
  - Added new tests to validate property chunking in declarative source connectors, ensuring correct handling of chunked HTTP requests and responses.
  - Introduced sample response files for testing currency exchange rate data with property chunking.
  - Updated tests to reflect the new stream slicer decorator usage for limiting slices during test reads.
  - Added tests verifying the behavior and type recognition of the new stream slicer decorator.
- **Bug Fixes**
  - Improved argument handling for internal components to ensure correct parameter passing during test read operations.
- **Refactor**
  - Simplified test read mode by removing slice-limiting retriever decorators and adopting a streamlined stream slicer decorator, centralizing slice limiting logic.
  - Enhanced type checking for stream slicers to recognize decorated instances transparently.
- **New Features**
  - Exposed a new stream slicer decorator to limit the number of slices during test reads, enhancing control over test data volume and debugging.

### DIFF
--- a/airbyte_cdk/sources/declarative/manifest_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/manifest_declarative_source.py
@@ -128,7 +128,7 @@ class ManifestDeclarativeSource(DeclarativeSource):
             component_factory
             if component_factory
             else ModelToComponentFactory(
-                emit_connector_builder_messages,
+                emit_connector_builder_messages=emit_connector_builder_messages,
                 max_concurrent_async_job_count=source_config.get("max_concurrent_async_job_count"),
             )
         )
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -510,7 +510,6 @@ from airbyte_cdk.sources.declarative.retrievers import (
     AsyncRetriever,
     LazySimpleRetriever,
     SimpleRetriever,
-    SimpleRetrieverTestReadDecorator,
 )
 from airbyte_cdk.sources.declarative.retrievers.file_uploader import (
     ConnectorBuilderFileUploader,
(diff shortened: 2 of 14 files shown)


airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -128,7 +128,7 @@ def __init__(
             component_factory
             if component_factory
             else ModelToComponentFactory(
-                emit_connector_builder_messages,
+                emit_connector_builder_messages=emit_connector_builder_messages,
```
aldogonzalez8, line 131: It was instead passed to `limit_pages_fetched_per_slice`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3312,6 +3312,7 @@ def _get_url() -> str:
                 config=config,
                 maximum_number_of_slices=self._limit_slices_fetched or 5,
                 ignore_stream_slicer_parameters_on_paginated_requests=ignore_stream_slicer_parameters_on_paginated_requests,
+                additional_query_properties=query_properties,
```
maxi297, line 3315: Another way to see the problem: Right now, this PR resolves the problem which is a good thing. I'm wondering if we can dissolve the problem instead. It seems like problems like this occurs because we need to maintain two instances for the SimpleRetriever. It also seems like the only functionality missing to remove this class is to have the `maximum_number_of_slices` elsewhere. Would it be too complicated to have a decorator on the StreamSlicer level instead and it would remove the SimpleRetrieverTestReadDecorator making it that when we add something to the SimpleRetriever, we don't forget about the SimpleRetrieverTestReadDecorator?

do we want to just remove the SimpleRetrieverTestReadDecorator? It seems like the only missing feature in order to do that is the slicing

maxi297 (approved):
Just one comment that if you think is worth, maybe we could tackle now. I'm still approving because this fix improves things

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
aldogonzalez8: Ok, I've added a wrapper at stream slicer level

lmossman:
@aldogonzalez8 any reason why we shouldn't merge this Monday morning?

aldogonzalez8:
> @aldogonzalez8 any reason why we shouldn't merge this today / Monday morning?

No, this is good, we can merge early Monday Morning.

PR #568, opened 2025-05-28
fix(concurrent-perpartition-cursor): Fix memory issues

tolik0:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory management and resource cleanup for partition processing, reducing the risk of memory leaks during concurrent operations.

- **Tests**
  - Enhanced test coverage for partition handling, including scenarios with duplicate partitions and improved assertions on internal state cleanup.

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
@@ -295,16 +295,20 @@ class ConcurrentPerPartitionCursor(Cursor):
             ):
                 self._partition_parent_state_map[partition_key] = (deepcopy(parent_state), seq)
 
-        for cursor_slice, is_last_slice, _ in iterate_with_last_flag_and_state(
-            cursor.stream_slices(),
-            lambda: None,
-        ):
-            self._semaphore_per_partition[partition_key].release()
-            if is_last_slice:
-                self._partitions_done_generating_stream_slices.add(partition_key)
-            yield StreamSlice(
-                partition=partition, cursor_slice=cursor_slice, extra_fields=partition.extra_fields
-            )
+        try:
+            for cursor_slice, is_last_slice, _ in iterate_with_last_flag_and_state(
+                cursor.stream_slices(),
+                lambda: None,
+            ):
+                self._semaphore_per_partition[partition_key].release()
+                if is_last_slice:
+                    self._partitions_done_generating_stream_slices.add(partition_key)
+                yield StreamSlice(
+                    partition=partition, cursor_slice=cursor_slice, extra_fields=partition.extra_fields
+                )
+        finally:
+            del cursor
+            del partition
 
     def _ensure_partition_limit(self) -> None:
         """
@@ -492,11 +496,10 @@ class ConcurrentPerPartitionCursor(Cursor):
     def _create_cursor(
         self, cursor_state: Any, runtime_lookback_window: int = 0
     ) -> ConcurrentCursor:
-        cursor = self._cursor_factory.create(
+        return self._cursor_factory.create(
             stream_state=deepcopy(cursor_state),
             runtime_lookback_window=timedelta(seconds=runtime_lookback_window),
         )
-        return cursor
 
     def should_be_synced(self, record: Record) -> bool:
         return self._get_cursor(record).should_be_synced(record)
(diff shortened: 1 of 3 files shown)


airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -474,6 +453,25 @@
...
+           return
+
+        self._semaphore_per_partition.pop(partition_key, None)
+        self._finished_partitions.discard(partition_key)
+
+        seq = self._seq_by_partition.pop(partition_key)
+        self._open_seqs.remove(seq)
+
+        logger.debug(f"Partition {partition_key} fully processed and cleaned up.")
+
```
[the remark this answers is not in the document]
tolik0: _cleanup_if_done is called under the lock

airbyte_cdk/sources/declarative/incremental/concurrent_partition_cursor.py
```diff
@@ -99,19 +99,29 @@ def __init__(
...
-        self._partition_parent_state_map: OrderedDict[str, Mapping[str, Any]] = OrderedDict()
+        self._partition_parent_state_map: OrderedDict[str, tuple[Mapping[str, Any], int]] = (
+            OrderedDict()
+        )
+        self._parent_state: Optional[StreamState] = None
 
+        # Tracks when the last slice for partition is emitted
         self._finished_partitions: set[str] = set()
+        # Used to track the sequence numbers of open partitions
+        self._open_seqs: deque[int] = deque()
```
maxi297, line 110: I think it would be worth defining "open" and "sequence". Here is an attempt at it:
* open means that the not all stream slices for a specific partition key have been processed
* sequence is an number that is being incremented every time we generate stream slices based on a partition. In other words, it is the number of parent records being processed

I'm a bit wary of using the word `sequence` not because I don't think it reflects what it is but rather because we had yet another concept to this class which adds cognitive load. If the definition above is correct, should we change the naming of `next_seq` for `parent_record_count` and `open_seqs` to `processing_parent_indices` or something like that?

maxi297 (commented):
My understanding of this PR: 
* We stopped relying on the `_finished_partitions` to clean the parent state. This allows us to do the following point
* We clean everything once we close a partition which the partition key has been fully processed (i.e. stream slices were created and were all closed)

The drawback is that we now have an integer in memory for each partition (which should be very limited as an impact). The thing I haven't thought of is how this works if the same partition key is emitted multiple time by the parent stream. Let's say we have:
* Parent record ID = 1
* Parent record ID = 2
* Parent record ID = ...
* Parent record ID = 1000
* Parent record ID = 1

It feels like there are two cases here:
* The memory items related to parent record ID = 1 are cleaned (because all stream slices have been processed) before we emit parent record ID = 1 again which means that we will cleaned stuff up for parent record ID = 1 and in that case, it is like we have never processed parent record ID = 1
* We are still processing the stream slices for parent record ID = 1 before we emit it a second time. In that case, does that mean that we won't clean all the parent ids from 2 to 1000 until we reprocess all the parent record ID = 1 again?

unit_tests/sources/declarative/incremental/test_concurrent_perpartitioncursor.py
```diff
@@ -3773,3 +3788,196 @@ def test_given_global_state_when_read_then_state_is_not_per_partition() -> None:
...
+    pr = cursor._partition_router
+    pr.stream_slices.return_value = iter(partitions)
+    pr.get_stream_state.return_value = {}
+
+    generated = list(cursor.stream_slices())
+    # Only “1” and “2” emitted – duplicate “1” skipped
+    assert len(generated) == 2
+
+    # Close “2” first
+    cursor.close_partition(
```
maxi297, line 3970: After closing 2, do we clean the semaphore for 2 as well? I would assume not because 1 hasn't been completed on the second parent record `1` but I fear this might lead to partitions not being cleaned in some very specific case. For example, if even other parent record is the first parent record like [1, 2, 1, 3, 1, 4, 1, 5, 1, 6, 1]. In this case, since 1 pops up frequently, we won't be able to clean 2, 3, 4, 5 and 6.

tolik0:
Regression tests:
- [Jira](https://github.com/airbytehq/airbyte/actions/runs/15583380335/job/43883679444)
- [Klaviyo](https://github.com/airbytehq/airbyte/actions/runs/15583388134/job/43883704844)

PR #569, opened 2025-05-28
feat(connector-builder): add version constraint and wildcard support to manifest migrations

lazebnyi:
## What 
Currently, the migration only accepts catalogs with versions that are not higher than the one specified in the `registry.yaml` file. This causes issues when some catalogs cannot be migrated due to stricter version constraints. We need to enhance the version handling to support a broader range of version definitions such as `*`, `==6.48.3`, `~=1.2`, `>=1.0.0,<2.0.0`, and `6.48.3`.

Original slack thread - https://airbytehq-team.slack.com/archives/C08GNBT7AV7/p1747700015297039
## How 
Extended the `_version_is_valid_for_migration` method to support various version definition formats. Updated `_handle_migration` to adjust the version if necessary. Corresponding tests have been updated accordingly.

## Summary by CodeRabbit

- **Documentation**
  - Improved documentation for manifest migration registration by adding explicit examples of valid version specifiers.
- **Bug Fixes**
  - Enhanced migration version matching to support wildcards and version specifiers, providing more nuanced migration application and version bumping behavior.
- **Tests**
  - Updated and expanded test fixtures and test cases to reflect new versioning logic and migration metadata.
  - Introduced a dummy migration class for testing purposes.
- **Chores**
  - Updated migration registry configuration to use a wildcard version, applying migrations to all versions.

### DIFF
--- a/airbyte_cdk/manifest_migrations/migration_handler.py
+++ b/airbyte_cdk/manifest_migrations/migration_handler.py
@@ -5,6 +5,7 @@
 
 import copy
 import logging
+import re
 from datetime import datetime, timezone
 from typing import Tuple, Type
 
@@ -26,7 +27,7 @@ from airbyte_cdk.manifest_migrations.migrations_registry import (
 METADATA_TAG = "metadata"
 MANIFEST_VERSION_TAG = "version"
 APPLIED_MIGRATIONS_TAG = "applied_migrations"
-
+WILDCARD_VERSION_PATTERN = ".*"
 LOGGER = logging.getLogger("airbyte.cdk.manifest_migrations")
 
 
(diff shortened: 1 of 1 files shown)


airbyte_cdk/manifest_migrations/migration_handler.py
```diff
@@ -112,18 +116,30 @@ def _version_is_valid_for_migration(
...
+           – Wildcard: anything matches.
+        2. String starts with a PEP 440 operator (``==``, ``!=``, ``<=``, ``>=``,
+           ``<``, ``>``, ``~=``, etc.)
+           – Treat *migration_version* as a SpecifierSet and test the manifest
+             version against it.
+        3. Plain version
+           – Interpret both strings as concrete versions and return
+             ``manifest_version <= migration_version``.
+        """
+        if migration_version == "*":
```
bnchrch, line 135: I'd change this to a const and make it *.*.* so that the door is open to regex matches later.

bnchrch (approved):
One nit. But otherwise this is great.

lmossman:
@lazebnyi think you can address Ben's comment and then we can get this merged by EOD Monday?

PR #571, opened 2025-05-29
feat: add `sanitize_url` jinja macro

pnilan:
## What
- Adds `sanitize` jinja macro which defaults to urllib.parse `quote_plus` sanitization.


brianjlai (approved):
lgtm!

PR #576, opened 2025-06-03
fix(async-retriever): fix calling transformations twice

aldogonzalez8:
## What

The async retriever is calling the record transformations twice. Additionally, it appears that the record_selector is duplicated, with one instance not including the record_filter, and then it performs transformations on unfiltered data, which is undesirable.

## Yes, but why?

Async retriever requires a [download retriever](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3447) that under the hood is a [SimpleRetriever](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3468-L3476) with a record_selector, to this record_selector [we pass transformations, but not schema normalization, nor filters](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3451-L3453). 

[Async retriever ](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3633-L3638)also [holds a record_selector](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3504-L3511) with transformations, schema normalization, and filters (all of those if applied in the manifest definition). 

Then both record_selectors try to [filter_and_transform](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/extractors/record_selector.py#L94), causing transformations to apply twice, but with no filters for the record_selector in the download retriever (simple retriever). 

## What can we do?

If we don't provide schema normalization or filters, then we don't need to apply transformations to this other record_selector unless I'm missing something obvious.

Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/13200

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved data retrieval reliability by ensuring that certain data transformations and filters are correctly applied during asynchronous data fetching.  
  - Enhanced data processing accuracy by isolating transformations and filters to specific retrieval steps, preventing unintended data modifications.  
  - Increased test robustness by making stream and record order irrelevant in validation checks.

### DIFF
--- a/unit_tests/sources/declarative/resolvers/test_config_components_resolver.py
+++ b/unit_tests/sources/declarative/resolvers/test_config_components_resolver.py
@@ -203,6 +203,8 @@ def test_dynamic_streams_read_with_config_components_resolver(
             ]
 
         assert len(actual_catalog.streams) == len(expected_stream_names)
-        assert [stream.name for stream in actual_catalog.streams] == expected_stream_names
+        # Use set comparison to avoid relying on deterministic ordering
+        assert set(stream.name for stream in actual_catalog.streams) == set(expected_stream_names)
         assert len(records) == len(expected_stream_names)
-        assert [record.stream for record in records] == expected_stream_names
+        # Use set comparison to avoid relying on deterministic ordering
+        assert set(record.stream for record in records) == set(expected_stream_names)


aldogonzalez8:
> I see it was done as part of [this PR](https://github.com/airbytehq/airbyte-python-cdk/pull/192). I see the reason for the `extra_fields` stuff but I don't see the reason for the transformations [here](https://github.com/airbytehq/airbyte-python-cdk/pull/192#discussion_r1901702963). Was there a bug when Artem added this? @brianjlai do you have more context on this as the reviewer?

@brianjlai if it helps I see [here](https://airbytehq-team.slack.com/archives/C063B9A434H/p1735352429228189) @artem1205 was working on "identified a gap in the functionality of AsyncRetriever needed for amazon ads". Not sure if there was a track for such gap.

brianjlai:
The only thing I can think of was that while fixing the other bug, he noticed that we weren't passing the transformations into the selector and thought it was another oversight so he added it in. But I don't think any of us questioned it and the oversight is we all didn't consider that the AsyncRetriever was responsible for owning the selector and transformation of records.

I think we're fine to make this change

brianjlai (approved):
🛠️

PR #581, opened 2025-06-04
feat(cdv-parser): provide option to provide an array of values that should be set to None in the csv file

aldogonzalez8:


ChristoGrab (approved):
:shipit:

PR #587, opened 2025-06-06
test: remove expected-to-fail tests from fast standard tests (resolves false-positive failures)

aaronsteers:
Eliminate tests that are expected to fail from the fast standard tests to prevent false-positive failures in the test suite. This change improves the reliability of test results.

### DIFF
--- a/airbyte_cdk/test/standard_tests/connector_base.py
+++ b/airbyte_cdk/test/standard_tests/connector_base.py
@@ -163,13 +163,24 @@ class ConnectorTestSuiteBase(abc.ABC):
             ):
                 continue
 
-            test_scenarios.extend(
-                [
-                    ConnectorTestScenario.model_validate(test)
-                    for test in all_tests_config["acceptance_tests"][category]["tests"]
-                    if "config_path" in test and "iam_role" not in test["config_path"]
-                ]
-            )
+            for test in all_tests_config["acceptance_tests"][category]["tests"]:
+                scenario = ConnectorTestScenario.model_validate(test)
+
+                if "config_path" in test and "iam_role" in test["config_path"]:
+                    # We skip iam_role tests for now, as they are not supported in the test suite.
+                    continue
+
+                if scenario.expect_exception:
+                    # For now, we skip tests that are expected to fail.
+                    # This is because they create false-positives in the test suite
+                    # if they fail later than expected.
+                    continue
+
+                if scenario.config_path in [s.config_path for s in test_scenarios]:
+                    # Skip duplicate scenarios based on config_path
+                    continue
+
+                test_scenarios.append(scenario)
 
         connector_root = cls.get_connector_root_dir().absolute()
         for test in test_scenarios:


airbyte_cdk/test/standard_tests/connector_base.py
```diff
@@ -163,13 +163,24 @@ def get_scenarios(
...
+                    # We skip iam_role tests for now, as they are not supported in the test suite.
+                    continue
+
+                if scenario.expect_exception:
+                    # For now, we skip tests that are expected to fail.
+                    # This is because they create false-positives in the test suite
+                    # if they fail later than expected.
+                    continue
+
+                if scenario.config_path in [s.config_path for s in test_scenarios]:
```
aaronsteers, line 179:
```suggestion
                if scenario.config_path and scenario.config_path in [s.config_path for s in test_scenarios]:
```

airbyte_cdk/test/standard_tests/models/scenario.py
```diff
@@ -24,6 +24,8 @@ class ConnectorTestScenario(BaseModel):
     acceptance test configuration file.
     """
 
+    model_config = ConfigDict(frozen=True)
```
aaronsteers, line 27:
```suggestion
    # Allows the class to be hashable, which PyTest will require
    # when we use to parameterize tests.
    model_config = ConfigDict(frozen=True)
```

airbyte_cdk/test/standard_tests/models/scenario.py
```diff
@@ -83,3 +96,29 @@ def __str__(self) -> str:
             return f"'{self.config_path.name}' Test Scenario"
 
         return f"'{hash(self)}' Test Scenario"
+
+    def without_expecting_failure(self) -> ConnectorTestScenario:
```
aaronsteers, line 100: This method and the below allow us to get a mutated instance of the scenario, while still treating each individual scenario instance as immutable.

airbyte_cdk/test/standard_tests/util.py
```diff
@@ -67,7 +67,7 @@ def create_connector_test_suite(
         )
 
     subclass_overrides: dict[str, Any] = {
-        "get_connector_root_dir": lambda: connector_directory,
+        "get_connector_root_dir": classmethod(lambda cls: connector_directory),
```
aaronsteers, line 70: This fixes previous override not properly dealing with the `cls` arg of the classmethod.

airbyte_cdk/test/entrypoint_wrapper.py
```diff
@@ -24,6 +24,7 @@
 from typing import Any, List, Mapping, Optional, Union
 
 import orjson
+from langsmith import expect
```
aaronsteers, line 27:
```suggestion
```

airbyte_cdk/test/standard_tests/models/scenario.py
```diff
@@ -83,3 +135,38 @@ def __str__(self) -> str:
...
         return f"'{hash(self)}' Test Scenario"
+
+    def without_expected_outcome(self) -> ConnectorTestScenario:
+        """Return a copy of the scenario that does not expect failure or success.
+
+        This is useful when running multiple steps, to defer the expectations to a later step.
+        """
+        return ConnectorTestScenario(
+            **self.model_dump(exclude={"status"}),
+        )
```
aaronsteers, line 146: @dbgold17 - Per your feedback, and after further consideration, I've refactored this as a tri-state value, and renamed from `expect_exception` to `expected_outcome`. That makes this `without_()` method make more sense, and then we separately have bools for `with_expecting_failure()` and `with_expecting_success()`. Some additional refactoring was needed, since the same code was called in other places, but this implementation makes a lot more sense and can handle more test scenarios.

airbyte_cdk/test/utils/reading.py
```diff
@@ -19,8 +20,16 @@ def read_records(
...
+    *,
+    expecting_exception: bool | None = None,  # Deprecated, use expected_outcome instead.
+    expected_outcome: ExpectedOutcome | None = None,
 ) -> EntrypointOutput:
     """Read records from a stream."""
+    if expecting_exception is not None and expected_outcome is not None:
+        raise ValueError("Cannot set both expecting_exception and expected_outcome.")
+
+    if expected_outcome is None:
+        expected_outcome = ExpectedOutcome.from_expecting_exception_bool(expecting_exception)
```
dbgold17, line 32: Why do we need to keep the old `expecting_exception` flag here?

If we do need to keep it, what do you think about requiring exactly one of the two parameters (`expecting_exception`, `expected_outcome`) to be set? 

It seems like, as written, the default case where both `expecting _exception` and `expected_outcome` are not set will lead to a different behavior than previously defined. Now they lead to an `ALLOW_ANY` outcome which seems to be different from the default case prior to this refactor (essentially `EXPECT_SUCCESS`).

airbyte_cdk/test/models/outcome.py
```diff
@@ -0,0 +1,57 @@
...
+        if expecting_exception is None:
+            return ExpectedOutcome.ALLOW_ANY
+
+        return (
+            ExpectedOutcome.EXPECT_EXCEPTION
+            if expecting_exception
+            else ExpectedOutcome.EXPECT_SUCCESS
+        )
+
+    def expect_exception(self) -> bool:
```
dbgold17, line 51: should this and `expect_success` also return true if `self == ExpectedOutcome.ALLOW_ANY`?

dbgold17 (commented):
a couple comments, but generally looks good. Would you like help testing this out on some connectors?

airbyte_cdk/test/utils/reading.py
aaronsteers: @dbgold17 - Some connectors reference this code directly. I'm biasing towards _not breaking_ any connectors if I can avoid it. Once connectors use the new paradigm, the old can be phased out without impacting anything.

dbgold17 (approved):
lgtm

PR #589, opened 2025-06-06
feat: add conditions to validators

pnilan:
## What
- Adds conditions to validators, enabling validators to only be invoked when they meet certain conditions, for example validate that a value adheres to schema only if its optional key is included in the config.

## Summary by CodeRabbit

- **New Features**
  - Added support for conditional validation in DpathValidator and PredicateValidator, enabling validation strategies to be applied only when specified conditions based on configuration are met.
- **Tests**
  - Added tests confirming validation is skipped when conditions evaluate to false; updated existing tests to include the new condition parameter.


pnilan:
Closing in favor of simply skipping validation if returned values are falsy: https://github.com/airbytehq/airbyte-python-cdk/pull/590

PR #602, opened 2025-06-18
fix: `QueryProperties` component creation in `model_to_component_factory.py`

pnilan:


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3193,7 +3193,10 @@ def _get_url() -> str:
 
             query_properties_definitions = []
             for key, request_parameter in model.requester.request_parameters.items():  # type: ignore # request_parameters is already validated to be a Mapping using _query_properties_in_request_parameters()
-                if isinstance(request_parameter, QueryPropertiesModel):
```
pnilan, line 3196: the `request_parameter` has not been parsed as a `QueryPropertiesModel` at this point, it is still a dictionary so this condition never evaluates to true.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3193,7 +3193,10 @@ def _get_url() -> str:
...
+                if (
```
dbgold17, line 3196: so this is a pretty baseless comment because I've only spent a tiny amount of time looking at the m2c factory and dcs, but it feels a little icky to me that we need to write code like this. Conceptually I thought we should only need to define the schema in terms of pydantic models once and then it can handle the branching cases of when to hydrate which component. Idk if that makes any sense. Figured I'd pass on my high-level impression even if I don't have any specific recommendations for what to do differently.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: This is a good catch, but what I'm trying to figure out in the code is why we aren't able to parse this into the model. My understanding (which embarrassingly I should know since I wrote it) is that we parse the entire stream dictionary into the entire pydantic nested all the way down. So what is weird is that I would think it could parse down to the HttpRequester and below, so its odd that it stays a dictionary.

What you have seems like it would work, but it feels like a bug that we aren't able to parse this into a `QueryPropertiesModel` and are forced to catch it and convert this into a model here in  `create_simple_retriever()` when it should've already been.

Not a complete blocker and if we do go with this solution we should comment this, but something still feels suspicious and we might want to dig into why we don't properly turn this into a model earlier when we parse it into models using `model_type.parse_obj(component_definition)`

PR #605, opened 2025-06-19
fix(low-code): sync records without cursor field in ConcurrentPerPartitionCursor

darynaishchenko:
`ConcurrentPerPartitionCursor.observe` catch case when record doesn't have a cursor value, skip updating state logic for this record.
Same logic as in `DatetimeBasedCursor.observe`: https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py#L155

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of records missing a cursor value to prevent errors and ensure stable incremental data syncing.
	- Enhanced test reliability by relaxing order dependency in record verification.
	- Added new tests to verify correct behavior when records lack cursor values, ensuring consistent incremental sync state management.


darynaishchenko:
> Can you please add the unit test for records without a cursor?

added

PR #611, opened 2025-06-25
fix: override configure method in `ManifestDeclarativeSource` to use migrated/transformed config

pnilan:
## What
- The entrypoint was passing the unmigrated/untransformed config to each operation methods (check/discover/read)
- This fix overrides the Source method `configure` in the `ManifestDeclarativeSource` ultimately returning the transformed/migrated config to the entrypoint, where the config can then be validated against the spec and passed to each operation method.

### DIFF
--- a/airbyte_cdk/sources/declarative/manifest_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/manifest_declarative_source.py
@@ -238,6 +238,11 @@ class ManifestDeclarativeSource(DeclarativeSource):
         self._spec_component.transform_config(mutable_config)
         return mutable_config
 
+    def configure(self, config: Mapping[str, Any], temp_dir: str
+    ) -> Mapping[str, Any]:
+        config = self._config or config
+        return super().configure(config, temp_dir)
+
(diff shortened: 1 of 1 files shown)


brianjlai (commented):
change seems fine, but can we test this e2e w/ amazon ads to make sure this works.

Also how useful or easy would it be to unit test this. It's probably trivial or the result might be the same, but still feels like we should exercise the code path

brianjlai (approved):

pnilan:
Note:
- I manually tested this change by creating a [SDM dev image](https://hub.docker.com/layers/airbyte/source-declarative-manifest/6.56.4.post1.dev15883596331/images/sha256-50171e0e706617859b9b90770d91f368959cceb0fde500f3e41decd6fbabb39a) and then testing transformations on `source-amazon-seller-partner` and `source-amplitude`. I observed that w/o this change, the untransformed config was being used for check/discover/read operations. When updating the underlying SDM images to the dev image, the check/discover/read operations began using the transformed config as expected.

PR #614, opened 2025-06-26
fix(low-code): handle `ScannerError` in `ConfigComponentsResolver`

darynaishchenko:
For [bing-ads custom reports](https://github.com/airbytehq/airbyte-internal-issues/issues/12995) we need to define cursor format based on report aggregation provided in the config by users.
Manifest will look like this:
```
 - type: ComponentMappingDefinition
   create_or_update: true
   field_path: [ "incremental_sync", "datetime_format" ]
   value: "{{ '%Y-%m-%dT%H:%M:%S+00:00' if components_values['report_aggregation'] == 'Hourly' else '%Y-%m-%d' }}"
```
When format string is being parsed in the `_parse_yaml_if_possible` Scanner error is raised. 
Added check for the error message for datetime format case to pass this value to stream template.

### DIFF
--- a/airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
+++ b/airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
@@ -11,6 +11,7 @@ import dpath
 import yaml
 from typing_extensions import deprecated
 from yaml.parser import ParserError
+from yaml.scanner import ScannerError
 
 from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
 from airbyte_cdk.sources.declarative.resolvers.components_resolver import (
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/resolvers/config_components_resolver.py
```diff
@@ -185,4 +186,8 @@ def _parse_yaml_if_possible(value: Any) -> Any:
                 return yaml.safe_load(value)
             except ParserError:  # "{{ record[0] in ['cohortActiveUsers'] }}"   # not valid YAML
                 return value
+            except ScannerError as e:  # "%Y-%m-%d'   # not valid yaml
+                if "expected alphabetic or numeric character, but found '%'" in str(e):
+                    return value
+                raise e
```
aldogonzalez8, line 192: Can we check the value is in a set of expected values for dateformat?

e.g. ('%Y-%m-%d', '%Y-%m-%dT%H:%M:%S')

darynaishchenko: I don't think this is a part of component resolver logic, `DatetimeBasedCursor` will handle this if value in record doesn't match expected type

aldogonzalez8 (approved):
Approved

PR #628, opened 2025-06-30
chore: add unidecode dep to python CDK

ChristoGrab:
## What

As a prerequisite of migrating Google Sheets to manifest-only, we need to include the unidecode dependency to the CDK. 

[Unidecode](https://pypi.org/project/Unidecode/) lets us transliterate Unicode text into plain ASCII characters. In other words, it converts characters with accents, diacritics, or non-Latin scripts into a closest possible ASCII representation:

```
"kožušček" → "kozuscek"
"北亰" → "Bei Jing"
```

## Summary by CodeRabbit

* **Chores**
  * Added a deprecated dependency to support a specific migration, with plans for future removal.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 


ChristoGrab:
Probably not needed since the `anyascii` library has similar functionality. Leaving in draft for now but will likely close this PR.

ChristoGrab:
Psych! Reopening as regression tests showed differences in how emojis are handled by these libraries. Since we create schema properties from sheet cell names in Google Sheets, this means stream property names can be altered, which would be a breaking change. We are going to include the dep in the CDK to unblock the migration, with a comment that it should not be used outside Google Sheets and is intended to be deprecated (probably make for a good L3 issue)

aldogonzalez8 (approved):
APPROVED

PR #633, opened 2025-07-01
feat: make use_parent_params configurable

ChristoGrab:
## What

In version 6.43.1 we introduced the `use_parent_parameters` param to allow parent parameters to take precedence over component parameters when propagating them in a component. However, because we set `use_parent_parameters` to always be true for dynamic streams, connectors that need to override parent parameters dynamically can no longer do so. This PR makes the parameter configurable in the manifest. We still default to True as I do not want to risk breaking connectors that are already using the parent_param logic.

## How

- Added `use_parent_parameters` to the DeclarativeDynamicStream schema, making it configurable in the manifest.

## Summary by CodeRabbit

* **New Features**
  * Added a new option to control whether parent parameters are prioritized over component parameters when configuring dynamic streams. This option defaults to prioritizing parent parameters for compatibility.

* **Tests**
  * Introduced new test cases to verify parameter precedence behavior based on the new configuration option in both static and dynamic stream scenarios.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 


unit_tests/sources/declarative/parsers/test_manifest_component_transformer.py
```diff
@@ -569,3 +569,220 @@ def test_propagate_property_chunking():
     transformer = ManifestComponentTransformer()
     actual_component = transformer.propagate_types_and_parameters("", component, {})
     assert actual_component == expected_component
+
+
+def test_use_parent_parameters_configuration():
```
aldogonzalez8, line 574: Nit: can we parametrize this one?

ChristoGrab: Done 👍

PR #635, opened 2025-07-02
chore: reorder inject_into in RequestOption and fix title and description in refresh token updater

lmossman:
Inject Into should be above field name/path, as this makes more conceptual sense in the UI.

I also changed some of the titles, descriptions, and default values in refresh_token_updater to better match what the user should expect in the UI

## Summary by CodeRabbit

* **Refactor**
  * Updated terminology and descriptions related to token refresh functionality for clarity.
  * Reorganized request option settings for improved schema consistency.

### DIFF
--- a/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
+++ b/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
@@ -1307,6 +1307,12 @@ class InjectInto(Enum):
 
 class RequestOption(BaseModel):
     type: Literal["RequestOption"]
+    inject_into: InjectInto = Field(
+        ...,
+        description="Configures where the descriptor should be set on the HTTP requests. Note that request parameters that are already encoded in the URL path will not be duplicated.",
+        examples=["request_parameter", "header", "body_data", "body_json"],
+        title="Inject Into",
+    )
     field_name: Optional[str] = Field(
         None,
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -477,19 +477,19 @@ class RefreshTokenUpdater(BaseModel):
         title="Refresh Token Property Name",
     )
     access_token_config_path: Optional[List[str]] = Field(
-        ["credentials", "access_token"],
+        ["access_token"],
```
aldogonzalez8, line 480: Nit: 
> better match what the user should expect in the UI

Is that currently confusing?

aldogonzalez8 (approved):
APPROVED
just left an annoying question behind. Non blocker.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
lmossman: It is confusing because in the Builder, nested config paths aren't fully supported, so users are more likely to just want to use `access_token` instead of `credentials.access_token`

aldogonzalez8: Oh yes, I think I remember that problem, trying to fork connectors with nested credentials.

PR #637, opened 2025-07-03
fix: properly set token_expiry_is_time_of_expiration and mask access token when logging

lmossman:
## What

When testing out some Builder changes relating to OAuth, I noticed two issues with the implementation of OAuth in the CDK:
1. When I didn't set a token expiry date in my config, I was receiving this error: `Invalid expires_in value: 2025-08-01T21:34:33Z. Expected number of seconds when no format specified.`
2. The access token was not being masked with `****` in the OAuth response shown in the Builder - I could see the raw value

## How

I traced the first issue down to the fact that `token_expiry_is_time_of_expiration` is not being set when constructing `DeclarativeSingleUseRefreshTokenOauth2Authenticator` [here](https://github.com/airbytehq/airbyte-python-cdk/blob/a562875fafc8b0c6bb658dd2be20f7475cc5dd0b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L2758C20-L2758C71).

This caused this [if statement](https://github.com/airbytehq/airbyte-python-cdk/blob/4e6b9f230bd129f8271cd2b2eb820064902d5b0c/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L277) to always return False, causing the error to be thrown.

To fix this, I simply set the `token_expiry_is_time_of_expiration` the same way it is being set when constructing `DeclarativeOauth2Authenticator` [below](https://github.com/airbytehq/airbyte-python-cdk/blob/a562875fafc8b0c6bb658dd2be20f7475cc5dd0b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L2827)

---

The second issue I traced down to the fact that the OAuth response was being logged before anything had a chance to add the access token to the secrets list. The fix was to extract the access token and add it to the secrets mask list before logging the response

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of OAuth token expiry by correctly setting the token expiry parameter during authentication.
  * Enhanced security by masking access tokens in logs immediately after retrieval from the token refresh endpoint.
  * Improved error handling by checking response status before processing OAuth token requests.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2801,6 +2801,7 @@ class ModelToComponentFactory:
                 ).eval(config),
                 scopes=model.scopes,
                 token_expiry_date_format=model.token_expiry_date_format,
+                token_expiry_is_time_of_expiration=bool(model.token_expiry_date_format),
                 message_repository=self._message_repository,
                 refresh_token_error_status_codes=model.refresh_token_updater.refresh_token_error_status_codes,
                 refresh_token_error_key=model.refresh_token_updater.refresh_token_error_key,
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -225,10 +225,14 @@ class AbstractOauth2Authenticator(AuthBase):
 
             response_json = response.json()
 
-            # extract the access token and add to secrets to avoid logging the raw value
-            access_key = self._extract_access_token(response_json)
-            if access_key:
-                add_to_secrets(access_key)
+            try:
+                # extract the access token and add to secrets to avoid logging the raw value
+                access_key = self._extract_access_token(response_json)
+                if access_key:
+                    add_to_secrets(access_key)
+            except ResponseKeysMaxRecurtionReached as e:
+                # could not find the access token in the response, so do nothing
+                pass
             
             self._log_response(response)
 
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -218,10 +218,14 @@ class AbstractOauth2Authenticator(AuthBase):
                 headers=self.build_refresh_request_headers(),
             )
             response_json = response.json()
-            # extract the access token and add to secrets to avoid logging the raw value
-            access_key = self._extract_access_token(response_json)
-            if access_key:
-                add_to_secrets(access_key)
+            try:
+                # extract the access token and add to secrets to avoid logging the raw value
+                access_key = self._extract_access_token(response_json)
+                if access_key:
+                    add_to_secrets(access_key)
+            except ResponseKeysMaxRecurtionReached as e:
+                ## Could not find the access token in the response, so do nothing
+                pass
             # log the response even if the request failed for troubleshooting purposes
             self._log_response(response)
             response.raise_for_status()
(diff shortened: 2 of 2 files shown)


airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -217,10 +217,15 @@ def _make_handled_request(self) -> Any:
                 data=self.build_refresh_request_body(),
                 headers=self.build_refresh_request_headers(),
             )
+            response_json = response.json()
+            # extract the access token and add to secrets to avoid logging the raw value
+            access_key = self._extract_access_token(response_json)
```
dbgold17, line 222: maybe a more complex change than what's in scope for this PR, but it feels like a safer default to have whatever method actually make the API request (maybe `make_handled_request` to fetch access tokens also call the `add_to_secrets()` method. That way its impossible to not have it masked if you are using the cdk for auth. wdyt?

lmossman: This change is being applied to `_make_handled_request`, so anything that uses that will have the access token be added to secrets. 

I don't think there is any lower-level method that this logic can be added to, because this method just calls `requests.request()` which is a library method

dbgold17: oh my mistake. I thought this change was in another file. Looks great 👍 

I did notice that we make another call to `add_to_secrets()` here which is probably now unnecessary.
https://github.com/airbytehq/airbyte-python-cdk/blob/5824a5eabcd4e4bd72cbb06af7d6f18430ca1be5/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L260-L262

Should we take that out to avoid confusion for why we are doing it twice?

Probably also worth checking that all flavors of OAuth classes are calling the `add_to_secrets` method either through this method or if they are using some other method to make the auth request.

dbgold17: on that last point, I'm not 100% sure that all Auth classes subclass the `AbstractOauth2Authenticator` class and use `_make_handled_request()`

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2801,6 +2801,7 @@ def create_oauth_authenticator(
                 ).eval(config),
                 scopes=model.scopes,
                 token_expiry_date_format=model.token_expiry_date_format,
+                token_expiry_is_time_of_expiration=bool(model.token_expiry_date_format),
```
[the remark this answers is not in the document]
lmossman: No, I want empty strings to be treated as False

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -217,10 +217,33 @@ def _make_handled_request(self) -> Any:
                 data=self.build_refresh_request_body(),
                 headers=self.build_refresh_request_headers(),
             )
+
```
dbgold17, line 220: nit: but wonder if there is a more straightforward way to write this logic. Something like

```
            # log the response even if the request failed for troubleshooting purposes
            self._log_response(response)
            response.raise_for_status()
            
            response_json = response.json()

            try:
                # extract the access token and add to secrets to avoid logging the raw value
                access_key = self._extract_access_token(response_json)
                if access_key is not None:
                  add_to_secrets(access_key)
            except ResponseKeysMaxRecurtionReached as e:
                # could not find the access token in the response, so do nothing
                pass
   
            return response_json
```

dbgold17: also, are we sure it makes sense to ignore the `ResponseKeysMaxRecurtionReached` exception? Admittedly this code is weird, but it seems like we could theoretically hit a max recursive depth but not fully explore the response json and therefore if we ignore the error, the access_key can actually be returned and we would not mask it.

lmossman: Discussed this over zoom and came to a simpler implementation that we both liked

PR #638, opened 2025-07-03
feat(property-chunking): Allow the definition of query_properties on the HttpRequester component

brianjlai:
## What

While migrating the last `source-instagram` stream `UserInsights` to low-code, I found a gap in the CDK functionality in how we chunk properties to make multiple requests and merge the records back together.

The current implementation only allows us to put all query properties under a single field in the `request_properties`. However, for Instagram, we need to load certain query properties into one parameter field and other properties to another. For example:

```
METRICS_BY_PERIOD = {
        "day": [
            "follower_count",
            "reach",
        ],
        "week": ["reach"],
        "days_28": ["reach"],
        "lifetime": ["online_followers"],
    }

We want to create 4 chunks where query params are:
?period=day&metric=follower_count,reach
?period=week&metric=reach
?period=days_28&metric=reach
?period=lifetime&metric=online_followers
```

And we still want to be able to merge the record back together for all the chunks based on a target merge key.

## How

In reality how we do this quite easy, but relied mostly on an oversight in how I had previously designed the `fetch_properties_from_endpoint` improvement from a few months ago. I had originally envisioned that we would not really need to do property chunking here and if it was needed, it would be handled from the definition in `request_properties`. However, we're now seeing a case where we want to load parameters into different fields.

By allowing for the `QueryProperties` which supports property chunking to be defined the HttpRequester, we can perform chunking and choose where they go.

I've marked `fetch_properties_from_endpoint` as deprecated and would ideally want to do an under the table deprecation of the old field. Only `source-hubspot` uses this and we can swap this out. Once we do, we can silently deprecate.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new property for HTTP requesters to explicitly specify query properties for APIs that require them.

* **Deprecation**
  * Marked the existing property for fetching properties from endpoints as deprecated, with guidance to use the new query properties feature.

* **Bug Fixes**
  * Updated tests to use the new query properties structure, ensuring compatibility and correctness.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -2043,9 +2043,15 @@ definitions:
           - "$ref": "#/definitions/NoAuth"
           - "$ref": "#/definitions/LegacySessionTokenAuthenticator"
       fetch_properties_from_endpoint:
+        deprecated: true
+        deprecation_message: "Use `query_properties` field instead."
         title: Fetch Properties from Endpoint
         description: Allows for retrieving a dynamic set of properties from an API endpoint which can be injected into outbound request using the stream_partition.extra_fields.
         "$ref": "#/definitions/PropertiesFromEndpoint"
+      query_properties:
+        title: Query Properties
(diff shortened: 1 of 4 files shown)


aldogonzalez8 (approved):
APPROVED

PR #646, opened 2025-07-05
feat(rate-lmimit): update debugging

aldogonzalez8:


airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -506,6 +507,7 @@ def send_request(
         dedupe_query_params: bool = False,
         log_formatter: Optional[Callable[[requests.Response], Any]] = None,
         exit_on_rate_limit: Optional[bool] = False,
+        stream_slice: Optional[Mapping[str, Any]] = None,
```
maxi297, line 510: Is the connector you are trying to update low-code? 

Wild idea: instead of having to modify all the interfaces that would require us to low more information, could we create [a thread-local](https://docs.python.org/3/library/threading.html#thread-local-data) in a threading.py file that users that want to log more information like the slice could have access to this? `PartitionReader.process_partition` would basically do `thread_local.stream_slice = partition.to_slice` and at that point, http_client can just get the information from `thread_local.stream_slice`. If anyone else than the HttpClient need access to this information, they just need to do the same

aldogonzalez8: > Is the connector you are trying to update low-code?

@maxi297  Yes, it's Slack, I received the feedback from a user who was unable to say after a few hours if [the attempt was "stuck"](https://cloud.airbyte.com/workspaces/bfc80cbb-66e1-4f60-af3c-63f6970e1986/connections/bf9006b2-c7f2-44f1-abec-215daa50ff45/timeline?openLogs=true&eventId=154b5a5c-8370-4cdd-8b43-ae4abe704f1f). Honestly, I couldn't say as the rate limit logs won't tell me too much. 

I was recently working on[ bumping Cdk 6](https://github.com/airbytehq/airbyte/pull/62110) for this connector and was trying to get better logs on the fly.

aldogonzalez8: @maxi297 I need to do some cleanup, but I added the singleton class that inherits from local and added some tests around.

airbyte_cdk/sources/declarative/request_local/request_local.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @classmethod
+    def get_instance(cls):
+        """
+        Get the singleton instance of RequestLocal.
+        This is the recommended way to get the instance.
+        """
+        if cls._instance is None:
+            cls._instance = cls()
+        return cls._instance
```
aldogonzalez8, line 39: I think we don't need this

airbyte_cdk/sources/declarative/requesters/http_requester.py
```diff
@@ -473,6 +477,7 @@ def send_request(
             dedupe_query_params=True,
             log_formatter=log_formatter,
             exit_on_rate_limit=self._exit_on_rate_limit,
+            stream_slice=stream_slice,
```
aldogonzalez8, line 480: remove this.

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -243,6 +243,7 @@ def _send_with_retry(
         request_kwargs: Mapping[str, Any],
         log_formatter: Optional[Callable[[requests.Response], Any]] = None,
         exit_on_rate_limit: Optional[bool] = False,
+        stream_slice: Optional[Mapping[str, Any]] = None,
```
aldogonzalez8, line 246: remove this

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -526,6 +528,7 @@ def send_request(
             request_kwargs=request_kwargs,
             log_formatter=log_formatter,
             exit_on_rate_limit=exit_on_rate_limit,
+            stream_slice=stream_slice,
```
aldogonzalez8, line 531: remove this

unit_tests/sources/declarative/requesters/test_http_requester.py
```diff
@@ -229,6 +254,19 @@ def test_basic_send_request():
...
+    options_provider.get_request_headers.return_value = {"my_header": "my_value"}
+    requester = create_requester_rate_limited()
+    requester._request_options_provider = options_provider
+    with caplog.at_level(logging.INFO, logger="airbyte"):
+        with pytest.raises(RateLimitBackoffException):
+            requester.send_request(stream_slice={"start": "2012"})
+
+
+    logged_messages = [record.message for record in caplog.records]
+    assert "Caught retryable error 'Too many requests.' after 1 tries. Waiting 1 seconds then retrying for slice: {'start': '2012'}..." in logged_messages
```
aldogonzalez8, line 269: Here we can see the stream slice read from request_local

PR #649, opened 2025-07-09
fix(low-code cdk): Add literal type for ConfigNormalizationRules components

brianjlai:
While implementing config validations for a connector, I noticed that we weren't requiring a `type` on the `ConfigNormalizationRules` component. For consistency we should do so.

The one annoying part is that this would technically constitute a breaking change because `type` will now be required on all instances of the type and no existing connectors contain it. Rather than institute a trivial breaking change, I propose that we prepare PRs for the few uses we have of this component that add the type. Merge this as a minor version, then update the base images for those connectors.

This should be relatively low risk since it will fail quite loudly and the changes themselves are quite trivial

## Summary by CodeRabbit

* **New Features**
  * The configuration now requires a "type" field for config normalization rules, ensuring clearer identification and validation.
* **Tests**
  * Updated test manifests to include the required "type" field for config normalization rules.

### DIFF
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -3835,7 +3835,12 @@ definitions:
         title: Config Normalization Rules
         type: object
         additionalProperties: false
+        required:
+          - type
         properties:
+          type:
+            type: string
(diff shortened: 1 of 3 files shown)


pnilan (approved):
Thanks!

PR #654, opened 2025-07-15
fix: fast tests conditionally skip streams

dbgold17:
## What

fixes: https://github.com/airbytehq/airbyte-internal-issues/issues/13521

This PR attempts to make FAST tests respect the `empty_streams` key in acceptance-test-config.yml files by choosing not to sync data from those streams. Historically the `empty_streams` key was used not only to denote streams that don't have any data to sync but also streams that will fail syncs for a variety of reasons. Running tests against these streams has been leading to false positive failures. This PR attempts to replicate the behavior of CATs with a bit less sophistication, fitting with the goals of the FAST test suite. 

This implementation somewhat messy due to how the test suite currently parses these files - extracting only a subset of test scenarios to run all operations (spec, check, read) against.

I see this as a temporary fix until we are able to redesign how we want to configure FASTs and deprecate these files.

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying empty streams with optional bypass reasons in connector test scenarios. Scenarios can now identify streams expected to be empty and exclude them from read tests.

* **Bug Fixes**
  * Improved test behavior to correctly filter out empty streams during connector acceptance tests.

* **Refactor**
  * Enhanced merging of duplicate test scenarios to preserve all empty stream information.
  * Updated function signatures and internal logic for better configuration handling and code clarity.
  * Removed `@dataclass` decorator from a custom exception class for simplicity.

* **Chores**
  * Removed unused imports for cleaner codebase.

### DIFF
--- a/airbyte_cdk/test/entrypoint_wrapper.py
+++ b/airbyte_cdk/test/entrypoint_wrapper.py
@@ -51,7 +51,6 @@ from airbyte_cdk.sources import Source
 from airbyte_cdk.test.models.scenario import ExpectedOutcome
 
 
-@dataclass
 class AirbyteEntrypointException(Exception):
     """Exception raised for errors in the AirbyteEntrypoint execution.
 
--- a/airbyte_cdk/test/standard_tests/docker_base.py
+++ b/airbyte_cdk/test/standard_tests/docker_base.py
@@ -143,7 +143,7 @@ class DockerConnectorTestSuite:
         This has to be a separate function because pytest does not allow
         parametrization of fixtures with arguments from the test class itself.
         """
-        categories = ["connection", "spec"]
+        categories = ["spec", "connection", "basic_read"]
         try:
             all_tests_config = cls.acceptance_test_config
         except FileNotFoundError as e:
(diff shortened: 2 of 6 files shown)


airbyte_cdk/utils/connector_paths.py
```diff
@@ -86,8 +86,6 @@ def _find_in_adjacent_dirs(current_dir: Path) -> Path | None:
 
 def resolve_connector_name_and_directory(
     connector_ref: str | Path | None = None,
-    *,
-    connector_directory: Path | None = None,
```
dbgold17, line 90: this parameter is never passed in

dbgold17:
Just got tests passing on `source-monday` locally which expects the suite to bypass the `teams` stream!

I'm going to test this out on a couple other connectors in the next day or so

airbyte_cdk/test/standard_tests/docker_base.py
```diff
@@ -121,15 +123,29 @@ def get_scenarios(
...
+        # Remove duplicate scenarios based on config_path.
+        deduped_test_scenarios: list[ConnectorTestScenario] = []
+        for scenario in test_scenarios:
+            for existing_scenario in deduped_test_scenarios:
+                if scenario.config_path == existing_scenario.config_path:
+                    # If a scenario with the same config_path already exists, we merge the empty streams.
+                    # scenarios are immutable, so we create a new one.
+                    all_empty_streams = (existing_scenario.empty_streams or []) + (
+                        scenario.empty_streams or []
+                    )
```
aaronsteers, line 137: This implementation looks correct but we may want to dedupe it after creation.


```suggestion
                    all_empty_streams = (existing_scenario.empty_streams or []) + (
                        scenario.empty_streams or []
                    )
                    # Dedupe the list:
                    all_empty_streams = list(set(all_empty_streams))
```

aaronsteers (approved):
Looks great, @dbgold17! Thanks for creating this. One comment on deduping, but honestly even without the deduping, the code should work correct (deduped in a later step).

aaronsteers:
> Just got tests passing on `source-monday` locally which expects the suite to bypass the `teams` stream!
> 
> I'm going to test this out on a couple other connectors in the next day or so

Awesome! 🎉

airbyte_cdk/test/entrypoint_wrapper.py
```diff
@@ -51,7 +51,6 @@
 from airbyte_cdk.test.models.scenario import ExpectedOutcome
 
 
-@dataclass
```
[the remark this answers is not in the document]
dbgold17: jc, what is the case for deriving Exceptions from dataclasses as opposed to regular subclassing of the Exception class?

PR #658, opened 2025-07-16
fix: add array type to field_pointers items

lmossman:
The `items` field of `field_pointers` in declarative_component_schema has a nested `items` field, since that is meant to contain nested array values.

However, there is no `type: array` on the outer `items` schema, which causes issues in the Builder UI as it relies on the types of fields to know what to render.

This PR fixes the issue by explicitly declaring `type: array` on the outer `items` field.

## Summary by CodeRabbit

* **New Features**
  * Expanded support for custom configuration transformations, allowing additional transformation types in configuration migration and normalization rules.

* **Bug Fixes**
  * Improved schema validation for transformation definitions by clarifying the expected structure of field pointers.


ChristoGrab (approved):
LGTM! Thanks for cleaning up the CustomConfigTransformation class as well 👍

PR #674, opened 2025-07-28
chore: remove cursor.is_greater_than_or_equal

maxi297:
## What

We still have a couple of usage of the declarative cursors. We would rather use the concurrent one because the declarative ones are maintained to a minimum. This will allow us to migrate to the DefaultStream and remove the DeclarativeStream down the line.

## How

* Clean `is_greater_than_or_equal`: based on the comment above in SimpleRetriever.read_records, it seems like we can tackle https://github.com/airbytehq/airbyte-internal-issues/issues/6955 and remove this
* [TBD] migrate SubstreamPartitionRouter to concurrent framework

## Summary by CodeRabbit

* **Refactor**
  * Removed record comparison features from multiple cursor types to simplify data processing.
  * Simplified record tracking and slice closure in data retrieval for better maintainability.

* **Tests**
  * Removed tests related to record comparison to reflect updated cursor behavior.


airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -528,35 +527,13 @@ def read_records(
...
-                # TODO Remove once all custom components implement `observe` method.
-                # https://github.com/airbytehq/airbyte-internal-issues/issues/6955
-                most_recent_record_from_slice = self._get_most_recent_record(
-                    most_recent_record_from_slice, current_record, _slice
-                )
                 yield stream_data
 
             if self.cursor:
-                self.cursor.close_slice(_slice, most_recent_record_from_slice)
+                self.cursor.close_slice(_slice)
```
brianjlai, line 533: when we were discussing this, you mentioned that this was needed to support custom cursor implementations, but now that we don't support anymore this would be safe to change.

My only other thought is that the cursor is removed whenever we call `ConcurrentDeclarativeSource._get_retriever()`. And the only case where we don't call this is for streams with a custom `Retriever` which automatically run as a synchronous stream. For this case are we okay not closing the slice with the most recent record?

brianjlai (commented):
Let's just chat about the one case I mentioned, I might not be understanding this final case quite right. Otherwise no other comments, but can 👍 after

PR #685, opened 2025-07-31
chore: remove unused availability strategy stuff

maxi297:
## What

We want to start migrating some streams from DeclarativeStream to DefaultStream. However, the check operation has not been implemented for DefaultStream. It seems like check and availability strategies are intertwined i.e. both the declarative and file-based stuff calls those. I wanted to clean this before actually implementing the check for DefaultStream to make sure we weren't migrating stuff that wasn't necessary.

### History Lesson

Availability strategies were initially added to be called during `READ` commands in order for on stream not to break all the other streams in the sync so we would execute the availability strategy (i.e. most of the time trying to fetch the first records) and if it wasn't working, we would simply skip this stream. This caused problematic (we needed to fetch some records that we would later fetch again during the actual sync) and could be solved by improving our error handling instead. Hence, we started deprecating this. However, it's been deprecated a couple of months after we started the concurrent framework so this was implemented partially there as well (see [this]([Stream.check_availability method](https://github.com/airbytehq/airbyte-python-cdk/blob/5fe69efea69593c3d55c8d6288d927c4ca838a09/airbyte_cdk/sources/streams/core.py#L199-L212))). Once the cleaning started on the legacy CDK, the concurrent part was never cleaned up. There was also a usage that has been added in the File Based CDK but this one seems legit so it seems we need to keep it. 

### What it means for this PR

* Removing concurrent CDK availability strategies from the concurrent stuff
    * This includes removing the counter part in the File Based CDK
* Remove the deprecation mention on the File Based CDK as it is used (rightfully so I guess) [here](https://github.com/airbytehq/airbyte-python-cdk/blob/e4cbaaf55eb099ff243e6e4b4b17f12b263d2277/airbyte_cdk/sources/file_based/file_based_source.py#L189)
* Keep the `HttpAvailabilityStrategy` as it is the default strategy for HttpStream. I assume this will be moved to a legacy package once we have the AbstractStream version of it and we are ready to do the breaking change

## Summary by CodeRabbit

* **Refactor**
  * Removed all availability strategy classes and related functionality from concurrent streaming modules.
  * Updated method signatures and removed deprecated decorators for improved consistency.
  * Eliminated unused classes and parameters to streamline codebase.

* **Tests**
  * Removed tests related to availability strategy functionality and updated test scenarios to reflect code changes.

* **Chores**
  * Added a minor comment for future reference in the codebase.


airbyte_cdk/sources/streams/concurrent/default_stream.py
```diff
@@ -23,7 +19,6 @@ def __init__(
         partition_generator: PartitionGenerator,
         name: str,
         json_schema: Mapping[str, Any],
-        availability_strategy: AbstractAvailabilityStrategy,
```
maxi297, line 26: This is in theory a breaking change but the only other usage I could find of DefaultStream is in source-salesforce but it uses `StreamFacade.create_from_stream` so we should be good

airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -179,7 +179,6 @@ def record_passes_validation_policy(self, record: Mapping[str, Any]) -> bool:
             )
 
     @cached_property
-    @deprecated("Deprecated as of CDK version 3.7.0.")
```
brianjlai, line 182: and here i was praying that i'd never have to see the words `availability_strategy` ever again

brianjlai (approved):
🪓

PR #697, opened 2025-08-05
chore: incremental without partition router as defaultstream

maxi297:
## What

Following https://github.com/airbytehq/airbyte-python-cdk/pull/691, this tackles streams that are incremental but do not have partition routing

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -1940,22 +1940,11 @@ class ModelToComponentFactory:
         combined_slicers = self._merge_stream_slicers(model=model, config=config)
 
         primary_key = model.primary_key.__root__ if model.primary_key else None
-        stop_condition_on_cursor = (
-            model.incremental_sync
-            and hasattr(model.incremental_sync, "is_data_feed")
-            and model.incremental_sync.is_data_feed
+
+        stream_slicer = self._build_stream_slicer_from_partition_router(
+            model.retriever, config, stream_name=model.name
         )
-        client_side_filtering_enabled = (
-            model.incremental_sync
-            and hasattr(model.incremental_sync, "is_client_side_incremental")
-            and model.incremental_sync.is_client_side_incremental
-        )
-        concurrent_cursor = None
-        if stop_condition_on_cursor or client_side_filtering_enabled:
-            stream_slicer = self._build_stream_slicer_from_partition_router(
-                model.retriever, config, stream_name=model.name
-            )
-            concurrent_cursor = self._build_concurrent_cursor(model, stream_slicer, config)
+        concurrent_cursor = self._build_concurrent_cursor(model, stream_slicer, config)
 
         if model.incremental_sync and isinstance(model.incremental_sync, DatetimeBasedCursorModel):
             cursor_model = model.incremental_sync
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2066,37 +2053,60 @@ def create_declarative_stream(
...
-            # * Streams without partition router but with cursor
+            # * Streams without partition router but with cursor. This is the `isinstance(concurrent_cursor, ConcurrentCursor)` condition
             # * Streams with both partition router and cursor
             # We specifically exclude parent streams here because SubstreamPartitionRouter has not been updated yet
             # We specifically exclude Connector Builder stuff for now as Brian is working on this anyway
+
             stream_name = model.name or ""
+            stream_slicer: ConcurrentStreamSlicer = concurrent_cursor
+            cursor: Cursor = FinalStateCursor(stream_name, None, self._message_repository)
+            if isinstance(retriever, AsyncRetriever):
```
maxi297, line 2074: It is kind of annoying to have to manage AsyncRetriever in a different way. Maybe once the migration is done we should find a better way. The next PR is about substream partition router and I assume this is the last bucket so I'll try to tackle this here

brianjlai (approved):

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2023,16 +1997,27 @@ def create_declarative_stream(
                 model=model.file_uploader, config=config
             )
 
+        # When constructing a declarative stream, we assemble the incremental_sync component and retriever's partition_router field
+        # components if they exist into a single CartesianProductStreamSlicer. This is then passed back as an argument when constructing the
```
tolik0, line 2001: I know this is an old comment, but are we actually creating a `CartesianProductStreamSlicer` here?
As far as I understand, we only create it when there’s a list of partition routers.
When we have an incremental sync, the iteration over the partition router is handled by the `PerPartitionCursor`, not by a `CartesianProductStreamSlicer`.

maxi297: This is indeed not true and was just copy/pasted from above. I'll clean this up

brianjlai:
> @brianjlai is there a reason why
> 
> https://github.com/airbytehq/airbyte-python-cdk/blob/addd4431f514cc135fd36eca8f2216db96eb8756/unit_tests/connector_builder/test_property_chunking.py#L208-L210
> 
> has not been updated to use the ConcurrentDeclarativeSource?

probably an oversight, but this is addressed in https://github.com/airbytehq/airbyte-python-cdk/pull/704/files#diff-a0cf4a1d832d23490722285fffb5964b6ea08d67941348be673a8fa3dec97768R221 which moves all the relevant tests to `ConcurrentDeclarativeSources`

unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -2440,8 +2432,8 @@ def test_default_schema_loader(self):
                 "cursor_granularity": "PT0.000001S",
             },
             None,
-            DatetimeBasedCursor,
-            DeclarativeStream,
+            ConcurrentCursor,
+            DefaultStream,
             id="test_create_simple_retriever_with_incremental",
```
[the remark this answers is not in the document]
maxi297: I'm confused by this comment. Line 922 is for test `test_stream_with_incremental_and_retriever_with_partition_router`, not `test_datetime_based_cursor`. The former indeed does `assert isinstance(stream, DeclarativeStream)` because it is a per partition incremental case

PR #699, opened 2025-08-05
fix(connector-builder): remerge concurrent cdk builder changes

brianjlai:
Re-adds two of the commits related to moving the connector builder to the concurrent CDK which were originally reverted in: https://github.com/airbytehq/airbyte-python-cdk/pull/694

One additional fix that is included is that we now consume the remaining records in the queue in `message_grouper.py` because we noticed an issue where we would end up in a deadlock on the syncing thread if the amount of records retrieved exceeded the queue size which defaulted to 10,000 items

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable concurrency and record limits for sources, allowing control over maximum records, slices, and streams during syncs.
  * Added support for injecting custom queues and improved message ordering in concurrent processing.
  * Enhanced logging and message repository integration for better visibility and debugging.
  * Added global record limit enforcement across partitions during data reads.

* **Bug Fixes**
  * Improved handling and formatting of state messages for downstream processing.
  * Ensured correct message ordering and state serialization in concurrent scenarios.

* **Tests**
  * Updated and added tests to validate new concurrency limits and message handling.
  * Removed outdated tests related to previous record limit implementations.
  * Adjusted tests to reflect changes in source creation signatures and cursor handling.

* **Chores**
  * Refactored imports, removed unused code, and clarified parameter usage for improved maintainability.
  * Updated documentation comments to clarify future intentions and code behavior.


airbyte_cdk/connector_builder/test_reader/message_grouper.py
```diff
@@ -95,7 +95,17 @@ def get_message_groups(
...
 
-    while records_count < limit and (message := next(messages, None)):
+    while message := next(messages, None):
+        # Even though we do not emit records beyond the limit in the message group response, we still
+        # need to process messages off the queue in order to avoid a deadlock occurs if the amount
+        # of extracted records exceeds the size of the queue (which has a default of 10,000)
+        #
+        # A few other options considered was killing the thread pool, but that doesn't kill in-progress
+        # in-progress threads. We also considered adding another event to the main queue, but this is
+        # the simplest solution for the time being.
```
lmossman, line 105:
```suggestion
        # Even though we do not emit records beyond the limit in the message group response, we still
        # need to process messages off the queue in order to avoid a deadlock that occurs if the amount
        # of extracted records exceeds the size of the queue (which has a default of 10,000)
        #
        # A few other options considered was killing the thread pool, but that doesn't kill in-progress
        # threads. We also considered adding another event to the main queue, but this is
        # the simplest solution for the time being.
```

brianjlai:
Most of the changes to support limiting at the global level will be in `airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py`

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -70,30 +95,50 @@ def __init__(
...
     ) -> None:
         # todo: We could remove state from initialization. Now that streams are grouped during the read(), a source
         #  no longer needs to store the original incoming state. But maybe there's an edge case?
         self._connector_state_manager = ConnectorStateManager(state=state)  # type: ignore  # state is always in the form of List[AirbyteStateMessage]. The ConnectorStateManager should use generics, but this can be done later
 
+        # We set a maxsize to for the main thread to process record items when the queue size grows. This assumes that there are less
+        # threads generating partitions that than are max number of workers. If it weren't the case, we could have threads only generating
+        # partitions which would fill the queue. This number is arbitrarily set to 10_000 but will probably need to be changed given more
+        # information and might even need to be configurable depending on the source
+        queue: Queue[QueueItem] = Queue(maxsize=10_000)
```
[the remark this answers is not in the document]
brianjlai: out of scope

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
```diff
@@ -46,17 +57,24 @@ def __init__(
...
         self._stream_name = stream_name
         self._json_schema = json_schema
         self._retriever = retriever
         self._message_repository = message_repository
+        self._max_records_limit = max_records_limit
         self._stream_slice = stream_slice
         self._hash = SliceHasher.hash(self._stream_name, self._stream_slice)
 
     def read(self) -> Iterable[Record]:
         for stream_data in self._retriever.read_records(self._json_schema, self._stream_slice):
```
maxi297, line 72: I'm not sure anymore how generator works but it feels like this will perform an HTTP request even if we hit the limit before processing the slice. Should we have a `if total_record_counter >= self._max_records_limit:` condition before the for loop as well?

maxi297 (approved):
This makes sense. I only have one concern left but I'll approve

airbyte_cdk/sources/declarative/stream_slicers/declarative_partition_generator.py
brianjlai: yep makes sense, that should proactively stop from making the first request. i'll fix and add a test for this

PR #703, opened 2025-08-07
feat: pass transformations to retriever instance

maxi297:
## What

Currently if we wants transformations to be passed to the record selector when there is a custom retriever, we need to do [this](https://github.com/airbytehq/airbyte/blob/6d42d30fb64970608e212ba24bf0471dbf2078d7/airbyte-integrations/connectors/source-google-ads/source_google_ads/components.py#L431). 

## How

The ModelToComponentFactory will match function arguments not only on model parameters but also on kwargs being passed. If there is a collision, model parameters will be prioritize to keep backward compatibility


brianjlai (approved):

maxi297:
Should be covered by https://github.com/airbytehq/airbyte-python-cdk/pull/707/commits/25ca5b8ae688c5a361cb7834f7a4de00b4d90927

PR #704, opened 2025-08-08
chore(concurrent_declarative_source): Integration ManifestDeclarativeSource into ConcurrentDeclarativeSource

brianjlai:
With the change to move the connector builder to use the concurrent CDK, there is no longer a need to have the decoupled ManifestDeclarativeSource and ConcurrentDeclarativeSource.

To simplify our inheritance model, what we ideally want is a singluar ConcurrentDeclarativeSource which implements the Source interface. However, because we still have a few remaining legacy flows, we will have to for the time continue to implement `AbstractSource` and once finished with our refactor, replace it once and for all.

## A few notes on specific parts of this refactor

-


unit_tests/connector_builder/test_property_chunking.py
```diff
@@ -196,20 +200,33 @@
 
 @freezegun.freeze_time(f"{FREEZE_DATE}T00:00:00Z")
 def test_read():
+    # We need a state time earlier than the current time otherwise we will not generate any date range
+    # partitions to process on the concurrent engine
+    day_before = (ab_datetime_parse(f"{FREEZE_DATE}T00:00:00Z") - timedelta(days=1)).strftime(
```
brianjlai, line 205: just adding a comment here to highlight the above code comment.

There are a handful of these fixes in the tests which are a product of using concurrent_cursors instead legacy declarative cursors which differed in behavior

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -2108,3 +2131,2640 @@ def get_states_for_stream(
...
+                    "type": "Spec",
+                },
+            },
+            (
+                _create_page({"rates": [{"ABC": 0}, {"AED": 1}], "_metadata": {"next": "next"}}),
+                _create_page({"rates": [{"USD": 2}], "_metadata": {}}),
+            )
+            * 10,
+            [{"ABC": 0}, {"AED": 1}, {"USD": 2}],
+            [
```
maxi297, line 3809: This has changed from:
```
            [
                call({}, {}, None),
                call(
                    {"next_page_token": "next"},
                    {"next_page_token": "next"},
                    {"next_page_token": "next"},
                ),
            ],
```

Can you give me the TLDR as to why the pagination token is only sent on the last request now?

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -2108,3 +2131,2640 @@ def get_states_for_stream(
...
+
+    ConcurrentDeclarativeSource(
+        source_config=manifest,
+        config=input_config,
+        config_path="/fake/config/path",
+        catalog=create_catalog("lists"),
+        state=None,
+    )
+
+    migration_mocks["open"].assert_called_once_with("/fake/config/path", "w")
```
maxi297, line 4483: `test_given_unmigrated_config_when_migrating_then_config_is_migrated` does not validate `migration_mocks["message_repository"].emit_message.assert_called_once()` anymore which seems important to confirm that the config has been migrated. Can you explain why this was removed?

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -2108,3 +2131,2640 @@ def get_states_for_stream(
         for message in messages
         if message.state and message.state.stream.stream_descriptor.name == stream_name
     ]
+
+
+# The tests below were originally written to test the ManifestDeclarativeSource class. However,
+# after deprecating the class when migrating away from legacy synchronous CDK flow, the tests
+# were adjusted to validate ConcurrentDeclarativeSource.
```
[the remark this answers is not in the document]
maxi297: Those are the diffs I could identify:
* If schema can't be loaded, it'll fail early. This is interesting because it means that a stream that needs to perform a HTTP request to know it's schema will do so even if it's not in the catalog. We already have the issue [here](https://github.com/airbytehq/airbyte-python-cdk/blob/3244eec21920b88cb26ce9efd5d31395b59407c9/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L324) so I'm fine with merging this PR without this change. I assume that [this change](https://github.com/airbytehq/airbyte-python-cdk/pull/691/files#diff-34b63829c6602d2a94cfb3554f520b810c357b96bc44fd7dae2a245f57870e36R334) will fix the issue.
* One change regarding pagination that I'll ask for more information in a comment below
* State is not passed to `SimpleRetriever._fetch_next_page` which means that state interpolation will not work anymore for HTTP requests. Have we confirmed that we don't have certified connectors that do this anymore? 
* One change regarding config being emitted that I'll ask for more information in a comment below

Do you think of other changes I might have missed?

unit_tests/__init__.py
```diff
@@ -1,7 +1,9 @@
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+
 # THIS STOPS SOME MODELS TESTS FROM FALLING OVER. IT'S A HACK, WE SHOULD PIN DOWN WHAT'S ACTUALLY GOING ON HERE
 
 # Import the thing that needs to be imported to stop the tests from falling over
-from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+# from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
```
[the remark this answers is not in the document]
brianjlai: oh yeah oops, i comment out to make sure tests could still run, but intended to get rid of. removing

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -169,19 +276,146 @@ def __init__(
...
+
+        This property provides access to the internal source configuration as a mapping,
+        which contains all settings and parameters required to define the source's behavior.
+
+        Returns:
+            Mapping[str, Any]: The resolved source configuration manifest.
+        """
+        return self._source_config
+
+    @property
```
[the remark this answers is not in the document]
brianjlai: yep that's correct. We would deprecate the external method since we can access the field directly from in here.

I'll add a comment here and in advance of removing the method, from within this class, I'll just directly reference `self._message_repository()` since its equivalent

unit_tests/sources/declarative/test_concurrent_declarative_source.py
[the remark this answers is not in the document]
brianjlai: > If schema can't be loaded, it'll fail early. This is interesting because it means that a stream that needs to perform a HTTP request to know it's schema will do so even if it's not in the catalog. We already have the issue here so I'm fine with merging this PR without this change. I assume that this change will fix the issue.

Am I following correctly, that the key change is `DeclarativePartitionFactory` uses `schema_loader=declarative_stream._schema_loader` instead of using `get_json_schema()` and we're covered because of the fix already merged that does so in both cases highlighted in https://github.com/airbytehq/airbyte-python-cdk/pull/691 . So no change on this side

> State is not passed to SimpleRetriever._fetch_next_page which means that state interpolation will not work anymore for HTTP requests. Have we confirmed that we don't have certified connectors that do this anymore?

Is this new to this PR that we no longer pass state? I didn't change the simple retriever. But regardless, the answer is no. We made a big push at the beginning of this year when we were deprecating custom cursors, we also removed usage of `stream_state` as a jinja context because it's not thread safe. We have one usage on a community connector exchange rates. So none of our certified rely on a non-thread safe `stream_state` value.

Addressed in respective comments:

> One change regarding pagination that I'll ask for more information in a comment below

commented below. The test hasn't actually changed, but i'm not rebased off latest master so it looks like there's a change

> One change regarding config being emitted that I'll ask for more information in a comment below

discussed there

And yeah what you mentioned tracks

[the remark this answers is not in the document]
maxi297: Ok, I'm all good with these. Thanks Brian!

unit_tests/legacy/sources/declarative/partition_routers/test_parent_state_stream.py
```diff
@@ -8,6 +8,9 @@
 import pytest
```
maxi297, line 8: Regarding the renaming of this file, is this really legacy? We're still building manifest that can create non-legacy components. Also, this test will inevitably fail when moving all the streams to DefaultStream because the read will go through `AbstractSource.read` and it will call `streams` that will return `DefaultStream` instead of `DeclarativeStream` although the AbstractSource will use the `DeclarativeStream` interface.

What do you think we should do about this?

maxi297 (commented):
One comment regarding a test that was moved to legacy

unit_tests/legacy/sources/declarative/partition_routers/test_parent_state_stream.py
brianjlai: > Also, this test will inevitably fail when moving all the streams to DefaultStream because the read will go through AbstractSource.read and it will call streams that will return DefaultStream instead of DeclarativeStream although the AbstractSource will use the DeclarativeStream interface.

There might be edge cases around this in how we do this w/o breaking interfaces midway through the migration, but we both agree that `AbstractSource` should end up deprecated and the remaining `Source` if needed just references `AbstractStream`s

And once `AbstractSource` does inherit `Source` (an by extension ConcurrentDeclarativeSource directly inherits `Source`), we would just leave the `AbstractSource.streams()` to return `List[Stream]` which is deprecated too. Maybe we can sync on this early next week, but in the long term, we would want `AbstractSource` to be deprecated, and if that were the case, we would just leave `AbstractSource.read()` alone since when everything uses `ConcurrentSource.read()` there shouldn't be any flows on the legacy path.

And as for why I considered this file to be legacy was because the test cases are covered under the `test_concurrent_perpartitioncursor.py` which instantiates `ConcurrentDeclarativeSource` from a manifest with a similar level of coverage or potentially better since it was designed to test this when per partition cursors were be built

airbyte_cdk/connector_builder/connector_builder_handler.py
```diff
@@ -90,7 +88,7 @@ def create_source(
 
 
 def read_stream(
-    source: DeclarativeSource,
+    source: ConcurrentDeclarativeSource[Optional[List[AirbyteStateMessage]]],
```
maxi297, line 91: out-of-scope for this PR: I always found the generic for `ConcurrentDeclarativeSource` dangerous and since `TState` is always `Optional[List[AirbyteStateMessage]]` and we assume so at a couple of places like [here](https://github.com/airbytehq/airbyte-python-cdk/blob/f525803b3fec9329e4cc8478996a92bf884bfde9/airbyte_cdk/sources/declarative/concurrent_declarative_source.py#L109). I'm wondering why it was put there in the first place and if we should clean that up as well

unit_tests/legacy/sources/declarative/partition_routers/test_parent_state_stream.py
maxi297: Yes. I think this one in particular is an easy one because everything it tests seems to be covered in the Concurrent CDK by `test_concurrent_perpartitioncursor.py`. There might be the question as to is we rewrite these if they don't have equivalent and if we just lose test coverage

maxi297 (approved):
LGTM! 🚢

airbyte_cdk/connector_builder/connector_builder_handler.py
brianjlai: yeah that's a good question. i can't remember why I wrote it this way when we built the class, but a part of me would think I wouldn't introduce generics unless it was absolutely necessary for mypy issues... I'm gonna add a note because I agree unless it's strictly needed, it does more harm than good

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -502,6 +864,127 @@
...
+    def _dynamic_stream_configs(
+        self,
+        manifest: Mapping[str, Any],
+        config: Mapping[str, Any],
+        with_dynamic_stream_name: Optional[bool] = None,
+    ) -> List[Dict[str, Any]]:
+        dynamic_stream_definitions: List[Dict[str, Any]] = manifest.get("dynamic_streams", [])
+        dynamic_stream_configs: List[Dict[str, Any]] = []
+        seen_dynamic_streams: Set[str] = set()
+
```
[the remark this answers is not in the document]
brianjlai: maybe, but out of scope for this PR

PR #719, opened 2025-08-22
feat: add manifest server

pedroslopez:
Introduces an HTTP interface for running declarative manifests, manifest-sererv. This will be used as a more lightweight and cdk-native replacement of the connector-builder-server that is currently a part of the Platform. More than just a connector-builder-specific server, this is structured as a more generic interface that can be leveraged outside of the platform itself, and have more flexible deployment options.

### Endpoints

This first iteration includes the following endpoints:

- `/v1/manifest/test_read` - Test reading from a specific stream in the manifest. Meant for use in the connector builder or to preview a stream.
- `/v1/manifest/check` - Check configuration against a declarative source.
- `/v1/manifest/discover` - Discover streams from a declarative source.

Connector builder / Manifest-composition helpers:
- `/v1/manifest/resolve` - Resolve a manifest to its final configuration.
- `/v1/manifest/full_resolve` - Fully resolve a manifest including dynamic streams.

#### Authentication

JWT-based authentication can be optionally enabled. This uses the same internal auth mechanism that airbyte-platform uses to make it easy to plug into an airbyte instance. To enable, `AB_JWT_SIGNATURE_SECRET` must be set, with which the JWT signatures will be validated. 

#### Custom Components

Custom python component support is off by default. To enable custom python components, set `AIRBYTE_ENABLE_UNSAFE_CODE=true`

### CLI

To make it easy to run the server and perform actions, a CLI command is exposed for the manifest sererv. 

#### OpenAPI

The platform will use the openapi file in this repo to generate an API client to talk to the manifest server. For this reason, the `openapi.yaml` file is checked into the repo. The file can be regenerated by using the cli.

## Summary by CodeRabbit

- New Features
  - Introduced a Manifest Server with endpoints for health, capabilities, and manifest operations (test_read, check, discover, resolve, full_resolve).
  - Optional JWT authentication via environment secret.
  - New CLI: manifest-server (start, info, generate-openapi).
  - OpenAPI spec generation and included YAML.
  - Dockerfile added with multi-arch image support; default container port 8080.

- Documentation
  - Added README covering installation, usage, endpoints, auth, and Docker instructions.

- Chores
  - New GitHub Action to prevent Docker tag overwrites.
  - CI workflows updated to build and publish multi-arch images and enforce tag checks.

- Tests
  - Added extensive unit tests for routers, auth, processor, and utilities.


.github/workflows/pypi_publish.yml
```diff
@@ -6,6 +6,8 @@
 # we have to also update the Trusted Publisher settings on PyPI.
 
 name: CDK Publish
+permissions:
+  contents: read
```
aaronsteers, line 10: @pedroslopez - I might be wrong on this, but I think we also need to add deployments: write, since this workflow publishes deployment records to named "environment: ..." maps in the yaml.

```suggestion
  contents: read
  deployments: write
```

https://chatgpt.com/share/68acb07d-c1fc-8004-9a79-d41436af4920

If you try to publish without this set, you should see a permission failure if my assertion is correct. It's a bit confusing because of the interplay between "environments" and "deployments".

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

pedroslopez: Hmm, I ran this manually and it seems like it worked https://github.com/airbytehq/airbyte-python-cdk/actions/runs/17219503797

where is this deployment record created?

airbyte_cdk/manifest_server/cli/_openapi.py
```diff
@@ -0,0 +1,43 @@
...
+
+@click.command("generate-openapi")
+@click.option(
+    "--output",
+    "-o",
+    default="airbyte_cdk/manifest_server/openapi.yaml",
+    help="Output path for the OpenAPI YAML file",
+    show_default=True,
+)
+def generate_openapi(output: str) -> None:
```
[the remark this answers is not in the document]
pedroslopez: Oh cool - just added the poe task and added as part of `build`. However, `assemble` is failing for me locally so wasn't able to test the combined one (the standalone one works for me though)

pedroslopez:
@aaronsteers I think this is ready for another look. I've made some changes to use the Concurrent source as well.

airbyte_cdk/manifest_server/routers/manifest.py
```diff
@@ -0,0 +1,158 @@
...
+    runner = ManifestCommandProcessor(source)
+    catalog = runner.discover(request.config.model_dump())
+    if catalog is None:
+        raise HTTPException(status_code=422, detail="Connector did not return a discovered catalog")
+    return DiscoverResponse(catalog=catalog)
+
+
+@router.post("/resolve", operation_id="resolve")
+def resolve(request: ResolveRequest) -> ManifestResponse:
+    """Resolve a manifest to its final configuration."""
```
aaronsteers, line 128: Nit: I understand "full resolve" expands dynamic schemas, but I still don't fully understand what this step does, even after working with it for a while.

What do you think of adding to the docstring an example of what gets resolved in this step and when this should be called? Also could be helpful to explain if the resolved version of the manifest replaces the original, or is it just used for json schema validation or other preprocessing needed for json schema validation, and/or something else? Thanks!

airbyte_cdk/manifest_server/routers/manifest.py
```diff
@@ -0,0 +1,158 @@
...
+
+router = APIRouter(
+    prefix="/manifest",
+    tags=["manifest"],
+    dependencies=[Depends(verify_jwt_token)],
+)
+
+
+@router.post("/test_read", operation_id="testRead")
+def test_read(request: StreamTestReadRequest) -> StreamRead:
```
aaronsteers, line 69: What do you think of a return class name of "StreamReadResponse" to match the naming convention used on the other routes?

airbyte_cdk/manifest_server/command_processor/processor.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> Tuple[bool, str]:
+        """
+        Check the connection to the source.
+        """
+
+        spec = self._source.spec(self._logger)
+        messages = AirbyteEntrypoint(source=self._source).check(spec, config)
+        messages_by_type = self._get_messages_by_type(messages)
+        self._raise_on_trace_message(messages_by_type)
+        connection_status = self._get_connection_status(messages_by_type)
```
aaronsteers, line 75: I believe `EntryPointOutput` already implements most (maybe all?) of these methods.

https://airbytehq.github.io/airbyte-python-cdk/airbyte_cdk/test/entrypoint_wrapper.html#EntrypointOutput

Image

airbyte_cdk/manifest_server/command_processor/processor.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def _get_connection_status(
+        self,
+        messages_by_type: Mapping[str, List[AirbyteMessage]],
+    ) -> Optional[AirbyteConnectionStatus]:
+        """
+        Get the connection status from the messages.
+        """
+        messages = messages_by_type.get(AirbyteMessageType.CONNECTION_STATUS)
+        return messages[-1].connectionStatus if messages else None
```
aaronsteers, line 117: I think this implementation can be replaced by a call to `output.connection_status_messages`.

https://airbytehq.github.io/airbyte-python-cdk/airbyte_cdk/test/entrypoint_wrapper.html#EntrypointOutput.connection_status_messages

airbyte_cdk/manifest_server/command_processor/processor.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def _get_catalog(
+        self,
+        messages_by_type: Mapping[str, List[AirbyteMessage]],
+    ) -> Optional[AirbyteCatalog]:
+        """
+        Get the catalog from the messages.
+        """
+        messages = messages_by_type.get(AirbyteMessageType.CATALOG)
+        return messages[-1].catalog if messages else None
```
aaronsteers, line 127: `AirbytEntrypoint.catalog`: https://airbytehq.github.io/airbyte-python-cdk/airbyte_cdk/test/entrypoint_wrapper.html#EntrypointOutput.catalog

airbyte_cdk/manifest_server/command_processor/processor.py
aaronsteers, line 1: (Non-blocking, long-term design question:)

This `airbyte_cdk.manifest_server.command_processor` module (`util.py` and `processor.py`) appears to contain business logic and execution logic that should probably be common and shared also by CLI invocations as well as the REST invocations. The only piece that appears different is the wrapping of errors as HTTPException, which could live in a helper function specific to the REST implementation.

There are actually three or four implementations that share the same requirements for calling the same/similar test read, validate, logic as of now:
1. Current UI implementation - which calls the python module with the `python -m` arg IIRC.
2. This new REST implementation, replaces (1).
3. The new Connector Builder MCP, which needs all the same functions for local dev/test by the LLM agent.
4. Local automated testing use cases - where we'd pass all connector manifests to the latest engine to confirm they can validate and run when we release new CDK versions.

It feels like the right approach is probably to keep the business logic for how to run/validate/test/process/resolve manifests into a common public module like `airbyte_cdk.manifest_processor` (or similar), and have the various implementations (MCP, CLI, and REST) each call to that library for their processing and business logic. 

Probably more than we want to tackle in this PR (hence my note that this should be non-blocking), but wdyt?

aaronsteers (approved):
@pedroslopez - Really nice work on this! I'm approving to unblock merging today if ready, with just one change I'd like to see in this iteration before merging: specifically my inline comments to see if we can reduce the amount of code to be maintained by leveraging existing methods on `EntrypointOutput`:

https://airbytehq.github.io/airbyte-python-cdk/airbyte_cdk/test/entrypoint_wrapper.html#EntrypointOutput

My other comment is in regards to having a single module (outside the server module) where local CLI and MCP use cases could use the same logic. The MCP code ([here](https://github.com/airbytehq/connector-builder-mcp/blob/main/connector_builder_mcp/validation_testing.py#L12C6-L19)) as of now makes heavy use of the `airbyte_cdk.connector_builder.connector_builder_handler` module ([pdoc link here](https://airbytehq.github.io/airbyte-python-cdk/airbyte_cdk/connector_builder/connector_builder_handler.html)). Where possible (at least in the long run), we should probably see about centralizing common logic into that module or another one shared by the different implementations.

airbyte_cdk/manifest_server/command_processor/processor.py
pedroslopez: We can talk more about this separately - generally I do agree with the sentiment that there should be one way to run things that is shared, though we should address that separately. I'm not sure I'm familiar with the "cli invocations" you're referring to so maybe it's worth a chat.

For whatever it's worth I think leveraging the EntrypointOutput and unifying around that moves us towards that goal, though perhaps the actual calling of the Entrypoint / building the source could also be unified.

airbyte_cdk/manifest_server/routers/manifest.py
pedroslopez: To be honest I also don't fully have my head around when this is necessary lol. I added to the docstring with the explanation Lake gave me. It seems like this is only called from the builder in a specific case.

PR #721, opened 2025-08-22
feat: add suggestions for datetime based cursor fields

lmossman:
Once https://github.com/airbytehq/airbyte-platform-internal/pull/17519 is merged, the Connector Builder will automatically render a list of values when `suggestions` is set on a `string` or `array<string>` field.

This PR adds common values to the `suggestions` properties of various DatetimeBasedCursor fields, as these tend to be confusing for users to configure and having suggestions makes this easier.


ChristoGrab (approved):
:shipit:

PR #725, opened 2025-08-26
fix: monkey patch SQLiteDict in order to fix sqlite3.InterfaceError

maxi297:
## What

https://airbytehq-team.slack.com/archives/C0980AJD0F7/p1756166020078259?thread_ts=1756138243.067239&cid=C0980AJD0F7

https://github.com/airbytehq/airbyte-python-cdk/actions/runs/17223546541/job/48863700904?pr=719#step:8:33688

## How

There is no version of requests_cache that got release with this fix so we monkey_patch stuff until there is a new release

I've added a test so that when the version gets updated, we can re-assess what is the next step

## Summary by CodeRabbit

- Bug Fixes
  - Resolved intermittent cache read errors that could cause HTTP-based syncs to fail.
  - Improved reliability when accessing the local cache, reducing unexpected exceptions during runs.

- Chores
  - Applied a temporary workaround to ensure stable cache behavior without requiring user configuration changes.

- Tests
  - Added a safeguard test to detect incompatible cache library versions early.

### DIFF
--- a/unit_tests/test_requests_cache_monkeypatch_version.py
+++ b/unit_tests/test_requests_cache_monkeypatch_version.py
@@ -1,0 +1,11 @@
+import requests_cache
+
+
+def test_assert_requests_cache_version():
+    """
+    We need to be alerted once the requests_cache version is updated. The reason is that we monkey patch one of the
+    method in order to fix a bug until a new version is released.
+
+    For more information about the reasons of this test, see monkey_patched_get_item in http_client.py
+    """
+    assert requests_cache.__version__ == "1.2.1"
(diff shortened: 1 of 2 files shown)


unit_tests/test_requests_cache_monkeypatch_version.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_assert_requests_cache_version():
+    """
+    We need to be alerted once the requests_cache version is updated. The reason is that we monkey patch one of the
+    method in order to fix a bug until a new version is released.
+
+    For more information about the reasons of this test, see monkey_patched_get_item in http_client.py
+    """
+    assert requests_cache.__version__ == "1.2.1"
```
brianjlai, line 11: i hate that we have to do this, but i love that there's a test to tell us to fix the `requests_cache`'s bad implementation whenever it becomes available

brianjlai (approved):
pain levain

PR #729, opened 2025-08-27
fix(excel-parser): Fix memory issue in `ExcelParser`

tolik0:
## Summary by CodeRabbit

- **Refactor**
  - Optimized Excel record parsing to stream rows incrementally, reducing memory usage and improving performance on large files.
  - Maintains the same output format and row order to ensure compatibility with existing integrations.
  - Enhances stability during long-running or large dataset syncs by avoiding large in-memory conversions.


maxi297 (commented):
The change makes sense as we won't pull everything in memory to yield it after. However, I have one small comment regarding copilot's suggestion

PR #732, opened 2025-08-28
fix(manifest-server): reset global record counter before test reads

pedroslopez:


pedroslopez:
> Could we have a test that does two READ operation with test limits and both returns records?
@maxi297 Added this at the declarative partition factory level. I tried to add higher up (ConcurrentDeclarativeSource, TestReader, etc) but all the mocking that happens makes it quite hard to do at that level. I think this fulfills the goal though.

maxi297 (approved):
LGTM!

PR #735, opened 2025-08-28
feat: add robust datetime parsing fallback to DatetimeBasedCursor

aaronsteers:
# feat: add robust datetime parsing fallback to DatetimeBasedCursor

## Summary

This PR enhances the DatetimeBasedCursor's datetime parsing to eliminate parsing errors for valid datetime formats while maintaining full backward compatibility. The implementation adds `ab_datetime_try_parse` as a fallback when the expected `cursor_datetime_formats` fail to parse a datetime value.

**Key changes:**
- Enhanced `parse_date()` method to use robust fallback parsing via `ab_datetime_try_parse`
- Updated schema documentation to clarify the new fallback behavior  
- Fixed test case that expected ValueError for valid ISO datetime (now uses truly unparseable string)
- Preserved original cursor value format in state storage for backward compatibility

**Behavior:** The cursor now tries user-specified formats first (as before), but falls back to robust parsing that handles most ISO8601/RFC3339 compliant formats, Unix timestamps, and other common datetime formats. This should eliminate the majority of datetime parsing errors while keeping the same performance for connectors using expected formats.

**High Priority (3 items):**

- [ ] **Test backward compatibility** - Verify existing connectors with DatetimeBasedCursor still work correctly, especially those with custom `cursor_datetime_formats`
- [ ] **Test fallback parsing** - Verify the robust fallback successfully parses various ISO8601/RFC3339 formats, Unix timestamps, and edge cases that previously failed
- [ ] **Verify state storage behavior** - Confirm cursor state preservation works as expected and doesn't introduce format inconsistencies

**Recommended end-to-end test plan:**
1. Test source-recurly and source-instagram (which use DatetimeBasedCursor) to ensure no regressions
2. Try parsing datetime values in formats not listed in `cursor_datetime_formats` to verify fallback works
3. Check that cursor state values maintain their original format after syncing

### Notes

- All 57 DatetimeBasedCursor unit tests pass
- Linting, formatting, and type checking all pass  
- Changes are minimal and focused to reduce review complexity
- Requested by AJ Steers (@aaronsteers) in Devin session: https://app.devin.ai/sessions/18f0d77923df402bab74cda8d9108655
- This is the first PR for CDK changes; a second PR with example connector updates will follow

### DIFF
--- a/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
+++ b/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
@@ -21,6 +21,7 @@ from airbyte_cdk.sources.declarative.requesters.request_option import (
 )
 from airbyte_cdk.sources.message import MessageRepository
 from airbyte_cdk.sources.types import Config, Record, StreamSlice, StreamState
+from airbyte_cdk.utils.datetime_helpers import ab_datetime_format, ab_datetime_try_parse
 from airbyte_cdk.utils.mapping_helpers import _validate_component_request_option_paths
 
 
@@ -313,6 +314,11 @@ class DatetimeBasedCursor(DeclarativeCursor):
                 return self._parser.parse(date, datetime_format)
             except ValueError:
                 pass
+
+        parsed_dt = ab_datetime_try_parse(date)
+        if parsed_dt:
+            return parsed_dt
+
         raise ValueError(f"No format in {self.cursor_datetime_formats} matching {date}")
 
     @classmethod
--- a/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -944,7 +944,9 @@ definitions:
         items:
           type: string
         description: |
-          The possible formats for the cursor field, in order of preference. The first format that matches the cursor field value will be used to parse it. If not provided, the Outgoing Datetime Format will be used.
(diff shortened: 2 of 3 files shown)


airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -313,6 +314,10 @@ def parse_date(self, date: str) -> datetime.datetime:
                 return self._parser.parse(date, datetime_format)
             except ValueError:
                 pass
+
+        if parsed_dt is not None:
```
aaronsteers, line 318:
```suggestion
        parsed_dt = ab_datetime_try_parse(date)
        if parsed_dt is not None:
```

airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -313,6 +314,11 @@ def parse_date(self, date: str) -> datetime.datetime:
                 return self._parser.parse(date, datetime_format)
             except ValueError:
                 pass
+
+        parsed_dt = ab_datetime_try_parse(date)
```
aaronsteers, line 318:
```suggestion
        # If we have not parsed by now, use the robust parser which handles
        # all common formats, including all formats supported by ISO8601
        # and RFC3306.
        parsed_dt = ab_datetime_try_parse(date)
```

PR #738, opened 2025-08-29
fix(concurrent_declarative_source): set `None` defaults for optional constructor args, remove unnecessary `Generic` implementation

aaronsteers:
# fix(concurrent_declarative_source): add default value to optional state parameter

## Summary
Fixes a breaking change introduced in PR #704 where the `state` parameter in `ConcurrentDeclarativeSource.__init__()` was typed as `Optional[List[AirbyteStateMessage]]` but lacked a default value, causing failures when `create_source()` was called with (implicit) `state=None`.

**Key changes:**
- Added `= None` default value to the `state` parameter
- Reordered constructor parameters to move `state` after `source_config` (required for Python syntax)
- Removed `Generic[TState]` from class definition as it provided no meaningful value (always the same concrete type)
- Updated all type annotations across 8 files to use concrete `ConcurrentDeclarativeSource` type
- Fixed test parameter order to match updated constructor signature

- [ ] **Verify no other direct calls to `ConcurrentDeclarativeSource` constructor are broken** - The parameter reordering from `(catalog, config, state, source_config)` to `(catalog, config, source_config, state=None)` could break existing code
- [ ] **Test `create_source()` with minimal arguments** - Ensure `create_source(config=config, limits=limits, catalog=None, state=None)` works without errors
- [ ] **Verify type checking passes in dependent projects** - Removing `Generic[TState]` changes the public API surface and could affect external CDK consumers

### Notes
- The existing test `test_create_source()` in `test_connector_builder_handler.py` already validates the minimal argument case and should now pass
- All linting, formatting, MyPy checks, and pytest-fast tests pass locally
- Requested by AJ Steers (@aaronsteers) via Devin session: https://app.devin.ai/sessions/31e3ee31b0494bd6acda7823111e3ae1

## Summary by CodeRabbit

* **New Features**
  * Added optional settings for manifest normalization, operational limits, and a configuration path when creating sources.

* **Refactor**
  * Simplified type annotations and standardized source initialization order; no runtime behavior changes.

* **Tests**
  * Updated tests to match the new initialization signature; no functional impact for users.

@coderabbit ignore


airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -171,8 +171,8 @@ def __init__(
         self,
         catalog: Optional[ConfiguredAirbyteCatalog],
         config: Optional[Mapping[str, Any]],
-        state: TState,
         source_config: ConnectionDefinition,
+        state: Optional[List[AirbyteStateMessage]] = None,
```
aaronsteers, line 175: aside - 
Noting here: arg order had to change because `source_config` (aka the manifest) _is_ required and we can't have required positional args after non-required ones.

Ideally, we'd use "*," to force these all as keyword-only args, but that seems overkill here for what we're doing. Note though that if any callers are passing a positional-only list of args here, those calls could break.

ChristoGrab (approved):
Discussed offline, :shipit:

PR #742, opened 2025-08-29
fix: make concurrent declarative source test order-agnostic

aaronsteers:
# fix: make concurrent declarative source test order-agnostic

## Summary
Fixes a flaky test failure in `test_read_concurrent_declarative_source` that was occurring on Python 3.13 due to race conditions in concurrent processing. The test was using strict list equality (`assert output_data == expected_records`) which failed when concurrent threads returned records in different orders.

The fix replaces the strict equality check with a sorted comparison that validates the same records are present without caring about their order. This is appropriate for concurrent processing where order is not guaranteed.

The failing test case was: `test_no_pagination_with_partition_router-manifest4-pages4-expected_records4-expected_calls4`

- [ ] **Verify sorting key works for all parametrized test cases** - The change affects all test cases in the parametrized test, not just the failing partition router one. Check that the sorting key `(partition, ABC, AED)` with default values of 0 is appropriate for all test scenarios.
- [ ] **Run the full parametrized test suite** - Execute `pytest unit_tests/sources/declarative/test_concurrent_declarative_source.py::test_read_concurrent_declarative_source -v` to ensure all test cases still pass.
- [ ] **Consider scope appropriateness** - Evaluate if this change is too broad. The original issue was specific to partition router tests, but this change affects all concurrent declarative source tests.

### Notes
- Successfully tested the specific failing test case and it now passes
- Lint checks pass with no issues  
- This addresses the race condition mentioned in the Slack thread where Brian Lai noted "ordering should not matter in the case of this test"

**Link to Devin run**: https://app.devin.ai/sessions/3251fb4d62c54774ab4d264d4ffe8b9a  
**Requested by**: @aaronsteers

## Summary by CodeRabbit

* **Tests**
  * Made concurrency-related tests order-insensitive so they validate content regardless of record ordering.
  * Switched assertions to compare normalized, order-agnostic record sets to prevent flaky failures from non-deterministic ordering.
  * Improves CI stability and developer confidence with no changes to user-facing behavior.

### DIFF
--- a/unit_tests/sources/declarative/test_concurrent_declarative_source.py
+++ b/unit_tests/sources/declarative/test_concurrent_declarative_source.py
@@ -3885,7 +3885,7 @@ def test_read_concurrent_declarative_source(
         output_data = [
             message.record.data for message in _run_read(manifest, _stream_name) if message.record
         ]
-        sort_key = lambda x: (x.get("partition", 0), x.get("ABC", 0), x.get("AED", 0))
+        sort_key = lambda x: (x.get("partition", 0), x.get("ABC", 0), x.get("AED", 0), x.get("USD", 0))
         assert sorted(output_data, key=sort_key) == sorted(expected_records, key=sort_key)
         mock_retriever.assert_has_calls(expected_calls)
 


unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -3885,7 +3885,12 @@ def test_read_concurrent_declarative_source(
         output_data = [
             message.record.data for message in _run_read(manifest, _stream_name) if message.record
         ]
-        assert output_data == expected_records
+        assert sorted(
+            output_data, key=lambda x: (x.get("partition", 0), x.get("ABC", 0), x.get("AED", 0))
+        ) == sorted(
+            expected_records,
+            key=lambda x: (x.get("partition", 0), x.get("ABC", 0), x.get("AED", 0)),
+        )
```
brianjlai, line 3893: i think we're also missing extracting and checking the key for `USD` so we need one more `x.get("USD", 0)`

brianjlai (commented):
one adjustment then can ✅

unit_tests/sources/declarative/test_concurrent_declarative_source.py
aaronsteers: Thanks, @brianjlai - on it!

aaronsteers: Resolved!

brianjlai (approved):

PR #743, opened 2025-09-03
chore: clean up ConcurrentDeclarativeSource only processing DefaultStream and streamline inheritance to Source interface

brianjlai:
Note: This should not be merged until after https://github.com/airbytehq/airbyte-python-cdk/pull/707 since this PR depends on logic implemented there. And this will fail a lot of tests until after that change is rebased into this one

## What

**Warning this will be breaking and part of the CDK release to v7**

Now that the `ConcurrentDeclarativeSource` has been updated to only receive and process `DefaultStream` from the model to component factory [here](https://github.com/airbytehq/airbyte-python-cdk/pull/707), we can significantly simplify how we handle connector operations and process syncs without account for the legacy synchronous flow.

All the legacy `DeclarativeStream` and `AbstractStream` inheritance has been extracted from `ConcurrentDeclarativeSource`.

## How

Broadly the big changes are:

- `ConcurrentDeclarativeSource` no longer implements `AbstractSource` and instead directly implements the primary` Source` interface
- There is no more grouping logic because all streams are `DefaultStream`.
- Methods like `message_repository()` are removed since only `AbstractSource` required them
- Simplified some of the connection check logic. Ideally we would have streamlined this more, but because the `ConnectionChecker` component takes in the source itself. The `CheckStream` component does a lot of shenanigans withe source object itself so extracting it is a bit annoying
- Moved all legacy (non-concurrent) cursors to legacy
- Moved `DeclarativeStream` to the legacy folder

Something worth noting is that I did not move `AbstractSource` to the legacy package. It has a lot of dependencies in the file CDK, scenario builder, and various Legacy <> Synchronous CDK adapters, so its not quite a full on legacy class yet. Despite it still being in the main package, it has been fully extracted out of the `ConcurrentDeclarativeSource` flow so we're still left with a streamline code flow when processing all low-code syncs.

## Out of scope

Two things that I've left out of scope to make the change more manageable and because they should have minimal impact on the customer facing interface

- `SimpleRetriever` still has state methods and fields for legacy cursors. These should be more easily extractable because they're not invoked in the flow anymore
- The `AbstractStream` still makes references to things like RFR and checkpoint readers. While these are effectively legacy components, they're not referenced by the concurrent CDK so cleanup is neither urgent nor breaking.

## Summary by CodeRabbit

* **New Features**
  * check(logger, config) now returns a status object (Status.SUCCEEDED / Status.FAILED).
  * streams(config) accessor added to retrieve declarative stream lists.

* **Refactor**
  * Declarative sources now use the public Source typing and a unified config-driven flow for concurrent sources.
  * Several declarative cursor types and stream implementations moved to legacy namespaces; internal concurrency utilities reorganized.
  * Removed deprecation marker from AbstractStream.

* **Removal**
  * CustomIncrementalSync removed; incremental sync supports only DatetimeBasedCursor and IncrementingCountCursor.

* **Tests**
  * Unit tests updated to use the new status/check API, stream-access patterns, and adjusted imports.

### DIFF
--- a/unit_tests/connector_builder/test_connector_builder_handler.py
+++ b/unit_tests/connector_builder/test_connector_builder_handler.py
@@ -7,7 +7,7 @@ import dataclasses
 import json
 import logging
 import os
-from typing import List, Literal, Union
+from typing import List, Literal
 from unittest import mock
 from unittest.mock import MagicMock, patch
 
@@ -32,7 +32,6 @@ from airbyte_cdk.connector_builder.models import (
     StreamReadPages,
     StreamReadSlices,
 )
-from airbyte_cdk.legacy.sources.declarative.declarative_stream import DeclarativeStream
 from airbyte_cdk.models import (
     AirbyteLogMessage,
     AirbyteMessage,
@@ -440,12 +439,8 @@ MOCK_RESPONSE = {
 }
 
 
-def get_retriever(stream: Union[DeclarativeStream, DefaultStream]):
-    return (
-        stream.retriever
-        if isinstance(stream, DeclarativeStream)
-        else stream._stream_partition_generator._partition_factory._retriever
-    )
+def get_retriever(stream: DefaultStream):
+    return stream._stream_partition_generator._partition_factory._retriever
 
 
 @pytest.fixture
@@ -983,12 +978,6 @@ def create_mock_retriever(name, url_base, path):
     http_stream.requester.get_path.return_value = path
     http_stream._paginator_path.return_value = None
     return http_stream
-
-
-def create_mock_declarative_stream(http_stream):
-    declarative_stream = mock.Mock(spec=DeclarativeStream, autospec=True)
-    declarative_stream.retriever = http_stream
-    return declarative_stream
 
 
 @pytest.mark.parametrize(
--- a/unit_tests/sources/declarative/partition_routers/test_cartesian_product_partition_router.py
+++ b/unit_tests/sources/declarative/partition_routers/test_cartesian_product_partition_router.py
@@ -22,6 +22,8 @@ from airbyte_cdk.sources.declarative.requesters.request_option import (
 from airbyte_cdk.sources.types import StreamSlice
 
 
+# todo: All these tests rely on stream_slicers that are of a the deprecated legacy class DatetimeBasedCursor these
+#  should really be ConcurrentCursor, but this fix is a bit tedious and are tested in other parts of the code
 @pytest.mark.parametrize(
     "test_name, stream_slicers, expected_slices",
     [
--- a/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
+++ b/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
@@ -569,6 +569,12 @@ class SimpleRetriever(Retriever):
         """
         return self.stream_slicer.stream_slices()
 
+    # todo: There are a number of things that can be cleaned up when we remove self.cursor and all the related
+    #  SimpleRetriever state management that is handled by the concurrent CDK Framework:
+    #  - ModelToComponentFactory.create_datetime_based_cursor() should be removed since it does need to be instantiated
+    #  - ModelToComponentFactory.create_incrementing_count_cursor() should be removed since it's a placeholder
+    #  - test_simple_retriever.py: Remove all imports and usages of legacy cursor components
+    #  - test_model_to_component_factory.py:test_datetime_based_cursor() test can be removed
     @property
     def state(self) -> Mapping[str, Any]:
         return self.cursor.get_stream_state() if self.cursor else {}
--- a/unit_tests/sources/declarative/retrievers/test_simple_retriever.py
+++ b/unit_tests/sources/declarative/retrievers/test_simple_retriever.py
@@ -10,7 +10,6 @@ from unittest.mock import MagicMock, Mock, patch
 import pytest
 import requests
 
-# todo: These can be removed once SimpleRetriever deprecated the now unused state methods
 from airbyte_cdk.legacy.sources.declarative.incremental import (
     DatetimeBasedCursor,
     DeclarativeCursor,


airbyte_cdk/__init__.py
```diff
@@ -65,6 +65,8 @@
 from .connector import BaseConnector, Connector
 from .destinations import Destination
 from .entrypoint import AirbyteEntrypoint, launch
+from .legacy.sources.declarative.declarative_stream import DeclarativeStream
+from .legacy.sources.declarative.incremental import DatetimeBasedCursor
```
brianjlai, line 69: I'd have liked to move these out of this class so we don't expose the classes in the main package structure, but for some reason, removing them creates a cyclical dependency between `RecordSelector` and `AsyncRetriever`. I dug into a bit and got stuck and moved on.

unit_tests/sources/streams/concurrent/test_cursor.py
```diff
@@ -12,9 +13,11 @@
 import pytest
 from isodate import parse_duration
 
+from airbyte_cdk.legacy.sources.declarative.incremental.datetime_based_cursor import (
+    DatetimeBasedCursor,
+)
```
brianjlai, line 18: We defined this test very stupidly. For some reason we instantiate a `DatetimeBasedCursor` only to then extract its value and build a concurrent Cursor. I'll investigate why and update the test or this comment as to the reasoning

unit_tests/sources/declarative/stream_slicers/test_stream_slicer_read_decorator.py
```diff
@@ -4,18 +4,18 @@
...
-from airbyte_cdk.sources.declarative.datetime.min_max_datetime import MinMaxDatetime
-from airbyte_cdk.sources.declarative.incremental import (
+from airbyte_cdk.legacy.sources.declarative.incremental import (
     CursorFactory,
     DatetimeBasedCursor,
     GlobalSubstreamCursor,
     PerPartitionWithGlobalCursor,
 )
-from airbyte_cdk.sources.declarative.incremental.declarative_cursor import DeclarativeCursor
+from airbyte_cdk.legacy.sources.declarative.incremental.declarative_cursor import DeclarativeCursor
```
brianjlai, line 13: brian todo: delete these and instead usage `ConcurrentCursor` and `SubstreamPartitionRouter` to test cases instead of legacy components

airbyte_cdk/sources/declarative/checks/check_stream.py
```diff
@@ -64,7 +64,10 @@ def _log_error(self, logger: logging.Logger, action: str, error: Exception) -> T
         return False, error_message
 
     def check_connection(
-        self, source: AbstractSource, logger: logging.Logger, config: Mapping[str, Any]
+        self,
+        source: Source,
```
maxi297, line 68: I feels like this is the wrong interface as we call `resolved_manifest` below. Should this be `ConcurrentDeclarativeSource` ?

maxi297 (commented):
I added a couple of comments. Should we redo a bit of regression tests once this is merged with the removal of DeclarativeStream stuff?

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2660,7 +2658,9 @@ def create_dynamic_schema_loader(
             config=config,
             name=name,
             primary_key=None,
-            stream_slicer=self._build_stream_slicer_from_partition_router(model.retriever, config),
+            partition_router=self._build_stream_slicer_from_partition_router(
```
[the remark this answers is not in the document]
brianjlai: hmm this is weird, I don't know where these changes even came from because I don't remember adjusting this for my PR, I'm just going to revert these back to latest master w/ your latest changes. so these should get resolved

unit_tests/sources/declarative/partition_routers/test_substream_partition_router.py
```diff
@@ -8,16 +8,15 @@
 
 import pytest as pytest
 
+from airbyte_cdk.legacy.sources.declarative.incremental import (
+    ChildPartitionResumableFullRefreshCursor,
+    ResumableFullRefreshCursor,
+)
```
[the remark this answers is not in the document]
brianjlai: yep after rebasing, these legacy components weren't needed in the test

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -417,47 +370,22 @@
...
+            yield from self._concurrent_source.read(selected_concurrent_streams)
 
     def discover(self, logger: logging.Logger, config: Mapping[str, Any]) -> AirbyteCatalog:
-        concurrent_streams, synchronous_streams = self._group_streams(config=config)
         return AirbyteCatalog(
-            streams=[
-                stream.as_airbyte_stream() for stream in concurrent_streams + synchronous_streams
-            ]
+            streams=[stream.as_airbyte_stream() for stream in self.streams(config=self._config)]
         )
```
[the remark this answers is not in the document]
brianjlai: this is intentional as we want to _eventually_ deprecate the incoming `config` parameter in favor of the initialized `self._config` value

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -417,47 +370,22 @@ def read(
...
+        selected_concurrent_streams = self._select_streams(
+            streams=self.streams(config=self._config),  # type: ignore  # We are migrating away from the DeclarativeStream implementation and streams() only returns the concurrent-compatible AbstractStream. To preserve compatibility, we retain the existing method interface
+            configured_catalog=catalog,
+        )
 
-        yield from super().read(logger, config, filtered_catalog, state)
+        # It would appear that passing in an empty set of streams causes an infinite loop in ConcurrentReadProcessor.
+        # This is also evident in concurrent_source_adapter.py so I'll leave this out of scope to fix for now
+        if len(selected_concurrent_streams) > 0:
+            yield from self._concurrent_source.read(selected_concurrent_streams)
```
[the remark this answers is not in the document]
brianjlai: this is intentional as we want to _eventually_ deprecate the incoming `config` parameter in favor of the initialized `self._config` value

airbyte_cdk/sources/declarative/checks/check_stream.py
brianjlai: The problem here was that the parent `ConnectionChecker` interface is still unfortunately referenced by some of our legacy components like `ManifestDeclarativeSource` and `DeclarativeSource` in the signature. So we can't quite drop in replace `CheckStream's` `AbstractSource` w/ `ConcurrentDeclarativeSource` without changing the parent signature. 

I suspect we've always had this issue though because even prior to our recent refactor, up, we were referencing `AbstractSource` instead of `ManifestDeclarativeSource` which would have been more accurate. And in this method's code we already had the ugly `hasattr` checks:
```
should_check_dynamic_streams = (
  hasattr(source, "resolved_manifest")
  and hasattr(source, "dynamic_streams")
  and self.dynamic_streams_check_configs
)
```

Would you be okay if we left this as is for the moment?

airbyte_cdk/__init__.py
brianjlai: I'd say pretty close mainly the few places like the aforementioned SimpleRetriever when finally yank our the cursor field.

I also went back and cleaned out various usages of `from airbyte_cdk import` from our declarative package, not strictly needed but might as well

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2618,6 +2613,8 @@ def create_gzip_decoder(
             fallback_parser=gzip_parser.inner_parser,
         )
 
+    # todo: This method should be removed once we deprecate the SimpleRetriever.cursor field and the various
```
maxi297, line 2616: nit: I feel this comment would be more actionable if it were in the SimpleRetriever

unit_tests/connector_builder/test_connector_builder_handler.py
```diff
@@ -32,6 +32,7 @@
     StreamReadPages,
     StreamReadSlices,
 )
+from airbyte_cdk.legacy.sources.declarative.declarative_stream import DeclarativeStream
```
maxi297, line 35: I think this can be removed. It was still used during the migration but I think it's not anymore

unit_tests/sources/declarative/partition_routers/test_cartesian_product_partition_router.py
```diff
@@ -5,8 +5,10 @@
 
 import pytest as pytest
 
+from airbyte_cdk.legacy.sources.declarative.incremental.datetime_based_cursor import (
```
maxi297, line 8: It feels like things should just be mocked and we could remove this but this seems like a non-trivial effort so can we just add a comment saying "screw this"?

maxi297 (approved):
LGTM, let's do this!

PR #755, opened 2025-09-09
feat(requests-ca): respect REQUESTS_CA_BUNDLE environment variable

bosper351:
Current implementation of `HttpClient` uses `requests` library to create a `PreparedRequest` and send it, but do not take into consideration the environment variables that `requests` do. This make it impossible to use custom CA certificates together with Airbyte. This PR adds the support of env settings to `HttpClient`.

The implementation is doing what `requests` library recommends when working with `PreparedRequests` to properly handle self-signed certificates: https://requests.readthedocs.io/en/latest/user/advanced/#prepared-requests

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP requests now honor environment-based session settings (e.g., proxies, custom CA bundles, SSL verification, client certificates), improving compatibility with enterprise networks and custom trust stores.

* **Tests**
  * Added a unit test confirming that a custom CA bundle specified via environment variable is correctly applied during requests.


unit_tests/sources/streams/http/test_http.py
```diff
@@ -101,8 +104,16 @@ def test_request_kwargs_used(mocker, requests_mock):
 
     list(stream.read_records(sync_mode=SyncMode.full_refresh))
 
-    stream._http_client._session.send.assert_any_call(ANY, **request_kwargs)
```
bosper351, line 104: `proxies` dict is converted to OrderedDict somewhere under the hood. This plain assertion fails due to type mismatch between a dict and OrderedDict. To avoid this mismatch, each key of `proxies` is compared separately.

bosper351:
@aaronsteers it's resolved. Please check out when you have a moment.

PR #756, opened 2025-09-09
chore: minor improvements to working on manifest only connectors locally

dbgold17:
I found these changes to be helpful when working on updating the cdk to enable new functionality in a manifest-only connector so I figured I'd contribute them back.

1. upgrade the dagger version used by the `bin/generate_component_manifest_files.py` script for generating pydantic models from the declarative_component_schema.yaml file. This fixes an issue I ran into locally when using `0.13.3` where the script would error our with 
```
error: parse selections: parse field "args": __Field.args has no such argument: "includeDeprecated"
⠴ Establishing connection to the API serverError: Sequence aborted after failed subtask 'assemble' 
```
2. declare the dagger dependency as a normal dev dependency in the pyproject.toml file instead of hardcoded in the script.

3. run `poe ruff-fix` automatically after regenerating pydantic models to avoid seeing unrelated spacing / indentation diffs

4. modify the `debug_manifest.py` script to allow the caller to specify a path to a manifest yaml file and custom components python file, matching what exists for configs and catalogs. I found it useful when I had multiple slightly varying copies of a manifest file locally. If not specified, the script will default to looking for the file in the previously hardcoded location.

## Summary by CodeRabbit

* **New Features**
  * Debug tool accepts a manifest path (with a sensible default) and an optional components path for loading extra components.

* **Documentation**
  * Updated debugging instructions and VSCode launch example to show manifest-path and components-path usage; added a placeholder interpreter note.

* **Chores**
  * Bumped dagger-io to 0.18.6.
  * Removed explicit dependency install step from the manifest generator script.
  * Added automatic code-formatting to build tasks.

### DIFF
--- a/debug_manifest/debug_manifest.py
+++ b/debug_manifest/debug_manifest.py
@@ -19,8 +19,8 @@ def debug_manifest(source: YamlDeclarativeSource, args: list[str]) -> None:
 
 def _register_components_from_file(filepath: str) -> None:
     """
-    Dynamically load a Python file containing custom component definitions and register it 
-    under specific module names in sys.modules to ensure that these classes can be properly 
+    Dynamically load a Python file containing custom component definitions and register it
+    under specific module names in sys.modules to ensure that these classes can be properly
     resolved during hydration of the manifest yaml file.
 
     This is a somewhat hacky replacement for the file structure manipulation we do when building
@@ -29,7 +29,7 @@ def _register_components_from_file(filepath: str) -> None:
     import importlib.util
     import sys
     from pathlib import Path
-    
+
     components_path = Path(filepath)
     if not components_path.exists():
         raise FileNotFoundError(f"Components file not found: {components_path}")


debug_manifest/debug_manifest.py
```diff
@@ -24,13 +24,15 @@ def debug_manifest(source: YamlDeclarativeSource, args: list[str]) -> None:
 
 if __name__ == "__main__":
     args = sys.argv[1:]
+    parsed_args = AirbyteEntrypoint.parse_args(args)
+    manifest_path = AirbyteEntrypoint.parse_args(args).manifest_path or "resources/manifest.yaml"
```
dbgold17, line 28:
```suggestion
    manifest_path = parsed_args.manifest_path or "resources/manifest.yaml"
```

debug_manifest/README.md
```diff
@@ -22,11 +22,15 @@ To configure the debugger in VSCode to run the `debug_manifest`, follow these st
       "request": "launch",
       "console": "integratedTerminal",
       "cwd": "${workspaceFolder}/debug_manifest",
-      "python": "<PATH_TO_CDK_ENV>/bin/python",
+      "python": "<PATH_TO_CDK_ENV>/bin/python", // REPLACE ME
       "module": "debug_manifest",
       "args": [
         // SPECIFY THE COMMAND: [spec, check, discover, read]
         "read",
+        // SPECIFY THE MANIFEST FILE
```
maxi297, line 30: What about the components.py file though? It seems odd to have the discrepancy where the manifest can be specified but not the components file. It seems like we could do something like `PYTHONPATH=/path/to/your/components:$PYTHONPATH python debug_manifest.py read <...>>` but I haven't tried

maxi297 (approved):
That makes sense to me

debug_manifest/README.md
dbgold17: ok I added something

maxi297: When I test locally, I just copy the file under `debug_manifest/resources` because I added it to my PYTHONPATH. I'm not sure what the folder `source_declarative_manifest` has to do with it. Can you add more context here so that I can understand?

dbgold17: so I'm still figuring this out myself, but it seems like when the model to components factory loads in custom components, it looks for a python module specified by `class_name` in the manifest file. And it seems like we have the convention of namespacing the custom component under `source_declarative_manifest`.  Which looks something like this:
```
  requester:
    type: CustomRequester
    class_name: "source_declarative_manifest.components.MondayGraphqlRequester"
    http_method: POST
    schema_loader:
      type: InlineSchemaLoader
```


I believe when we build connector images, we move the custom components file under that path ([code here](https://github.com/airbytehq/airbyte/blob/c85cacacfad06dd07940bfce86319c51140a0ceb/docker-images/Dockerfile.manifest-only-connector#L26)).

For this test, I didn't want to enforce a naming convention for the file name or file path, and I found this snippet of code being used in the SDM CLI already.

maxi297 (approved):

debug_manifest/README.md
dbgold17: woops, just updated that link to point to master. And yes I saw that, I think that may still cause some issues if the file isn't named `components.py` but regardless, makes sense and I'll add a bit more clarity in comments in this helper method

maxi297: I think it is a requirement for manifest only connectors that the code is in a file called `components.py`

PR #762, opened 2025-09-15
feat(http_client): respective failure_type is raised after exhausting request attempts and failing w/ backoff exception

pnilan:
## What
After exhausting request attempts, the backoff exception handled by the retry decorator in `HttpClient._send_with_retry` would bubble up w/o a defined failure type. This would cause backoff exceptions that exhaust the all retries to be categorized as `FailureType.system_error`. This PR adds exception handling in the `_send_with_retry` method to catch backoff exceptions *after* all request attempts are exhausted and re-raise as an `AirbyteTracedException` w/ the failure_type defined in the `ErrorResolution` that is returned from the error handler's interpret_response method. If failure_type is None then defaults to `system_error`.

- Updates the `BaseBackoffException`, `DefaultBackoffException`, `UserDefinedBackoffException`, and `RateLimitedBackoffException` to take an optional `failure_type` property.
- Additionally, I updated the default status error mapping for clearer error messaging.

## Impact
- Fewer non-actionable pages/sentry alerts to the Maintain Team.
- Clearer error messaging in Cloud UI

## Recommended Reading Order
- `http_client.py`
- `exceptions.py`
- Tests

## Summary by CodeRabbit

* **Improvements**
  * HTTP error messages now include status codes and an "Error:" prefix for clearer diagnostics.
  * Connection checks report unavailable streams including the stream name for easier troubleshooting.
  * Exhausted retries now raise a traced error, log contextual details, and preserve failure classification for accurate handling.
  * Token refresh backoff now classifies transient errors explicitly.

* **Bug Fixes**
  * Standardized user-facing error text across HTTP handlers for consistency.

* **Tests**
  * Unit tests updated to expect revised messages and traced-exception behavior.

### DIFF
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -240,7 +240,11 @@ class AbstractOauth2Authenticator(AuthBase):
         except requests.exceptions.RequestException as e:
             if e.response is not None:
                 if e.response.status_code == 429 or e.response.status_code >= 500:
-                    raise DefaultBackoffException(request=e.response.request, response=e.response)
+                    raise DefaultBackoffException(
+                        request=e.response.request,
+                        response=e.response,
+                        failure_type=FailureType.transient_error,
+                    )
             if self._wrap_refresh_token_exception(e):
                 message = "Refresh token is invalid or expired. Please re-authenticate from Sources/<your source>/Settings."
                 raise AirbyteTracedException(


airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -290,15 +292,25 @@ def _send_with_retry(
...
+            return response
+        except BaseBackoffException as e:
+            self._logger.error(f"Retries exhausted with backoff exception.", exc_info=True)
+            raise MessageRepresentationAirbyteTracedErrors(
+                internal_message=f"Exhausted available request attempts. Exception: {e}",
+                message=f"Exhausted available request attempts. Please see logs for more details. Exception: {e}",
+                failure_type=FailureType.transient_error,
+                exception=e,
+                stream_descriptor=StreamDescriptor(name=self._name),
+            )
```
pnilan, line 313: This is the main change for this PR to raise failed retries as transient.

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -290,15 +292,25 @@ def _send_with_retry(
...
+        try:
+            response = backoff_handler(rate_limit_backoff_handler(user_backoff_handler))(
+                request,
+                request_kwargs,
+                log_formatter=log_formatter,
+                exit_on_rate_limit=exit_on_rate_limit,
+            )  # type: ignore # mypy can't infer that backoff_handler wraps _send
+
+            return response
+        except BaseBackoffException as e:
```
maxi297, line 305: Should it be `RateLimitBackoffException` instead of `BaseBackoffException`? I see `DefaultBackoffException` and `UserDefinedBackoffException` inheriting from `BaseBackoffException` and those don't seem related to rate limiting

unit_tests/sources/streams/http/test_http.py
```diff
@@ -580,14 +603,16 @@ def test_http_stream_adapter_http_status_error_handler_should_retry_false_raise_
...
+    res.headers = {}
     mocker.patch.object(requests.Session, "send", return_value=res)
     mocker.patch.object(stream._http_client, "_logger")
-    with pytest.raises(requests.exceptions.HTTPError):
-        response = stream._http_client.send_request("GET", "https://g", {}, exit_on_rate_limit=True)
+    with pytest.raises(AirbyteTracedException):
+        _, response = stream._http_client.send_request(
+            "GET", "https://g", {}, exit_on_rate_limit=True
+        )
         stream._http_client.logger.error.assert_called_with(response.text)
```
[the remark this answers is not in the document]
pnilan: Yes

airbyte_cdk/sources/streams/http/exceptions.py
```diff
@@ -7,14 +7,18 @@
...
+from airbyte_cdk.models import FailureType
+
 
 class BaseBackoffException(requests.exceptions.HTTPError):
     def __init__(
         self,
         request: requests.PreparedRequest,
         response: Optional[Union[requests.Response, Exception]],
         error_message: str = "",
+        failure_type: Optional[FailureType] = None,
```
maxi297, line 19: I see usage of `DefaultBackoffException` in `abstract_oauth.py`. Should we update this one as well?

unit_tests/sources/streams/http/test_http_client.py
```diff
@@ -744,3 +744,56 @@ def test_given_different_headers_then_response_is_not_cached(requests_mock):
...
+        (403, FailureType.transient_error, "test error message", UserDefinedBackoffException),
+        (400, FailureType.system_error, "test error message", DefaultBackoffException),
+        (401, FailureType.config_error, "test error message", DefaultBackoffException),
+        (403, FailureType.transient_error, "test error message", DefaultBackoffException),
+        (400, FailureType.system_error, "test error message", RateLimitBackoffException),
+        (401, FailureType.config_error, "test error message", RateLimitBackoffException),
+        (403, FailureType.transient_error, "test error message", RateLimitBackoffException),
+    ],
+)
+def test_send_with_retry_raises_airbyte_traced_exception_with_failure_type(
```
maxi297, line 764: I like the fact that you added test coverage but I fear this test might be fragile i.e. that a change on how HttpClient is implemented might break this test even if this behavior does not change. This is caused by this test calling/mocking internal methods like `_send` and `_send_with_retry`.  Would there be a way to refactor this to ensure it is not as fragile? Can we use our HttpMocker to simulate the use cases?

maxi297 (approved):
Two comments but nothing blocking. Thanks @pnilan !

airbyte_cdk/sources/streams/http/http_client.py
pnilan: Follow-up: updated this to add a failure_type to `BaseBackoffException`, now when the `AirbyteTracedException` is raised after exhausting all attempts, we'll raise w/ the specified failure type.

PR #774, opened 2025-09-26
fix: properly handle multi-line secrets in secret masking

aaronsteers:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI secret masking to recursively find secrets in nested configurations (dicts/lists) and mask multi-line values line-by-line, preventing accidental exposure while preserving non-secret output and CI gating behavior.

* **Tests**
  * Added comprehensive unit tests validating masking across flat, nested, list-based, multi-line and edge-case configurations.

### DIFF
--- a/airbyte_cdk/cli/airbyte_cdk/_secrets.py
+++ b/airbyte_cdk/cli/airbyte_cdk/_secrets.py
@@ -466,7 +466,8 @@ def _print_ci_secret_mask_for_string(secret: str) -> None:
     For multi-line secrets, we must print a secret mask for each line separately.
     """
     for line in secret.splitlines():
-        print(f"::add-mask::{line!s}")
+        if line.strip():  # Skip empty lines
+            print(f"::add-mask::{line!s}")
 
 
 def _print_ci_secret_mask_for_value(value: Any) -> None:
@@ -491,7 +492,8 @@ def _print_ci_secret_mask_for_value(value: Any) -> None:
     # For any other types besides dict and list, we convert to string and mask each line
     # separately to handle multi-line secrets (e.g. private keys).
     for line in str(value).splitlines():
-        _print_ci_secret_mask_for_string(line)
+        if line.strip():  # Skip empty lines
+            _print_ci_secret_mask_for_string(line)
 
 
 def _print_ci_secrets_masks_for_config(
--- a/airbyte_cdk/cli/airbyte_cdk/_secrets.py
+++ b/airbyte_cdk/cli/airbyte_cdk/_secrets.py
@@ -459,31 +459,39 @@ def _print_ci_secrets_masks(
         _print_ci_secrets_masks_for_config(config=config_dict)
 
 
+def _print_ci_secret_mask_for_string(secret: str) -> None:
+    """Print GitHub CI mask for a single secret string.
+
+    We expect single-line secrets, but we also handle the case where the secret contains newlines.
+    For multi-line secrets, we must print a secret mask for each line separately.
+    """
+    for line in secret.splitlines():
+        print(f"::add-mask::{line!s}")
+
+
 def _print_ci_secret_mask_for_value(value: Any) -> None:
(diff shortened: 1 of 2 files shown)


airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -459,25 +459,60 @@ def _print_ci_secrets_masks(
...
+        if line.strip():  # Skip empty lines
+            print(f"::add-mask::{line!s}")
+
+
+def _print_ci_secret_mask_for_value(value: Any) -> None:
+    """Print GitHub CI mask for a single secret value.
+
+    Call this function for any values identified as secrets, regardless of type.
+    """
+    if isinstance(value, dict):
```
dbgold17, line 478: If I understand correctly, this method `_print_ci_secret_mask_for_value` only get's called once we deem we've found a secret. So even if the value happens to be a dict, can we just treat it as one big opaque secret blob for the purposes of masking? i.e. do we need to recurse here?

FWIW, this is the first time I'm looking at how we do secrets masking, so I don't have a great understanding yet.

airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -459,25 +459,60 @@ def _print_ci_secrets_masks(
...
+        # For nested dicts, we call recursively on each value
+        for v in value.values():
```
dbgold17, line 480: as written, wouldn't this not mask the entire dictionary.

My reading is that given a dictionary deemed to be a secret like this:
```
{
"k1": {
  "k2": "v1",
  "k3": "v2"
}
}
```
we'd only mask the terminal values like this:

```
{
"k1": {
  "k2": "***",
  "k3": "***"
}
}
```

airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -459,25 +459,60 @@ def _print_ci_secrets_masks(
         _print_ci_secrets_masks_for_config(config=config_dict)
 
 
+def _print_ci_secret_mask_for_string(secret: str) -> None:
+    """Print GitHub CI mask for a single secret string.
+
+    We expect single-line secrets, but we also handle the case where the secret contains newlines.
+    For multi-line secrets, we must print a secret mask for each line separately.
+    """
+    for line in secret.splitlines():
```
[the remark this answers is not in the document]
dbgold17: oh I just saw your response to Jose asking the same question

airbyte_cdk/cli/airbyte_cdk/_secrets.py
aaronsteers: @dbgold17 - I did consider this but decided to only mask the values if the secrets are nested.

I tried but could not think of any cases where the key of the dictionary would be a secret on its own. (Again, this is scoped to the manner in which we have users specifying config in our connectors.)

The problem with masking they keys of the dictionary: If the parent is "credentials" and the sub-key is "password" and the value of credentials.password is the password, we do want to mask the password, but masking the word "password" would make it censored everywhere in our logs. So an error like "Missing password" would become "Missing ****", which is not what we want.

airbyte_cdk/cli/airbyte_cdk/_secrets.py
```diff
@@ -459,25 +459,60 @@ def _print_ci_secrets_masks(
...
+        if line.strip():  # Skip empty lines
+            print(f"::add-mask::{line!s}")
+
+
+def _print_ci_secret_mask_for_value(value: Any) -> None:
```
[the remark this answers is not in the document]
aaronsteers: @jpefaur - Good question. The only reason I created it is because we have to call it twice in the calling function - once for lists and once for dictionaries. And I wanted to be extra careful to keep things DRY for safety and clarity.

dbgold17 (approved):
discussed online, approving!

PR #776, opened 2025-09-30
feat(jwt authenticator): Allow configuring the signed JWT token to be injected as a request_option for JwtAuthenticator

brianjlai:
## What

When first developing the JWT authenticator, based on various specs and usage by connectors, we did not see a use case where the signed JWT token needed to be injected anywhere but under the `Authorization` header of requests.

While working on a Box connector, this is the first usage where we need to load it into the `request_json_body` under the `assertion` key.

## How

This PR adds functionality so that the `JwtAuthenticator` can now specify an optional `request_option` to inject the signed token into different request options. Optional so that we retain backwards compatibility

## Validation steps

- I wrote a test manifest using this flow and loading into the json body
- Validate against existing connectors like `source-google-search-console` which use the default `Authorization` header behavior to ensure regressions

## Summary by CodeRabbit

- **New Features**
  - JWT authenticator supports configurable token injection targets (header, query parameter, form data, JSON body) and customizable header key (default: Authorization).

- **Schema**
  - Declarative schema now allows specifying request_option on JWT authenticators.

- **Parser**
  - Manifests with request_option for JWT auth are propagated into created components.

- **Tests**
  - Added coverage for JWT injection into headers, params, form data, JSON body, and custom header keys.

- **Chores**
  - Dependency bumped: dagger-io updated.


airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -350,6 +350,35 @@ class Algorithm(Enum):
     EdDSA = "EdDSA"
 
 
+class InjectInto(Enum):
```
brianjlai, line 353: I've been having trouble getting the `poetry run poe build` command to run locally for me right now. I made similar changes to what was noted by David https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1757430835462669 

However, I'm still seeing the same error despite upgrading to dagger 18. I did some shenanigans by writing this by hand so I don't block testing or review. But i'll look into this more tomorrow

PR #790, opened 2025-10-14
poc: ErrorHandler.interpret_response can return None

maxi297:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected a spelling inconsistency in variable names related to response filters.

* **Refactor**
  * Modified error handling so that default HttpResponseFilters are not automatically added; only explicitly provided filters are used.
  * Updated various error handler methods to allow for absent (None) error resolutions instead of constructing default messages.
  * Streamlined how HTTP error resolutions are handled, ensuring a clear distinction between success and error outcomes.

* **Tests**
  * Simplified test cases to reflect updated error resolution handling.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2277,9 +2277,6 @@ class ModelToComponentFactory:
                 response_filters.append(
                     self._create_component_from_model(model=response_filter_model, config=config)
                 )
-        response_filters.append(
-            HttpResponseFilter(config=config, parameters=model.parameters or {})
-        )
 
         return DefaultErrorHandler(
             backoff_strategies=backoff_strategies,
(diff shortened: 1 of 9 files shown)


hai-ld:
This patch seems to fix a bug where an error response doesn't trigger the matching handler if
- there are two handlers with either FAIL or RATE_LIMIT actions
- the first handler match, but the second one doesn't match
- the error response has 200 OK status (https://github.com/airbytehq/airbyte-python-cdk/pull/790/changes#diff-45d069b580f03bb689c2e0d1bcfe2b29a9b268c405fad26762ab44821035fc12R122)

This is due to the weird check on https://github.com/airbytehq/airbyte-python-cdk/pull/790/changes#diff-d7b968fcabe9aa5a2f910d18228e8241fedfa45c84493088c3092248057a6e7aL69 that ignores the resolution of FAIL and RATE_LIMIT handlers.

PR #792, opened 2025-10-15
fix(manifest server): default concurrency settings

maxi297:
Users are entering a deadlock state where the partition generator threads are filling the queue of partitions to process without having a thread to relieve the queue. 

See: https://airbytehq-team.slack.com/archives/C09B97YAB2B/p1760532962689779

## Summary by CodeRabbit

* **Bug Fixes**
  * The system no longer injects a default concurrency value when a manifest omits concurrency settings — manifests without an explicit concurrency level remain unchanged.
  * Manifests that already specify a concurrency level continue to behave as configured.

* **Tests**
  * Tests updated to validate behavior when concurrency is present or absent; an explicit-case test for enforced concurrency was added and one overly-specific constructor assertion test was removed.


pedroslopez (approved):
:shipit:

PR #808, opened 2025-10-22
feat: cache properties from endpoint

maxi297:
## What 

https://github.com/airbytehq/airbyte-internal-issues/issues/14929

## How

Adding a cache within PropertiesFromEndpoint. Note that with this solution, every instance of PropertiesFromEndpoint will have a different cache so it may be that the same stream as a child/main stream vs as a parent stream have different properties from endpoint if a field is added between the read of those streams. I don't see a case where this would be problematic though.

## Summary by CodeRabbit

* **Performance Improvements**
  * Endpoint properties are cached to avoid redundant retrievals on repeated calls.

* **Bug Fixes**
  * Property values are consistently returned as strings.
  * Chunking no longer mutates input property lists.

* **Improvements**
  * Property chunking and property retrieval APIs simplified and use consistent list types.

* **Tests**
  * New tests cover single-call caching, type coercion, and input immutability.

* **Documentation**
  * Clarified that stream slices can't be interpolated from this retriever.

### DIFF
--- a/airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
+++ b/airbyte_cdk/sources/declarative/requesters/query_properties/query_properties.py
@@ -34,10 +34,6 @@ class QueryProperties:
         """
         Uses the defined property_list to fetch the total set of properties dynamically or from a static list
         and based on the resulting properties, performs property chunking if applicable.
-        :param stream_slice: The StreamSlice of the current partition being processed during the sync. This is included
-        because subcomponents of QueryProperties can make use of interpolation of the top-level StreamSlice object
-        :param configured_stream: The customer configured stream being synced which is needed to identify which
-        record fields to query for and emit.
         """
         fields: List[str]
         configured_properties = self.property_selector.select() if self.property_selector else None


maxi297:
/autofix
> **Auto-Fix Job Info**
>
> This job attempts to auto-fix any linting or formating issues. If any fixes are made,
> those changes will be automatically committed and pushed back to the PR.
>
> Note: This job can only be run by maintainers. On PRs from forks, this command requires
> that the PR author has enabled the `Allow edits from maintainers` option.

> PR auto-fix job started... [Check job output.][1]

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/18718343885
> ✅ Changes applied successfully.

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
```diff
@@ -22,19 +22,27 @@ class PropertiesFromEndpoint:
...
-            records_schema={}, stream_slice=stream_slice
-        )
-        for property_obj in response_properties:
-            path = [
-                node.eval(self.config) if not isinstance(node, str) else node
-                for node in self._property_field_path
-            ]
-            yield dpath.get(property_obj, path, default=[])  # type: ignore # extracted will be a MutableMapping, given input data structure
+    def get_properties_from_endpoint(self, stream_slice: Optional[StreamSlice]) -> List[str]:
+        if self._cached_properties is None:
```
maxi297: Concern to be resolved as if the cache should be per StreamSlice

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
```diff
@@ -22,19 +22,27 @@ class PropertiesFromEndpoint:
...
-        response_properties = self.retriever.read_records(
-            records_schema={}, stream_slice=stream_slice
-        )
-        for property_obj in response_properties:
-            path = [
-                node.eval(self.config) if not isinstance(node, str) else node
-                for node in self._property_field_path
-            ]
-            yield dpath.get(property_obj, path, default=[])  # type: ignore # extracted will be a MutableMapping, given input data structure
+    def get_properties_from_endpoint(self, stream_slice: Optional[StreamSlice]) -> List[str]:
```
brianjlai: one risk that this does pose because we cache is if we have a crazy large number of properties and we run out of memory. That was why I had originally made this use an iterable so that we yield properties as they come in.

Granted it feels pretty necessary in order to allow caching, but it does also help make the argument for only maintaining a global set of property mapping. Otherwise if we had a large set of properties and we had to store it for every slice that might be more concerning

maxi297: The other risk I've thought of is if someone removes a property during the sync. Then, it might fail because we ask for a property that HubSpot doesn't know about. That being said, I assume this happens rarely and if it's too much of a pain, we can have a refresh on properties maybe later...

maxi297:
/autofix
> **Auto-Fix Job Info**
>
> This job attempts to auto-fix any linting or formating issues. If any fixes are made,
> those changes will be automatically committed and pushed back to the PR.
>
> Note: This job can only be run by maintainers. On PRs from forks, this command requires
> that the PR author has enabled the `Allow edits from maintainers` option.

> PR auto-fix job started... [Check job output.][1]

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/18759676487
> ✅ Changes applied successfully.

airbyte_cdk/sources/declarative/requesters/query_properties/properties_from_endpoint.py
brianjlai: yeah just adding in the description is fine. Since all components references the http requester, its kind of hard to represent different interpolation contexts. best we can do is the description.

And I don't see a lot of scenarios where we would need the stream slice to be used in the properties retrieval since we typically look at a stream being 1:1 w/ a schema/properties. And not a slice being 1:1 w/ a schema/properties. So it doesn't seem like we would ever need this as input

brianjlai (approved):

PR #829, opened 2025-11-07
feat(oauth): Support refresh_token_error on the OAuthAuthenticatorModel level and …

maxi297:
…add sensible defaults

## What

We've seen auth error being flagged system_error ([example with source-google-analytics-data-api](https://airbytehq.sentry.io/issues/6832022917)). It seems like we only support validating the oauth response if a refresh_token_updater is defined.

## How
I've moved the `refresh_token_error` on the root of the `OAuthAuthenticatorModel`. I've also added sensible defaults so that we don't have to configure this for all the connectors.

## Summary by CodeRabbit

* **New Features**
  * OAuth2 authenticator now exposes three configurable refresh-token error settings: status codes (e.g. [400, 500]), error key (e.g. "error"), and error values (e.g. ["invalid_grant","invalid_permissions"]).

* **Deprecations**
  * Refresh-token error fields on nested refresh-updater components are deprecated — please use the new top-level OAuth2 authenticator fields instead.

### DIFF
--- a/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
+++ b/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
@@ -926,28 +926,24 @@ class OAuthConfigSpecification(BaseModel):
     class Config:
         extra = Extra.allow
 
-    oauth_user_input_from_connector_config_specification: Optional[Dict[str, Any]] = (
-        Field(
-            None,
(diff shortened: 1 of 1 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2868,8 +2875,59 @@ def create_oauth_authenticator(
...
+        )
+        if is_defined_on_refresh_token_updated and is_defined_on_oauth_authenticator:
+            raise ValueError(
+                "refresh_token_error should either be defined on the OAuthAuthenticatorModel or the RefreshTokenUpdaterModel, not both"
+            )
+
+        if is_defined_on_refresh_token_updated:
+            not_optional_refresh_token_updater: RefreshTokenUpdaterModel = refresh_token_updater  # type: ignore  # we know from the condition that this is not None
+            return (
+                tuple(not_optional_refresh_token_updater.refresh_token_error_status_codes)
```
maxi297: Note that this will have the same behavior as the previous code but is a bit different. The previous code would pass `refresh_token_updater.refresh_token_error_status_codes` directly which felt very dangerous because the typing was different (`Tuple[int, ...]` vs `Optional[List[int]]`). I fear that the previous code could lead to None exception [here](https://github.com/airbytehq/airbyte-python-cdk/blob/650414816c04c644bff4ace1e4ac1408556b155e/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L182) if `refresh_token_error_status_codes` was not defined. I don't have signals to believe this actually happen in prod though...

Issue #833, opened 2025-11-08
[Epic] Declarative CDK: Add Fetch/Create/Update via SimpleBidirectionalRetriever


aaronsteers:
## Summary / Motivation

We're proposing extending the Python CDK declarative framework beyond list-only reads to support **fetch** (GET single record), **create** (POST), and **update** (PATCH/POST) operations. The goal is to enable bidirectional sync capabilities while maintaining the declarative, low-code nature of the CDK.

The new `SimpleBidirectionalRetriever` component will provide a DRY configuration that shares authentication, base URL, error handling, headers, and response parsing across all four operations (list, fetch, create, update). This allows connector builders to define write operations declaratively without custom Python code.

We'll ship incrementally across three phases, starting with read-only fetch operations and progressing to write operations. Search/filter capabilities will be integrated as a subset of the list operation, reusing the same DPath extractor and record parsing logic.

---

### In Scope
- New retriever type: `SimpleBidirectionalRetriever`
- CLI subcommands: `source-declarative-manifest fetch` (Phase 1), write operations via stdin (Phase 2/3)
- Catalog `record_filters` with list-of-lists-of-maps semantics (outer list = OR slices, inner list = AND conditions)
- Default conventions for conventional REST APIs with override capability
- Per-operation configuration (path, method, body, extractor)
- Optional response parsing for write operations (default: parse and verify)
- Search/filter as list operation capabilities (reuses extractor)

### Out of Scope (for now)
- DELETE operations
- Stripe `/customers/search` endpoint (we'll reserve config space but defer implementation)
- Airbyte protocol changes (Phase 2/3 use CLI-specific stdin JSONL envelope)
- Bulk operations (one record per request)

---

### Phase 1a: Fetch (No Credentials)
**Target APIs:** JSONPlaceholder, Pokémon TCG  
**Goal:** Prove fetch operations work with auth-less APIs

**CLI:**
```bash
source-declarative-manifest fetch --config config.json --catalog configured_catalog.json
```

**Catalog `record_filters` format:**
```json
{
  "streams": [{
    "stream": {"name": "posts", "primary_key": ["id"]},
    "record_filters": {
      "enabled": true,
      "identifiers": [
        [{"id": "1"}],     // Fetch post with id=1
        [{"id": "2"}],     // Fetch post with id=2
        [{"id": "100"}]    // Fetch post with id=100
      ]
    }
  }]
}
```

**Default conventions:**
[...]

PR #846, opened 2025-11-12
feat: Add `Source.fetch_record()` and`Stream.fetch_record()` backed by `SimpleRetriever._fetch_one()`

aaronsteers:
## Summary
Adds the ability to fetch individual records by primary key from declarative concurrent connectors. This implements Phase 1a of [#833](https://github.com/airbytehq/airbyte-python-cdk/issues/833) for PyAirbyte's use case with `ConcurrentDeclarativeSource`.

**What's being added to main:**
- `ConcurrentDeclarativeSource.fetch_record(stream_name, pk_value, config)` - Primary API for fetching single records
- `SimpleRetriever.fetch_one(pk_value, records_schema)` - Underlying HTTP GET mechanism
- `AbstractStream.fetch_record(pk_value)` - Interface method (raises NotImplementedError by default)
- `RecordNotFoundException` - Exception raised when record is not found
- 6 unit tests covering happy path, 404s, server errors, invalid inputs, and empty responses

**Key implementation details:**
- Accepts string primary keys only (e.g., `"123"`)
- Uses convention-based path construction: appends `/{pk_value}` to base path
- Raises `RecordNotFoundException` for empty/ignored responses
- HTTP errors (including 404) propagate from requester's error handling
- Only supports streams with `SimpleRetriever` (raises NotImplementedError for others)

**⚠️ CRITICAL - This PR has NOT been tested end-to-end with a real connector:**

- [ ] **Test with a real declarative connector** - The `ConcurrentDeclarativeSource.fetch_record()` method reconstructs components from the manifest but has never been tested with an actual connector. Test with a simple declarative connector (e.g., JSONPlaceholder, PokeAPI) to verify it works.

- [ ] **Verify component reconstruction** - The logic that creates `declarative_stream` from `stream_config` (lines 552-557 in concurrent_declarative_source.py) has never been exercised. Verify it correctly handles:
  - Standard streams with SimpleRetriever
  - Dynamic streams
  - Streams with custom schema loaders
  - Streams without SimpleRetriever (should raise NotImplementedError)

- [ ] **Test error handling** - Verify that:
  - 404 responses propagate as exceptions (not caught as RecordNotFoundException)
  - Empty responses raise RecordNotFoundException
  - Server errors (5xx) propagate correctly
  - Invalid stream names raise ValueError with helpful message

- [ ] **Verify path construction** - Test that the simple `base_path + "/" + pk_value` pattern works for your target use cases. This may not work for:
  - APIs that use query parameters for IDs (e.g., `?id=123`)
  - Nested resources (e.g., `/companies/{id}/properties/{prop}`)
  - APIs with non-standard path patterns

### Recommended Test Plan
```python
from airbyte_cdk.sources.declarative.concurrent_declarative_source import ConcurrentDeclarativeSource

# Load a simple declarative manifest
source = ConcurrentDeclarativeSource(
    config={"api_key": "..."},
    source_config=manifest_dict,
)

# Test happy path
record = source.fetch_record("posts", "1", config)
assert record["id"] == "1"

# Test 404 (should propagate as exception, not RecordNotFoundException)
try:
    source.fetch_record("posts", "999999", config)
except Exception as e:
    print(f"404 error: {e}")

# Test invalid stream
try:
    source.fetch_record("nonexistent", "1", config)
except ValueError as e:
    assert "not found in source" in str(e)
```

### Notes
- All 6 unit tests pass (covering SimpleRetriever.fetch_one scenarios)
- Ruff linter and formatter pass
- No integration tests exist for ConcurrentDeclarativeSource.fetch_record
- Only supports SimpleRetriever - other retriever types will raise NotImplementedError

---
**Requested by:** AJ Steers (aj@airbyte.io, @aaronsteers)  
**Devin Session:** https://app.devin.ai/sessions/e0d627ac7e89447a9ebddd515104e2c3

## Summary by CodeRabbit

* **New Features**
  * Streams and declarative sources can fetch individual records by primary key; declarative implementations return single-object responses or surface a clear "record not found" error when absent. Other stream types will explicitly indicate if per-record fetch is unsupported.
* **Tests**
  * Added unit tests for successful fetches, not-found cases, server errors, invalid inputs, and empty/no-response scenarios.

### DIFF
--- a/unit_tests/sources/declarative/retrievers/test_simple_retriever.py
+++ b/unit_tests/sources/declarative/retrievers/test_simple_retriever.py
@@ -1706,7 +1706,7 @@ def test_fetch_one_not_found():
 
     assert "999" in str(exc_info.value)
 
-def testfetch_one_server_error():
+def test_fetch_one_server_error():
     """Test fetch_one propagates non-404 errors."""
     requester = MagicMock()
     requester.get_path.return_value = "posts"
@@ -1731,7 +1731,6 @@ def testfetch_one_server_error():
         retriever.fetch_one("123", records_schema={})
 
     assert "Server error" in str(exc_info.value)
-
 
 def testfetch_one_invalid_pk_type():
     """Test fetch_one with non-string pk_value (should fail type checking but test runtime behavior)."""


airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,134 @@ def _to_partition_key(to_serialize: Any) -> str:
...
+            This implementation uses convention-based path construction (Option B from design). (important-comment)
+            For simple PKs: appends /{pk_value} to base path (important-comment)
+            For composite PKs: appends /{value1}/{value2}/... in key order (important-comment)
+
+            Alternative approaches that could be implemented in the future: (important-comment)
+            - Path template interpolation: Use a configurable template like "{base_path}/{id}" (important-comment)
+              See: https://github.com/airbytehq/airbyte-python-cdk/issues/833#phase-1a (important-comment)
+            - Field path configuration: Allow specifying which response field contains the record (important-comment)
+              for APIs that wrap single records in envelopes like {"data": {...}} (important-comment)
+        """
```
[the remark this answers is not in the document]
aaronsteers: Devin, please address this feedback from coderabbit. docstring is stale

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -507,6 +507,83 @@ def check(self, logger: logging.Logger, config: Mapping[str, Any]) -> AirbyteCon
...
+        Raises:
+            ValueError: If the stream name is not found in the source
+            NotImplementedError: If the stream doesn't support fetching individual records
+            RecordNotFoundException: If the record is not found (404 response)
+        """
+        from airbyte_cdk.sources.declarative.models.declarative_component_schema import (
+            DeclarativeStream as DeclarativeStreamModel,
+        )
+        from airbyte_cdk.sources.declarative.retrievers import SimpleRetriever
+        from airbyte_cdk.sources.declarative.schema import DefaultSchemaLoader
```
aaronsteers: NEVER DO THIS. No inline imports dumb dumb.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,123 @@ def _to_partition_key(to_serialize: Any) -> str:
...
+                stream_slice=stream_slice,
+                next_page_token=None,
+            ),
+            log_formatter=self.log_formatter,
+        )
+
+        if not response:
+            raise RecordNotFoundException(
+                f"Record with primary key {pk_value} not found (no response)"
+            )
```
aaronsteers: To confirm: are you sure response is falsey here? (If yes, that's great. I just want to check. And it might be worth a short comment to say so.)

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,123 @@ def _to_partition_key(to_serialize: Any) -> str:
...
+        elif isinstance(pk_value, Mapping):
+            sorted_values = [str(pk_value[key]).lstrip("/") for key in sorted(pk_value.keys())]
+            pk_path_segment = "/".join(sorted_values)
+            fetch_path = f"{base_path.rstrip('/')}/{pk_path_segment}"
+        else:
+            raise ValueError(f"pk_value must be a string or dict, got {type(pk_value).__name__}")
+
+        stream_slice = StreamSlice(partition={}, cursor_slice={})
+
+        response = self.requester.send_request(
```
aaronsteers: Devin, add a type annotation here so the reader knows what we're working with in the "response" variable.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,123 @@ def _to_partition_key(to_serialize: Any) -> str:
...
+
+        records = list(
+            self._parse_response(
+                response=response,
+                stream_state={},
+                records_schema=records_schema,
+                stream_slice=stream_slice,
+                next_page_token=None,
+            )
+        )
```
aaronsteers: This should have a type hint based on return type of _parse_response(). Also, I'm not sure if we strictly want to convert to a list. Can you confirm?

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,102 @@ def _to_partition_key(to_serialize: Any) -> str:
         # separators have changed in Python 3.4. To avoid being impacted by further change, we explicitly specify our own value
         return json.dumps(to_serialize, indent=None, separators=(",", ":"), sort_keys=True)
 
+    def _fetch_one(
```
brianjlai: Given that this is presumably going to be called from somewhere else outside of `SimpleRetriever`, we should make this public. While I understand this is not meant to be public facing or used by the protocol interface, it still feels like bad practice for this be a private method.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -626,6 +627,102 @@ def _to_partition_key(to_serialize: Any) -> str:
...
+            ),
+            log_formatter=self.log_formatter,
+        )
+
+        if not response:
+            raise RecordNotFoundException(
+                f"Record with primary key {pk_value} not found (no response)"
+            )
+
+        records_iter: Iterable[Record] = self._parse_response(
```
brianjlai: This section doesn't seem like it will work for a lot of cases. I feel like a common pattern is that for retrieving all objects like `prices`, we get back a response w/ a top-level array object which will contain an object for each price. And when we hit the individual API endpoint we get back a single top-level object representing the price.

The way this is written assumes that hitting the single record `fetch_path` will contain a top level list and extract each record into the `records_iter`. But now we'll be unable to extract anything, there won't be a `first_record` to extract.

Unless you think I'm missing something, it seems like we shouldn't be using the underlying extractor which is no longer suitable for individual records

unit_tests/sources/declarative/retrievers/test_simple_retriever.py
```diff
@@ -1644,3 +1645,162 @@ def _mock_paginator():
...
+    response = requests.Response()
+    response.status_code = 200
+    response._content = json.dumps({"id": "123", "title": "Test Post"}).encode("utf-8")
+
+    requester.send_request.return_value = response
+
+    record_selector = MagicMock()
+    record_selector.select_records.return_value = [
+        Record(data={"id": "123", "title": "Test Post"}, stream_name="posts", associated_slice=None)
+    ]
```
brianjlai: The way this test is written, we aren't actually testing the selector/extractor logic. Most extractor/selector are defined to extract a list of objects (Stripe almost certainly does it this way)

But here we just mock the single record response, and we mock calling `select_records()` so in a real world use case, this would fail since we'll attempt to extract `response.data` which doesn't exist. But all this is skipped over due to all the mocking

PR #971, opened 2026-04-01
fix: route RequestBodyPlainText to request_body_data instead of request_body_json

Airbyte-Support:
## Summary

`RequestBodyPlainText` was incorrectly grouped with `RequestBodyJsonObject` in `InterpolatedRequestOptionsProvider._resolve_request_body()`, causing plain text body values to be assigned to `request_body_json` instead of `request_body_data`. This resulted in a `ValueError("Request body json cannot be a string")` whenever a user configured a plain-text request body in the Connector Builder — for example, when creating an asynchronous job with a non-JSON body.

The fix splits the combined `elif` branch so that `RequestBodyPlainText` routes to `self.request_body_data` (like `RequestBodyUrlEncodedForm` already does), while `RequestBodyJsonObject` continues to route to `self.request_body_json`.

### Migration backward compatibility

The manifest migration (`HttpRequesterRequestBodyJsonDataToRequestBody`) previously converted string-valued `request_body_json` fields to `RequestBodyPlainText`. Before this fix, that worked by accident because `RequestBodyPlainText` was (incorrectly) routed to `request_body_json`. Now that `RequestBodyPlainText` correctly routes to `request_body_data`, the migration would break existing connectors that rely on string JSON templates being sent as JSON bodies.

The fix: string-valued `request_body_json` fields are now **skipped by the migration** and left as `request_body_json`, where `InterpolatedNestedRequestInputProvider` already handles them natively (interpolating the template string and parsing it via `ast.literal_eval` into a dict).

Changes to the migration:
- `should_migrate` returns `False` when the only migratable key is a string `request_body_json`
- `_migrate_body_json` returns early (no-op) for string values instead of converting to `RequestBodyPlainText`
- `validate` accepts manifests where string `request_body_json` remains unmigrated
- Removed unused `text_type` variable
- Added class and method docstrings explaining why string `request_body_json` is intentionally preserved

- [ ] **End-to-end in Connector Builder**: Configure an async job creation request with a plain-text request body and confirm the request is sent correctly without the `"Request body json cannot be a string"` error. This is the primary user-facing scenario.
- [ ] **Existing connectors with string `request_body_json`**: Confirm that connectors currently using string-valued `request_body_json` (e.g., Shopify GraphQL) continue to send JSON bodies with the correct `Content-Type` after this change. @agarctfi verified that without the migration fix, Shopify gets a `415 Unsupported Media Type`; with it, the connector works. Worth a second verification.
- [ ] **Migration `validate()` logic**: The three-state validation (`has_replacement`, `has_unmigrated`, `has_string_json`) is the most complex part of this change. Verify that manifests with only a string `request_body_json` (no other body keys) pass validation without requiring `request_body` to be present.
- [ ] **Known limitation — `ast.literal_eval` in interpolation layer**: `JinjaInterpolation` applies `ast.literal_eval` to rendered string values. A plain-text body whose content looks like a Python literal (e.g., `{"a": 1}` or `5`) could be parsed into a dict/int rather than staying a string. This is pre-existing behavior not introduced by this PR, but is now more relevant for `RequestBodyPlainText` bodies routed through `request_body_data`.

### Notes

- The old test `"test_string"` in `test_interpolated_request_json_using_request_body` was encoding the buggy behavior: it expected `RequestBodyPlainText` with value `'{"nested": { "key": "..." }}'` to produce a parsed JSON dict. This test has been removed and replaced with two new cases in `test_interpolated_request_data_using_request_body`.
- The migration test fixtures in `conftest.py` have been updated so the expected output for stream "D" (string `request_body_json`) retains the original `request_body_json` key rather than converting to `RequestBodyPlainText`.
- Verified locally by @agarctfi ([comment](https://github.com/airbytehq/airbyte-python-cdk/pull/971#issuecomment-4209374200)).
- Related oncall issue: `airbytehq/oncall#10360`

Link to Devin session: https://app.devin.ai/sessions/8a5ed7c6c890468ca63a35469e5c58c1
Requested by: @Airbyte-Support

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration and validation to stop converting string-encoded JSON into structured request bodies and to correctly distinguish plain-text vs JSON request bodies.
* **Tests**
  * Updated tests to focus on plain-text request bodies and interpolation, removed cases that treated plain-text as parsed JSON, and adjusted expected migrated manifests.

### DIFF
--- a/airbyte_cdk/manifest_migrations/migrations/http_requester_request_body_json_data_to_request_body.py
+++ b/airbyte_cdk/manifest_migrations/migrations/http_requester_request_body_json_data_to_request_body.py
@@ -26,9 +26,14 @@ class HttpRequesterRequestBodyJsonDataToRequestBody(ManifestMigration):
     replacement_key = "request_body"
 
     def should_migrate(self, manifest: ManifestType) -> bool:
-        return manifest[TYPE_TAG] == self.component_type and any(
-            key in list(manifest.keys()) for key in self.original_keys
-        )
+        if manifest[TYPE_TAG] != self.component_type:
+            return False
+        for key in self.original_keys:
+            if key in manifest:
+                if key == self.body_json_key and isinstance(manifest[key], str):
+                    continue
+                return True
+        return False
 
     def migrate(self, manifest: ManifestType) -> None:
         for key in self.original_keys:
@@ -38,9 +43,18 @@ class HttpRequesterRequestBodyJsonDataToRequestBody(ManifestMigration):
                 self._migrate_body_data(manifest, key)
 
     def validate(self, manifest: ManifestType) -> bool:
-        return self.replacement_key in manifest and all(
-            key not in manifest for key in self.original_keys
+        has_replacement = self.replacement_key in manifest
+        has_unmigrated = any(
+            key in manifest
+            for key in self.original_keys
+            if not (key == self.body_json_key and isinstance(manifest.get(key), str))
         )
+        has_string_json = self.body_json_key in manifest and isinstance(
+            manifest[self.body_json_key], str
+        )
+        if has_string_json:
+            return not has_unmigrated
+        return has_replacement and not has_unmigrated
 
     def _migrate_body_json(self, manifest: ManifestType, key: str) -> None:
         """
@@ -52,7 +66,7 @@ class HttpRequesterRequestBodyJsonDataToRequestBody(ManifestMigration):
         json_object_type = "RequestBodyJsonObject"
 
         if isinstance(manifest[key], str):
-            self._migrate_value(manifest, key, text_type)
+            return
         elif isinstance(manifest[key], dict):
             if isinstance(manifest[key].get(query_key), str):
                 self._migrate_value(manifest, key, graph_ql_type)
--- a/airbyte_cdk/manifest_migrations/migrations/http_requester_request_body_json_data_to_request_body.py
+++ b/airbyte_cdk/manifest_migrations/migrations/http_requester_request_body_json_data_to_request_body.py
@@ -15,6 +15,23 @@ class HttpRequesterRequestBodyJsonDataToRequestBody(ManifestMigration):
     This migration is responsible for migrating the `request_body_json` and `request_body_data` keys
     to a unified `request_body` key in the HttpRequester component.
     The migration will copy the value of either original key to `request_body` and remove the original key.
+
+    **String-valued `request_body_json` is intentionally left unmigrated.**
+
+    When `request_body_json` is a string (e.g. a Jinja template like
+    `'{"nested": {"key": "{{ config.option }}"}}'`), it is NOT converted to a
+    `RequestBodyPlainText` or any other typed `request_body` object. This is because:
+
+    1. The `InterpolatedRequestOptionsProvider` already handles string `request_body_json`
+       natively via `InterpolatedNestedRequestInputProvider`, which interpolates the
+       template and parses the result into a dict using `ast.literal_eval`, then sends
+       it as a JSON body.
+    2. Converting it to `RequestBodyPlainText` would route it to `request_body_data`
+       (raw string body) instead of `request_body_json` (JSON body), breaking connectors
+       that rely on the body being sent as JSON with the correct Content-Type header.
+    3. We cannot convert it to `RequestBodyJsonObject` because migrations run before
+       interpolation, so Jinja templates have not been resolved yet and the string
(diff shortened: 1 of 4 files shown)


agarctfi:
I had Devin do a mock test to see if the fix works, and it does. Also verified it locally.

darynaishchenko:
> The manifest migration (HttpRequesterRequestBodyJsonDataToRequestBody) previously converted string-valued request_body_json fields to RequestBodyPlainText. Before this fix, that worked by accident because RequestBodyPlainText was (incorrectly) routed to request_body_json. Now that RequestBodyPlainText correctly routes to request_body_data, the migration would break existing connectors that rely on string JSON templates being sent as JSON bodies.

@agarctfi Do we have such connectors? If we don't I think we can just go with the pure fix in InterpolatedRequestOptionsProvider without the migration.

agarctfi:
@darynaishchenko On Cloud, there is 1 connection, not sure about OSS or if we would even count that. 

[Here](https://app.devin.ai/sessions/1daa3505231c474e9685b7fc765ce0c1) is where I checked.

darynaishchenko:
> Daryna Ishchenko (@darynaishchenko) On Cloud, there is 1 connection, not sure about OSS or if we would even count that.
> 
> [Here](https://app.devin.ai/sessions/1daa3505231c474e9685b7fc765ce0c1) is where I checked.

@agarctfi could you please verify if this connection will actually fail after applying a cdk fix? I think because of this https://github.com/airbytehq/airbyte-python-cdk/pull/971#discussion_r3053997058 it should work fine. Were you able to reproduce the issue that is handled in the migration change?

agarctfi:
@darynaishchenko I checked, and that connector is for Shopify, so I used our test creds. Without the migration, we get a `415 Client Error: Unsupported Media Type for url: https://airbyte-integration-test.myshopify.com/admin/api/2024-01/graphql.json`

With the migration, we don't have this issue, so I think we should leave it. WDYT? 

[Here](https://app.devin.ai/sessions/1daa3505231c474e9685b7fc765ce0c1?tab=shopify_output_without_migration.json) is where I tested.

darynaishchenko:
> Daryna Ishchenko (@darynaishchenko) I checked, and that connector is for Shopify, so I used our test creds. Without the migration, we get a `415 Client Error: Unsupported Media Type for url: https://airbyte-integration-test.myshopify.com/admin/api/2024-01/graphql.json`
> 
> With the migration, we don't have this issue, so I think we should leave it. WDYT?
> 
> [Here](https://app.devin.ai/sessions/1daa3505231c474e9685b7fc765ce0c1?tab=shopify_output_without_migration.json) is where I tested.

@agarctfi Thanks for testing it!

1. Now the migration logic is changed in a way to keep the manifest as is and do not migrate to `request_body` if `request_body_json` is a string. Is there any risk in it (regarding the migration to `request_body`)? 
2. Should we don't just skip (do not migrate `request_body` if `request_body_json` is a string) but also try to migrate it to `RequestBodyJsonObject`? We can't simply convert it because the migration runs before interpolation, so the templates haven't been resolved yet. But it's related question to 1 about how it's better to handle it.
3. We should add doc string explaining "`request_body_json` is a string" case.

agarctfi:
>1. Now the migration logic is changed in a way to keep the manifest as is and do not migrate to `request_body` if `request_body_json` is a string. Is there any risk in it (regarding the migration to `request_body`)? 
>2. Should we don't just skip (do not migrate `request_body` if `request_body_json` is a string) but also try to migrate it to `RequestBodyJsonObject`? We can't simply convert it because the migration runs before interpolation, so the templates haven't been resolved yet. But it's related question to 1 about how it's better to handle it.
>3. We should add doc string explaining "`request_body_json` is a string" case.

Yes, it looks like Devin added the docstring, which looks good to me. Let me know if we need anything else for this one. @darynaishchenko , thanks!

darynaishchenko (approved):

pickfire:
This caused an issue for us (it used to work in 7.17.0 but not in 7.17.1).

```yaml
        request_body:
          type: RequestBodyPlainText
          value: "{\"id\": \"{{ config['id'] }}\"}"
```

Previously we use plain text since we need to put in the body `{"id": "1"}` instead of `{"id": 1}` which the server can't accept. So that's why plain text was used. With this update:

- connector builder still sends it as JSON
- but real connector no longer works, it sends it as form
- I tried JSON Object Body with `| string`, but it still failed in connector builder since it will get converted to integer

The fix: I went with `| tojson` so it works now using JSON Object Body.

PR #999, opened 2026-04-25
feat(test): auto-discover smoke-test scenarios from metadata.yaml

aaronsteers:
## Summary

Teaches the standard-tests suite to read smoke-test scenarios declared in a connector's `metadata.yaml`, so connectors can opt into 1Password-backed scenarios without per-connector CI YAML edits.

New module `airbyte_cdk.test.standard_tests.scenario_loader`:

- Parses `data.connectorTestSuitesOptions[].suite == 'smokeTests'` from `metadata.yaml`
- Converts each scenario into a `ConnectorTestScenario` (with `config_dict` populated and a stable `smoke-<name>` id)
- Resolves two sigils inside `configSettings` recursively:
  - `${secret-ref:<item>/<section>/<field>}` — 3-segment-only. Reads `secrets/.env.<section>` (literal 1P section name, no prefix stripping) and looks up `<ITEM_UPPER>_<FIELD_UPPER>`. The OPS CLI is responsible for writing those `.env.<section>` files ahead of test execution.
  - `${relative-date:-P<duration>}` — ISO-8601 duration anchored at today midnight UTC, rendered as RFC 3339 (`YYYY-MM-DDTHH:MM:SSZ`).
- Validates the provider env-var contract (`AIRBYTE_TEST_CREDS_PROVIDER=1pass`, `AIRBYTE_TEST_CREDS_1PASS_VAULT_NAME`) **only when** a `${secret-ref:...}` sigil is actually present. Pure literal or `${relative-date:...}`-only scenarios (source-faker style) resolve with no provider env vars.

`DockerConnectorTestSuite.get_scenarios()` now returns CAT-derived scenarios + smoke-test scenarios. Smoke-test scenarios are additive — connectors that have not opted in see no behavior change.

Companion to:
- `airbytehq/airbyte-connector-models#42` — schema for the metadata.yaml smoke-test dialect (literal section names, 3-segment sigils only)
- `airbytehq/airbyte-ops-mcp#733` — current `secrets fetch` flow; a follow-up will teach it to write per-section `.env.<section>` files when a connector has a smoke-test block

- [ ] Confirm the sigil shape matches the locked design: 3-segment only, section is the literal 1P section name (e.g. `test-credentials-api_key`), no `.env.default` fallback.
- [ ] Confirm the env-var contract feels right: `AIRBYTE_TEST_CREDS_PROVIDER` + `AIRBYTE_TEST_CREDS_1PASS_VAULT_NAME` required only when secret-ref sigils are present. The vault env var is read by the OPS CLI; CDK only checks it is set, since CDK reads pre-fetched `.env.<section>` files rather than calling `op` itself.
- [ ] Spot-check `_load_env_file` parsing — supports `KEY=value` and `KEY="value"` / `KEY='value'`, ignores comments and blank lines. Multi-line values inside quotes are NOT supported here (the OPS CLI multi-line writer in #733 produces them, but at parse time the test runner only consumes single-line values; flag if this is wrong).
- [ ] Try a local end-to-end run once a connector adopts the dialect: write `secrets/.env.test-credentials-api_key`, set the two env vars, and confirm `pytest` parametrizes a `smoke-default` scenario.

### Notes

- 14 unit tests cover: missing metadata, no smoke-test suite, literal-only resolves without env vars, relative-date math, secret-ref resolution, literal section name preservation, missing provider env vars, missing vault name, missing `.env.<section>` file, missing var inside file, unknown sigil, 2-segment sigil rejected, multi-scenario, recursive nested resolution.
- Held intentionally out of scope: OPS CLI changes to write per-section `.env.<section>` files, monorepo CI workflow updates to add the env vars, and the source-ashby `metadata.yaml` block. Each will follow as its own PR after this one is reviewed.
- Drafted as draft per project convention; not for merge without explicit approval.

Link to Devin session: https://app.devin.ai/sessions/5c2ccf537b70486b9e6fbc974d26063e
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Test scenarios can now be loaded from connector metadata files alongside traditional configuration files.
  * Added support for dynamic secret and relative date resolution in test configurations using specialized notation.

* **Tests**
  * Comprehensive test coverage added for scenario discovery, secret injection, and date handling.

### DIFF
--- a/airbyte_cdk/test/standard_tests/docker_base.py
+++ b/airbyte_cdk/test/standard_tests/docker_base.py
@@ -27,6 +27,9 @@ from airbyte_cdk.models import (
 from airbyte_cdk.models.connector_metadata import MetadataFile
 from airbyte_cdk.test.entrypoint_wrapper import EntrypointOutput
 from airbyte_cdk.test.models import ConnectorTestScenario
+from airbyte_cdk.test.standard_tests.scenario_loader import (
+    load_metadata_smoke_test_scenarios,
+)
 from airbyte_cdk.utils.connector_paths import (
     ACCEPTANCE_TEST_CONFIG,
     find_connector_root,
@@ -122,9 +125,25 @@ class DockerConnectorTestSuite:
     ) -> list[ConnectorTestScenario]:
         """Get acceptance tests for a given category.
 
+        Scenarios are sourced from two files, in this order:
+
+        1. `acceptance-test-config.yml` (CAT-style scenarios), parsed as today.
+        2. `metadata.yaml` `connectorTestSuitesOptions[].suite == 'smokeTests'`,
+           with `${secret-ref:...}` and `${relative-date:...}` sigils resolved.
+
+        Smoke-test scenarios are additive; CAT-style scenarios are unchanged
+        for any connector that has not opted into the metadata.yaml dialect.
+
         This has to be a separate function because pytest does not allow
         parametrization of fixtures with arguments from the test class itself.
         """
+        cat_scenarios = cls._get_acceptance_test_scenarios()
+        smoke_scenarios = cls._get_smoke_test_scenarios()
+        return cat_scenarios + smoke_scenarios
+
+    @classmethod
+    def _get_acceptance_test_scenarios(cls) -> list[ConnectorTestScenario]:
+        """Load CAT-style scenarios from `acceptance-test-config.yml`."""
         try:
             all_tests_config = cls.acceptance_test_config
         except FileNotFoundError as e:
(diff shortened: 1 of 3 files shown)


airbyte_cdk/test/standard_tests/scenario_loader.py
```diff
@@ -0,0 +1,288 @@
...
+        day=now.day,
+        tzinfo=timezone.utc,
+    )
+    rendered: datetime = today_midnight - duration
+    if rendered.tzinfo is None:
+        rendered = rendered.replace(tzinfo=timezone.utc)
+    return rendered.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def _load_env_file(
```
aaronsteers, line 224: Shouldn't we use the python dotenv library rather than rebuild this from scratch??

airbyte_cdk/test/standard_tests/docker_base.py
```diff
@@ -158,9 +177,17 @@ def get_scenarios(
...
-        return deduped_test_scenarios
+    @classmethod
+    def _get_smoke_test_scenarios(cls) -> list[ConnectorTestScenario]:
+        """Load smoke-test scenarios declared in `metadata.yaml`, if any."""
+        connector_root = cls.get_connector_root_dir()
+        metadata_path = connector_root / "metadata.yaml"
+        return load_metadata_smoke_test_scenarios(
+            metadata_path=metadata_path,
+            connector_root=connector_root,
+        )
```
aaronsteers: We should try to use the generated pydantic model from the connector models repo to deserialize and validate the metadata file and then parse smoke test from that parsed pydantic model. If we run into blockers for that route, we should discuss.

PR #1016, opened 2026-05-07
feat(async-jobs): add deferred retry with cooldown for failed async jobs

darynaishchenko:
## Summary

Adds an optional `failed_retry_wait_time_in_seconds` parameter to `AsyncRetriever` that defers retry of **real** FAILED async jobs (report created on the API but returned a failure status like Amazon SP-API's FATAL) until a cooldown period elapses — without blocking other jobs or using `time.sleep()`.

**Key design decisions:**
- **Only real FAILED jobs are deferred.** Creation-failure jobs (HTTP 429 when _starting_ a job — report never created) and `TIMED_OUT` jobs are retried immediately, preserving existing behavior.
- `AsyncJob` distinguishes the two via an `is_creation_failure` constructor parameter (no private-attribute writes from outside the class).
- The orchestrator uses a two-phase check: first arms `retry_after` on initial failure, then skips until the cooldown elapses.
- Exposed on the declarative schema with `minimum: 1` validation and config interpolation support.

**Related:** [airbytehq/airbyte#77837](https://github.com/airbytehq/airbyte/pull/77837) configures `failed_retry_wait_time_in_seconds: 1800` for source-amazon-seller-partner.

- [ ] **Trace the deferral state machine in `_replace_failed_jobs`** — the two guards (`ready_to_retry` then `retry_deferred`) must be checked in this order. On first failure both are false → second branch arms cooldown. Before cooldown expires, `ready_to_retry()` is false → first branch skips. After cooldown, both true → falls through to replacement. Verify no ordering produces unexpected behavior.
- [ ] **Verify creation-failure bypass** — confirm that `_create_failed_job()` passes `is_creation_failure=True` and that the condition `not job.is_creation_failure()` correctly gates deferral so 429-driven retries are never delayed.
- [ ] **Run tests**: `pytest unit_tests/sources/declarative/async_job/ -v` (43 tests). The regression test `test_given_real_failed_then_cooldown_elapses_then_start_returns_creation_failure_then_no_rearm` covers: real FAILED → arm cooldown → cooldown elapses → replacement is creation-failure → no re-arm on next tick.

### Notes
- The `<= 0` runtime guard in the orchestrator constructor is the only enforcement beyond the YAML `minimum: 1`; Pydantic model uses `Union[int, str]` without constraints.
- Tests access `job_tracker._jobs` (private set) directly to pre-register job IDs.

Link to Devin session: https://app.devin.ai/sessions/2cbe53a30ea14f9f8151f407f423283b

## Summary by CodeRabbit

* **New Features**
  * Added deferred retry scheduling for failed async jobs with configurable wait times (`failed_retry_wait_time_in_seconds`).
  * Enhanced job failure tracking to distinguish between creation-time failures and runtime failures.
  * Jobs now support retry deferral logic with automatic readiness checks based on elapsed time.

* **Tests**
  * Expanded test coverage for retry and cooldown behavior in async job orchestration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/airbytehq/airbyte-python-cdk/pull/1016)

### DIFF
--- a/airbyte_cdk/sources/declarative/async_job/job.py
+++ b/airbyte_cdk/sources/declarative/async_job/job.py
@@ -25,6 +25,7 @@ class AsyncJob:
         self._job_parameters = job_parameters
         self._status = AsyncJobStatus.RUNNING
         self._retry_after: Optional[datetime] = None
+        self._is_synthetic = False
 
         timeout = timeout if timeout else timedelta(minutes=60)
         self._timer = Timer(timeout)
@@ -54,6 +55,10 @@ class AsyncJob:
             self._timer.stop()
 
         self._status = status
+
+    def is_synthetic(self) -> bool:
+        """Return True if this job was never actually created on the API side."""
+        return self._is_synthetic
 
     def set_retry_after(self, retry_after: datetime) -> None:
         """Set the earliest time this job can be retried."""
--- a/airbyte_cdk/sources/declarative/async_job/job.py
+++ b/airbyte_cdk/sources/declarative/async_job/job.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2024 Airbyte, Inc., all rights reserved.
 
 
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Optional
 
 from airbyte_cdk.sources.declarative.async_job.timer import Timer
@@ -24,6 +24,7 @@ class AsyncJob:
         self._api_job_id = api_job_id
         self._job_parameters = job_parameters
         self._status = AsyncJobStatus.RUNNING
+        self._retry_after: Optional[datetime] = None
 
         timeout = timeout if timeout else timedelta(minutes=60)
         self._timer = Timer(timeout)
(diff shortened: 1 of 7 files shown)


tolik0 (approved):

tolik0:
A couple of concerns from review:

### 1. Double cooldown via synthetic FAILED jobs

`_start_job` doesn't always return a freshly-running job. When the API rejects the start, `_keep_api_budget_with_failed_job` synthesizes a `FAILED` `AsyncJob` with `_retry_after = None` and returns it (`job_orchestrator.py:260-281`). Tracing the SP-API cooldown path:

1. Job fails at `T=0` → cooldown armed, `retry_after = T + 1800`
2. At `T=1800`, cooldown elapses → `_replace_failed_jobs` calls `_start_job`
3. API still in cooldown → `_start_job` returns a synthetic `FAILED` job (fresh, no `_retry_after`)
4. Next tick → state machine sees a fresh `FAILED` job → arms cooldown again at `T ≈ 1800`
5. Effective total wait is roughly `2 × cooldown` per attempt, not `1 × cooldown`

For the motivating SP-API case, hitting `_keep_api_budget_with_failed_job` is the *expected* outcome of retrying during cooldown, so this isn't an edge case. Two reasonable fixes:

- Carry `_retry_after` from the old job onto the synthetic replacement, or
- Don't replace at all when `_start_job` would synthesize a failure during cooldown — keep the original job and re-arm.

Either way, worth a regression test that simulates `_start_job` returning a synthetic FAILED job and asserts the total wall-clock wait equals one cooldown, not two.

### 2. Three different lower bounds for `failed_retry_wait_time_in_seconds`

- YAML schema: `minimum: 1`
- Constructor: `if ... < 0: raise ValueError(...)` — allows `0`
- Runtime: `if self._failed_retry_wait_time_in_seconds:` — treats `0` as disabled

A value of `0` is rejected by the schema, silently accepted by the constructor, and silently disables the feature at runtime. Pick one: `is not None and > 0` in the runtime check, and `> 0` (or `>= 1`) in the constructor to match the YAML.

PR #1049, opened 2026-06-11
feat: add JsonItemsDecoder for streaming large JSON responses

tolik0:
## What

Adds a streaming JSON decoder (`JsonItemsDecoder`) for very large single-document JSON responses where the records of interest live under a nested array. `JsonItemsParser` yields each array element via [`ijson`](https://pypi.org/project/ijson/), so peak memory is bounded by a single record instead of the whole document.

**Adopts and supersedes #1026** (whose authoring Devin session can no longer respond to review).

## How

- `JsonItemsParser` in `composite_raw_decoder.py` — `ijson.items(data, f"{items_path}.item")`. Composes with the existing `CompositeRawDecoder` hierarchy (gzip/zip).
- `JsonItemsDecoder` added to `declarative_component_schema.yaml` and wired into the decoder unions; pydantic models regenerated; factory `create_json_items_decoder` registered.
- `ijson` added as a first-class CDK dependency (dropped from the deptry ignore list).
- `JsonItemsParser` honors a configured non-UTF-8 `encoding` by transcoding to UTF-8 bytes through a lazy streaming recoder, keeping ijson on its fast byte backend (text-mode input is deprecated in ijson) — verified to stay bounded-memory.

## Scope note (changed from the original revision)

An earlier revision of this PR also added an `always_decompress` flag to `GzipDecoder`, to handle Amazon Seller Partner reports that return gzip bodies mislabeled as `Content-Encoding: identity`. That turned out to be unnecessary: the consuming connector can instead request `enableContentEncodingUrlHeader=true` on `getReportDocument`, which makes Amazon set an honest `Content-Encoding: gzip` (when compressed) / `identity` (when not), so the **stock header-based `GzipDecoder` already handles both cases**. The gzip change has been dropped to keep this PR focused on `JsonItemsDecoder`.

## Validation

Validated against a real 3.45 GB Amazon Brand Analytics report (`GzipDecoder{JsonItemsDecoder}`, with honest gzip headers): **8,533,151 records, ~150 MB peak RSS** (vs ~10 GB for the buffering decoder), correct UTF-8.

## Tests

`unit_tests/.../decoders/test_composite_decoder.py` — JsonItems cases (top-level/nested/empty/encoding/gzip-composition/missing-path/lazy). 41 decoder tests pass.

> The generated `declarative_component_schema.py` was edited by hand (Dagger/Docker unavailable locally for codegen) — please confirm via the CI regen check.

## Release Notes

* **New Features**
  * Added streaming JSON decoder for extracting items from nested JSON arrays without loading entire responses into memory, with configurable encoding support and compatibility with existing decoders (gzip, zipfile, etc.).

* **Tests**
  * Added comprehensive test coverage for the new JSON streaming decoder functionality.


tolik0:
Thanks for the reviews — addressed in e687ea1a:

- **`JsonItemsParser.encoding` now honored** (CodeRabbit/Copilot): non-UTF-8 encodings are transcoded to UTF-8 bytes via a small lazy streaming recoder, so ijson stays on its native byte backend (text-mode input is deprecated in ijson) and memory stays bounded. Verified on a real 3.45GB report: peak RSS unchanged (~14MB for the parser in isolation), same `yajl2_c` backend.
- **`always_decompress` now respected in connector-builder mode** (Copilot): the `_emit_connector_builder_messages` early-return previously skipped it; it now keeps the `GzipParser` (with `stream_response=False`) when the flag is set.
- **Tests**: the encoding test now uses `ensure_ascii=False` so the non-ASCII bytes genuinely differ per encoding; the `always_decompress` negative test asserts `ijson.JSONError` instead of bare `Exception`.

Left as-is (happy to change if you prefer): `items_path` keeps a `""` default with a fast-fail runtime check, and `encoding` stays `Optional[str]` — both minor and would touch the existing requires-items_path test.

tolik0:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/27346896516
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/27346905256

darynaishchenko (approved):

tolik0:
Addressed the latest review feedback and rebased onto current main (v7.22.0):

- **`_Utf8Recoder` read contract** (Copilot, line 116): switched to a streaming `codecs.getincrementaldecoder`, reading byte-sized chunks from the underlying stream (handles multi-byte chars split across reads). ijson consumes the re-encoded UTF-8 bytes on its native byte backend; verified bounded memory + no deprecation warning.
- **UTF-8 alias detection** (Copilot, line 144): now uses `codecs.lookup(self.encoding).name != "utf-8"`, so aliases like `utf_8` / `u8` / `utf8` correctly skip transcoding instead of taking the recode path.

41 decoder tests pass.

tolik0:
Addressed: `create_json_items_decoder` now passes `encoding=model.encoding` directly (dropped the `or "utf-8"` coercion), so an explicit `encoding: null` is preserved. Verified `JsonItemsParser(encoding=None)` parses correctly (falls through to ijson's native byte/auto-detect path). 41 decoder tests pass.

PR #1060, opened 2026-06-25
fix: widen cryptography cap to allow >=48.0.1 and relax stale pytz pin

mcgizzle:
## What

Remove the `<47.0.0` upper bound on `cryptography` (the `!=45.0.0,!=45.0.1` regression exclusions are kept) and relax `pytz` from the exact `2024.2` pin to `>=2024.2`.

## Why

The `<47.0.0` cap forces consumers onto `cryptography==46.0.7`, which carries a HIGH-severity advisory (SNYK-PYTHON-CRYPTOGRAPHY-17344551): an out-of-bounds read in the ASN.1 decoder reachable via `d2i_X509()` / `d2i_PKCS7()`. The remediation ships in `cryptography==48.0.1` (a bundled-OpenSSL bump), which the current cap forbids.

The cap is not transitive-forced. The only packages in the tree that depend on `cryptography` (`google-auth` via `google-cloud-secret-manager`, `pdfminer.six`, `pyjwt`) impose lower bounds only. The earlier "Constrained as transitive dependency" note was already removed in #975; this change lifts the remaining arbitrary upper bound.

## Verification

- `poetry lock` followed by `poetry update cryptography` resolves the full dependency set to `cryptography==49.0.0` with no conflict; the lock diff is a single package bump (46.0.7 to 49.0.0).
- The declarative JWT auth tests (`unit_tests/sources/declarative/auth/`) pass against 49.0.0 (99 passed).

## Summary by CodeRabbit

* **Chores**
  * Updated dependency version constraints to allow newer supported releases.
  * Broadened the allowed range for one timezone library and removed the upper limit from the cryptography library, while still avoiding known problematic releases.
  * Aligned internal dependency-check settings with the updated version rules.


mcgizzle:
@aaronsteers the red checks here are all fork-secret limitations, not problems with the change.

The five connector jobs (source-intercom, source-pokeapi, source-shopify, source-hardcoded-records, destination-motherduck) all fail at the same line fetching test creds from GSM: "No Google Cloud credentials found", because GCP_GSM_CREDENTIALS isn't exposed to fork PRs. The "PR Welcome Message" job fails with a 403 "Resource not accessible by integration" for the same reason: the fork token is read-only.

Everything that validates the actual change passed: full pytest matrix (3.10 to 3.13), MyPy, Ruff, Deptry, and both Docker image builds. The connector jobs got past dependency resolution and CDK install before hitting the secret fetch, so the widened cryptography / relaxed pytz pins don't break installs.

Could you re-run the connector tests with secrets, or merge if the green checks are sufficient? Happy to rebase if needed.

aaronsteers:
@mcgizzle - Thanks very much for contributing to Airbyte. I'm evaluating these changes and will create a new PR shortly, possibly with a few tweaks. I'll include your commits in my branch so you retain credit when we do merge. 

If you have any testing data specifically from successful testing you might have done locally, please feel free to share that here or on the new PR when it opens. Much appreciated!

aaronsteers:
Follow here https://github.com/airbytehq/airbyte-python-cdk/pull/1061:
- https://github.com/airbytehq/airbyte-python-cdk/pull/1061

aaronsteers:
Above PR has been merged and released:

- https://github.com/airbytehq/airbyte-python-cdk/releases/tag/v7.23.4

@mcgizzle - Thank you again for this contribution 🙏

PR #1063, opened 2026-07-01
feat(file-based): bump unstructured to 0.18.x with custom markdown preserved

aaronsteers:
## Summary

Bumps `unstructured` from `0.10.27` to `>=0.18.27,<0.19` (resolves 0.18.32). Custom markdown rendering (`_render_markdown`/`_convert_to_markdown`) is **preserved** — upstream's `elements_to_md()` doesn't handle heading depths, list-item bullets, or formula code-blocks.

Migrates `FileType` API from removed dict lookups to enum methods:
```python
FILETYPE_TO_MIMETYPE[ft]  →  ft.mime_type
STR_TO_FILETYPE.get(m)    →  FileType.from_mime_type(m)   # None *or* FileType.UNK are treated as inconclusive
EXT_TO_FILETYPE.get(e)    →  FileType.from_extension(e)
detect_filetype(filename=) → detect_filetype(file_path=)
```
`from_mime_type("application/octet-stream")` returns `FileType.UNK` in 0.18.x (0.10.x had no entry, so it fell through). `_get_filetype` now skips `UNK` and falls back to path/content/extension detection; regression test added.

Pins `onnxruntime>=1.23,<1.24` — transitive dep of `unstructured[pdf]`; 1.24.x ships no cp310 wheels, >=1.25 requires Python >=3.11.

Partition functions typed as `Optional[Callable[..., Any]]`, imported lazily in `_import_unstructured()`; `_read_file_locally` raises `Exception("unstructured library is not available")` (same message as before) if they're unset rather than `assert`ing. Defensive `else` raises `ValueError` for unsupported file types.

`poetry.lock` was re-locked against `main` with `poetry lock --no-update` + `poetry update pdfminer.six` so only `unstructured`, `pdfminer.six` and their new transitive deps change (no drift of `requests-cache`, mypy stubs, etc.).

Scenario test expectations updated: corrupted-PDF error message changed from `"No /Root object!"` to `"Unable to get page count"` in 0.18.x, and PDF content classified as `NarrativeText` (not `Title`) so no `#` prefix.

Resolves https://github.com/airbytehq/oncall/issues/11267:
- https://github.com/airbytehq/oncall/issues/11267

Security: this bump remediates GHSA-gm8q-m8mv-jj5m / CVE-2025-64712 (`unstructured` < 0.18.18) for `source-s3`, `source-gcs`, and `source-azure-blob-storage`.

CI notes: `Check: destination-motherduck` fails on every recent PR (MotherDuck token auth); `Check: source-google-drive` fails with the same `SourceGoogleDrive.__init__() missing … 'catalog', 'config', 'state'` harness error on an unrelated file-based PR (run 32184488735, Aug 18), and this PR does not touch `airbyte_cdk/test/`. Flagging both for maintainer judgment rather than asserting they're unrelated.

Link to Devin session: https://app.devin.ai/sessions/a7c33e3a2ebc423d99df7a28c0f585ee
Open in Devin Desktop: https://app.devin.ai/desktop/session/a7c33e3a2ebc423d99df7a28c0f585ee?variant=devin


airbyte_cdk/sources/file_based/file_types/unstructured_parser.py
```diff
@@ -80,18 +74,33 @@ def optional_decode(contents: Union[str, bytes]) -> str:
...
-    unstructured_partition_pptx = partition_pptx
+        unstructured_partition_pptx = partition_pptx
+    except ImportError:
+        logging.debug("PPTX partition module unavailable; PPTX parsing will be disabled.")
+    try:
+        from unstructured.partition.pdf import partition_pdf
+
+        unstructured_partition_pdf = partition_pdf
+    except ImportError:
+        logging.debug("PDF partition module unavailable; PDF parsing will be disabled.")
```
aaronsteers: I feel like we should fail loudly and perhaps just use normal imports.

airbyte_cdk/sources/file_based/file_types/unstructured_parser.py
```diff
@@ -80,19 +74,30 @@ def optional_decode(contents: Union[str, bytes]) -> str:
...
-    unstructured_partition_pdf = partition_pdf
     unstructured_partition_docx = partition_docx
     unstructured_partition_pptx = partition_pptx
 
+    try:
+        from unstructured.partition.pdf import partition_pdf
+
+        unstructured_partition_pdf = partition_pdf
+    except ImportError:
+        logging.debug("PDF partition module unavailable; install unstructured[pdf] to enable.")
```
aaronsteers: No inline imports.

aaronsteers: Oh, wait. Nvm. I see we are already doing so in the prior version. Just no try-catch blocks.

aaronsteers: You have to assume the connector will always have the libraries it needs. If it doesn't, or if another connector accesses this code path without having the libraries installed, both are critical bugs in our implementation - neither should be silenced/supressed/mitigated or hidden. No silent failures.

unit_tests/sources/file_based/scenarios/unstructured_scenarios.py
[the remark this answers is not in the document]
aaronsteers: Wrong. Assume `unstructured[pdf]` and `torch` and `unstructured-inference` are always on the image and always installed when running these connectors.

Not acceptable to lose PDF processing capabilities.

aaronsteers:
Devin, fyi, take note I retitled the PR and don't override it please. Do let me know if my "latest" annotation on this version ref is not accurate. (Ideally it is the latest.)

PR #1066, opened 2026-07-06
feat(low-code): add RateLimitedMultipleTokenAuthenticator component

darynaishchenko:
## Summary

Adds a new declarative authenticator, `RateLimitedMultipleTokenAuthenticator`, that generalizes source-github's `MultipleTokenAuthenticatorWithRateLimiter` so it can be reused by any connector during declarative migrations (context: https://github.com/airbytehq/airbyte-internal-issues/issues/16512).

Behavior (per outgoing request, via `AuthBase.__call__`):
- Requests are classified into named quota pools (`quotas`) via `HttpRequestRegexMatcher`s; a pool with no matchers is the default (generalizes GitHub's hardcoded `"graphql" in path` REST-vs-GraphQL split).
- Each token tracks a local per-pool counter, seeded once per token from `quota_status_source` (e.g. GitHub `GET /rate_limit`) using `remaining_path`/`reset_path`/`limit_path` dpath extraction. Seeding is lazy (first request) and repeated after an exhaustion wait.
- Active token's counter is decremented; on exhaustion the authenticator rotates through tokens; when all are drained it sleeps until the earliest reset (heartbeat logs, floored at `MIN_EXHAUSTION_WAIT` = 5s so stale reset timestamps can't busy-loop the quota endpoint, bounded by `max_wait_time`, else raises `transient_error`).
- Proactive throttling: once every token drops below `max(budget_min_reserve, budget_reserve_fraction * limit)`, delays of `seconds_to_reset / total_remaining` (capped at 10s) are injected for graceful degradation instead of a hard wall.

Thread safety (new vs. the source-github implementation, required for the concurrent CDK):
- Token selection + decrement happen atomically under an `RLock`; `_acquire_call` returns the token chosen under the lock and the request header is built from that value, so a request is always signed with the token whose counter was decremented (no TOCTOU with concurrent rotation). All sleeps occur outside the lock.
- Double-checked re-seed via a separate `_refresh_lock` so only one thread refreshes counters after an exhaustion wait.
- `ModelToComponentFactory.create_rate_limited_multiple_token_authenticator` caches instances keyed on the model definition, so all streams share one counter set (same idea as the shared `api_budget`).

Example manifest usage (source-github):
```yaml
authenticator:
  type: RateLimitedMultipleTokenAuthenticator
  tokens: "{{ config['credentials']['personal_access_token'] }}"  # comma-separated PATs, or a list
  auth_method: "token"
  quota_status_source:
    type: QuotaStatusSource
    url: "{{ config.get('api_url', 'https://api.github.com') }}/rate_limit"
  quotas:
    - type: TokenQuota
      name: rest
      remaining_path: ["resources", "core", "remaining"]
      reset_path: ["resources", "core", "reset"]
      limit_path: ["resources", "core", "limit"]
    - type: TokenQuota
      name: graphql
      remaining_path: ["resources", "graphql", "remaining"]
      reset_path: ["resources", "graphql", "reset"]
      matchers:
        - type: HttpRequestRegexMatcher
          url_path_pattern: "/graphql"
```

Changes:
- `declarative_component_schema.yaml`: new `RateLimitedMultipleTokenAuthenticator`, `QuotaStatusSource`, `TokenQuota` definitions; added to `HttpRequester.authenticator` and `SelectiveAuthenticator.authenticators` unions (models regenerated; new classes applied on top of the checked-in file to avoid unrelated codegen churn).
- `airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py`: runtime implementation.
- `model_to_component_factory.py`: `create_rate_limited_multiple_token_authenticator` with token-string splitting, interpolation, and shared-instance caching.
- Unit tests: seeding/header injection, quota matching, rotation, exhaustion wait/raise, budget throttling, thread safety (no lost decrements across 200 threads; signed token matches decremented token under concurrent rotation), factory token parsing and instance sharing.

Requested by Daryna Ishchenko (darynaishchenko).

Link to Devin session: https://app.devin.ai/sessions/3ced8f6a11d44ed4abc517221294f767

## Summary by CodeRabbit

* **New Features**
  * Added a declarative authenticator that rotates across multiple tokens based on quota availability.
  * Added configurable quota pools, request matching, proactive throttling, automatic quota refresh, and reset-aware waiting.
  * Enabled the authenticator in declarative HTTP requester and selective authentication configurations.
  * Added support for configuring tokens, quota status endpoints, headers, and throttling limits.
* **Bug Fixes**
  * Improved authentication specification generation to preserve configured values and remain consistent across repeated calls.

### DIFF
--- a/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
+++ b/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
@@ -171,8 +171,8 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
                 token = self._active_token
                 state = self._states[token][quota.name]
                 if state.remaining > 0:
-                    budget_delay = self._compute_budget_delay(quota)
                     state.remaining -= 1
+                    budget_delay = self._compute_budget_delay(quota)
                 elif all(self._states[token][quota.name].remaining <= 0 for token in self._tokens):
                     min_time_to_wait = min(
                         (
--- a/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
+++ b/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
@@ -65,6 +65,7 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
 
     HEARTBEAT_INTERVAL = 60.0  # Log every 60s during exhaustion wait
     MAX_BUDGET_DELAY = 10.0  # Cap for the per-request proactive throttling delay
+    MIN_EXHAUSTION_WAIT = 5.0  # Floor for the exhaustion wait, so stale reset timestamps can't cause a refresh busy-loop
 
     def __init__(
         self,
@@ -126,14 +127,15 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
 
     @property
     def token(self) -> str:
-        return f"{self._auth_method} {self._active_token}".strip()
+        with self._lock:
+            return f"{self._auth_method} {self._active_token}".strip()
 
     def __call__(self, request: requests.PreparedRequest) -> Any:
         """Attach the HTTP headers required to authenticate on the HTTP request"""
         self._ensure_initialized()
         quota = self._match_quota(request)
-        self._acquire_call(quota)
-        request.headers.update(self.get_auth_header())
+        token = self._acquire_call(quota)
+        request.headers[self._header] = f"{self._auth_method} {token}".strip()
         return request
 
     def _ensure_initialized(self) -> None:
@@ -152,14 +154,22 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
                     return quota
             elif default_quota is None:
                 default_quota = quota
+        if default_quota is None:
+            self._logger.debug(
+                "Request %s did not match any quota pool; falling back to '%s'. Consider defining a matcher-less default pool.",
+                request.url,
+                self._quotas[0].name,
+            )
         return default_quota or self._quotas[0]
 
-    def _acquire_call(self, quota: TokenQuota) -> None:
+    def _acquire_call(self, quota: TokenQuota) -> str:
+        """Reserve one call from the matched pool and return the token it was charged to."""
         while True:
             budget_delay: Optional[float] = None
             wait_for_reset: Optional[float] = None
             with self._lock:
-                state = self._states[self._active_token][quota.name]
+                token = self._active_token
+                state = self._states[token][quota.name]
                 if state.remaining > 0:
                     budget_delay = self._compute_budget_delay(quota)
                     state.remaining -= 1
@@ -176,7 +186,7 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
                             internal_message=f"Rate limits for all tokens (quota: {quota.name}) were reached and the next reset exceeds max_wait_time",
                             message="Rate limit is exceeded for all provided tokens.",
                         )
-                    wait_for_reset = max(min_time_to_wait, 0)
+                    wait_for_reset = max(min_time_to_wait, self.MIN_EXHAUSTION_WAIT)
                 else:
                     self._active_token = next(self._tokens_iter)
                     continue
@@ -200,7 +210,7 @@ class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
                     )
                     self._budget_logged = True
                 time.sleep(budget_delay)
-            return
+            return token
 
     def _compute_budget_delay(self, quota: TokenQuota) -> Optional[float]:
         """Compute the proactive throttling delay. Must be called while holding the lock."""
--- a/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
+++ b/airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
@@ -1,0 +1,288 @@
+#
+# Copyright (c) 2026 Airbyte, Inc., all rights reserved.
+#
+
+import logging
+import threading
+import time
+from dataclasses import dataclass, field
+from datetime import timedelta
+from itertools import cycle
+from typing import Any, List, Mapping, Optional
+
+import requests
+
+from airbyte_cdk.models import FailureType
+from airbyte_cdk.sources.declarative.auth.declarative_authenticator import DeclarativeAuthenticator
+from airbyte_cdk.sources.streams.call_rate import RequestMatcher
+from airbyte_cdk.sources.streams.http import HttpClient
+from airbyte_cdk.sources.streams.http.requests_native_auth import TokenAuthenticator
+from airbyte_cdk.utils import AirbyteTracedException
+from airbyte_cdk.utils.datetime_helpers import AirbyteDateTime, ab_datetime_now, ab_datetime_parse
+
+
+@dataclass
+class TokenQuota:
+    """A named per-token quota pool.
+
+    `matchers` classify outgoing requests into the pool; a pool with no matchers acts as the
+    default pool. `remaining_path`/`reset_path`/`limit_path` locate the pool's values in the
+    quota status response.
+    """
+
+    name: str
+    remaining_path: List[str]
+    reset_path: List[str]
+    limit_path: Optional[List[str]] = None
+    matchers: List[RequestMatcher] = field(default_factory=list)
+
+
+@dataclass
+class _QuotaState:
+    remaining: int
+    reset_at: AirbyteDateTime
+    limit: int
+
+
+class RateLimitedMultipleTokenAuthenticator(DeclarativeAuthenticator):
+    """Authenticator that rotates between multiple interchangeable tokens with per-token quota tracking.
+
+    Each outgoing request is classified into a quota pool using the pool's request matchers.
+    The active token's counter for the matched pool is decremented locally; when it is exhausted
+    the authenticator rotates to the next token. When all tokens are exhausted for a pool, it
+    waits until the earliest quota reset (bounded by `max_wait_time`) and then refreshes all
+    counters from `quota_status_url`, or raises a transient error if the wait would be too long.
+
+    A proactive throttling budget spreads the last calls over the time remaining until reset:
+    once every token's remaining count for a pool drops below its reserve
+    (`max(budget_min_reserve, budget_reserve_fraction * limit)`), a small delay proportional to
+    `seconds_until_reset / total_remaining` (capped at 10s) is injected before each request.
+
+    Counters are seeded per token from `quota_status_url` on first use and refreshed after an
+    exhaustion wait. All state transitions are guarded by a lock so the authenticator can be
+    shared safely across concurrent streams; sleeps never hold the lock.
+    """
+
+    HEARTBEAT_INTERVAL = 60.0  # Log every 60s during exhaustion wait
+    MAX_BUDGET_DELAY = 10.0  # Cap for the per-request proactive throttling delay
+
+    def __init__(
+        self,
+        tokens: List[str],
+        quotas: List[TokenQuota],
+        quota_status_url: str,
+        quota_status_http_method: str = "GET",
+        quota_status_headers: Optional[Mapping[str, str]] = None,
+        auth_method: str = "Bearer",
+        header: str = "Authorization",
+        max_wait_time: timedelta = timedelta(hours=2),
+        budget_reserve_fraction: float = 0.1,
+        budget_min_reserve: int = 50,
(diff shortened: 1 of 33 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4501,6 +4514,83 @@ def create_http_request_matcher(
...
+        self,
+        model: RateLimitedMultipleTokenAuthenticatorModel,
+        config: Config,
+        **kwargs: Any,
+    ) -> RateLimitedMultipleTokenAuthenticator:
+        # Reuse the same instance for identical definitions so that all streams share
+        # the same token quota counters (similar to how api_budget is shared).
+        # `$parameters` are excluded since they are propagated automatically and can
+        # differ per stream even when the authenticator definition is identical.
+        cache_key = model.json(exclude={"parameters"}, sort_keys=True)
```
tolik0: **[High] Instance sharing silently breaks when streams carry `$parameters`.**

`model.json(exclude={"parameters"})` only excludes the *top-level* `parameters` field. `ManifestComponentTransformer.propagate_types_and_parameters` injects each stream's `$parameters` (and copies each key as an extra field) into the nested `quota_status_source` and every `quotas[i]`, and onto the authenticator model itself — all of which end up in the JSON cache key.

Reproduced on this head: a two-stream manifest with identical authenticator blocks but differing stream-level `$parameters` (e.g. `$parameters: {name: ...}` — the norm in hand-written/migrated manifests) produces **2 distinct authenticator instances and 4 seeding calls instead of 2**. With K streams, real token consumption is K× what each instance tracks, so tokens hit real 429s while local counters still show headroom, and rotation/exhaustion/throttling all operate on wrong numbers — a regression vs. today's single shared authenticator in source-github. It also defeats the substream-factory registry sharing added in d92aaac4 (same root cause: parent-context `$parameters` change the key).

Suggested fix: build the cache key from the *resolved constructor arguments* (tokens, quota names/paths/matcher specs, url, method, headers, auth_method, header, max_wait_time, budget params) after interpolation, instead of the raw model JSON — or recursively strip `$parameters`/extra fields first. Please also add a manifest-level regression test with two streams whose `$parameters` differ.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -0,0 +1,301 @@
...
+                    state.remaining -= 1
+                    budget_delay = self._compute_budget_delay(quota)
+                elif all(self._states[token][quota.name].remaining <= 0 for token in self._tokens):
+                    min_time_to_wait = min(
+                        (
+                            self._states[token][quota.name].reset_at - ab_datetime_now()
+                        ).total_seconds()
+                        for token in self._tokens
+                    )
+                    if min_time_to_wait >= self._max_wait_time.total_seconds():
```
tolik0: **[Medium] `max_wait_time` bounds a single wait, not the total — stale reset timestamps cause an infinite reseed loop.**

If the quota endpoint keeps returning `remaining=0` with a reset timestamp at/in the past (clock skew, misconfigured `reset_path`, or an API whose reset field doesn't behave like GitHub's), each cycle computes a small/negative `min_time_to_wait`, floors it to `MIN_EXHAUSTION_WAIT` (5s), sleeps, reseeds (one HTTP call per token), and repeats forever.

Reproduced: with `max_wait_time=timedelta(seconds=30)` and a mock returning `remaining=0, reset=now-100`, the authenticator loops indefinitely and never raises.

The source-github original has the same non-termination (worse — `sleep(0)` busy-loops the quota endpoint), so this is parity-with-a-known-wart, but as a *generic* component with user-configured reset semantics it deserves a cumulative bound: track total time waited per exhaustion episode and raise the existing `transient_error` once it exceeds `max_wait_time`.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -554,6 +554,120 @@ class HttpMethod(Enum):
...
+    )
+    quotas: List[TokenQuota] = Field(
+        ...,
+        description="Quota pools tracked per token. Each outgoing request is classified into the first pool whose matchers match the request; a pool with no matchers acts as the default. The `remaining_path` and `reset_path` locate each pool's values in the quota status response.\n",
+        title="Quota Pools",
+    )
+    max_wait_time: Optional[str] = Field(
+        "PT2H",
+        description="ISO 8601 duration. When all tokens are exhausted, the maximum time to wait for a quota reset before raising a transient error.",
+        examples=["PT2H", "PT30M"],
```
tolik0: **[Medium] Generated file is out of sync with the schema.**

The schema's `max_wait_time` examples include a third entry — `"PT{{ config.get('max_waiting_time', 120) }}M"` (declarative_component_schema.yaml:397, added with the interpolation commit) — that is missing here, so this file was hand-edited rather than regenerated after the schema change.

I replicated the codegen exactly (`datamodel-code-generator==0.26.3`, same flags and post-processing as `bin/generate_component_manifest_files.py`): besides this content mismatch, the new classes are also not where codegen places them, so the next `poe build` anyone runs will produce churn attributed to their PR. Please regenerate instead of hand-applying.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -0,0 +1,301 @@
...
+        response_body = response.json()
+
+        states = {}
+        for quota in self._quotas:
+            remaining = self._extract_path(response_body, quota.remaining_path)
+            reset = self._extract_path(response_body, quota.reset_path)
+            limit = (
+                self._extract_path(response_body, quota.limit_path)
+                if quota.limit_path
+                else remaining
```
tolik0, line 302: **[Low] The `limit := remaining_at_seed` fallback changes throttle onset vs source-github.**

The original computed the reserve from `max(5000, remaining)`, so GitHub always started throttling below ~500 remaining. Without `limit_path`, this fallback makes the reserve `max(50, 0.1 × remaining_at_seed)` — a sync starting when the pool is already low (e.g. remaining=300) gets a reserve of 50 instead of 500 and throttles much later.

The PR-description example manifest sets `limit_path` for `rest` but **not** for `graphql`, so the graphql pool would hit exactly this divergence after migration. Suggest fixing the example and documenting that migrations should set `limit_path` on every pool to preserve behavior.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4501,6 +4514,83 @@ def create_http_request_matcher(
...
+
+        if isinstance(model.tokens, str):
+            tokens_value = InterpolatedString.create(model.tokens, parameters={}).eval(config)
+            delimiter = model.token_delimiter or ","
+            tokens = [
+                token.strip() for token in str(tokens_value).split(delimiter) if token.strip()
+            ]
+        else:
+            tokens = [
+                str(InterpolatedString.create(token, parameters={}).eval(config))
```
tolik0: **[Low] List-path tokens aren't filtered for empty interpolation results.**

The string path strips and filters empty tokens after splitting, but this branch doesn't. A missing config key interpolates to `""`, producing an empty token that gets seeded and signs requests as `"Bearer "` → confusing 401s at runtime instead of a clear config error. Suggest filtering falsy results here and raising the same config error as the empty-tokens case when the list ends up empty.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -0,0 +1,301 @@
...
+                    return quota
+            elif default_quota is None:
+                default_quota = quota
+        if default_quota is None:
+            self._logger.debug(
+                "Request %s did not match any quota pool; falling back to '%s'. Consider defining a matcher-less default pool.",
+                request.url,
+                self._quotas[0].name,
+            )
+        return default_quota or self._quotas[0]
```
tolik0, line 168: **[Low] Silent fallback to `quotas[0]` can misattribute quota.**

When no matcher-less default pool exists and a request matches no matcher, the call is charged to `quotas[0]` with only a debug-level log — if `quotas[0]` is itself a matcher pool, unmatched requests silently drain the wrong pool. Consider validating at construction time that a default (matcher-less) pool exists when any pool has matchers, or at least log a warning once instead of debug.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -0,0 +1,301 @@
...
+        budget_min_reserve: int = 50,
+    ) -> None:
+        if not tokens:
+            raise AirbyteTracedException(
+                failure_type=FailureType.config_error,
+                internal_message="RateLimitedMultipleTokenAuthenticator requires at least one token",
+                message="Authentication tokens are missing from the configuration.",
+            )
+        if not quotas:
+            raise ValueError(
```
tolik0: **[Low] Inconsistent error type for an equivalent config mistake** — empty `tokens` raises `AirbyteTracedException(config_error)` a few lines up, but empty `quotas` raises a plain `ValueError`. Suggest a config_error here too so operators get the same UX.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4501,6 +4514,83 @@ def create_http_request_matcher(
...
+            quota_status_url=quota_status_url,
+            quota_status_http_method=(
+                model.quota_status_source.http_method.value
+                if model.quota_status_source.http_method
+                else "GET"
+            ),
+            quota_status_headers=quota_status_headers,
+            auth_method=model.auth_method or "Bearer",
+            header=model.header or "Authorization",
+            max_wait_time=parse_duration(
```
tolik0: **[Low] Calendar-unit durations crash at exhaustion time, not at construction.**

`isodate.parse_duration("P1M")` returns an `isodate.Duration`, which has no `total_seconds()` — `_acquire_call` would raise `AttributeError` mid-sync at the first exhaustion. Suggest validating here that the parsed value is a `datetime.timedelta` and raising a config error otherwise.

airbyte_cdk/sources/declarative/spec/spec.py
```diff
@@ -54,15 +55,16 @@ def generate_spec(self) -> ConnectorSpecification:
         if self.documentation_url:
             obj["documentationUrl"] = self.documentation_url
         if self.advanced_auth:
-            self.advanced_auth.auth_flow_type = self.advanced_auth.auth_flow_type.value  # type: ignore # We know this is always assigned to an AuthFlow which has the auth_flow_type field
+            if isinstance(self.advanced_auth.auth_flow_type, Enum):
```
tolik0: **[Note] Unrelated change bundled without a test.**

This idempotency guard is behavior-safe (previously a second `generate_spec()` call would `AttributeError` on already-converted enum values), but it's unrelated to the authenticator and ships untested — and it papers over the underlying issue that `generate_spec()` mutates shared model state. Consider splitting into its own PR, or at minimum adding a test that calls `generate_spec()` twice.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4132,6 +4142,9 @@ def create_parent_stream_config_with_substream_wrapper(
             ),
             api_budget=self._api_budget,
         )
+        # Share the authenticator registry so parent and child streams draw from the
+        # same token quota counters
+        substream_factory._rate_limited_authenticators = self._rate_limited_authenticators
```
tolik0: Verified this works: with a parent/child manifest (`SubstreamPartitionRouter`), the parent stream, child stream, and the router's embedded parent stream all share one authenticator instance (2 seeding calls for 2 tokens). Two notes: (1) assigning a private attribute post-construction is fragile — passing the registry through the constructor would be cleaner; (2) this sharing is still defeated by the `$parameters` cache-key issue flagged below in `create_rate_limited_multiple_token_authenticator`, since parent-context `$parameters` change the key.

tolik0 (commented):
Deep review with local reproductions (ran the PR's tests, drove it end-to-end through `ConcurrentDeclarativeSource` manifests, replicated the model codegen, and compared behavior against source-github's `MultipleTokenAuthenticatorWithRateLimiter`). The core machinery works and thread safety is done correctly, but two must-fix items remain — see inline comments:

1. **High**: the shared-instance cache key is poisoned by `$parameters` propagation, so streams get separate counter sets in realistic manifests (reproduced on this head).
2. **Medium**: the checked-in generated models file is out of sync with the schema (hand-edited, not regenerated).

Also a medium on `max_wait_time` not being a cumulative bound (reproduced infinite reseed loop), plus several low-severity robustness notes. The substream-factory registry sharing added in d92aaac4 was verified working with a parent/child manifest.

tolik0 (approved):

PR #1071, opened 2026-07-10
feat(low-code): add UnionPartitionRouter component

darynaishchenko:
## Summary

Adds a new first-class declarative component, `UnionPartitionRouter`, to unblock the source-github migration to manifest-only (see https://github.com/airbytehq/airbyte-internal-issues/issues/16492, https://github.com/airbytehq/airbyte-internal-issues/issues/16512 and airbytehq/airbyte#81428). It yields the deduplicated union of its child partition routers' slices, normalized to a single stable partition field:

```yaml
partition_router:
  type: UnionPartitionRouter
  partition_field: repository
  partition_routers:
    - type: SubstreamPartitionRouter   # repos from parent stream(s)
      parent_stream_configs: [...]
    - type: ListPartitionRouter        # repos from config
      cursor_field: repository
      values: "{{ config['repositories'] }}"
```

Behavior of `UnionPartitionRouter.stream_slices()`:
- Iterates children in order; first occurrence of a `partition_field` value wins, later duplicates are skipped.
- Every emitted slice's partition is exactly `{partition_field: value}`; any other partition keys from a child (e.g. `SubstreamPartitionRouter`'s `parent_slice`) are moved into `extra_fields`, so per-partition state keys stay stable regardless of which child produced the slice.
- `get_stream_state()` merges children's parent-stream states; request option methods return `{}` (children declaring `request_option` are rejected at factory time — consume partition values via `stream_partition` interpolation instead). Note: `CustomPartitionRouter` children are opaque, so the request-option rejection cannot cover them.

Wiring beyond the runtime class and schema:
- `create_union_partition_router` in `ModelToComponentFactory`, registered in `PYDANTIC_MODEL_TO_CONSTRUCTOR`; schema registered in `SimpleRetriever`/`AsyncRetriever` `partition_router` anyOf, with self-nesting allowed (`minItems: 2`).
- `LegacyToPerPartitionStateMigration` accepts `UnionPartitionRouter`: `_get_partition_field` reads `partition_field` directly, and the single-parent-stream check in `should_migrate` doesn't apply.
- Parent-stream `use_cache` auto-enable (`_initialize_cache_for_parent_streams`) and parent/ancestor dependency-graph collection (`_collect_all_parent_stream_names`, `_collect_all_ancestor_names`) now recurse into composed routers (`partition_routers` / `underlying_partition_router`).

Tests:
- Runtime unit tests (dedup, normalization, first-wins, nesting, state merge), factory tests (construction + request_option rejection), and legacy→per-partition state migration tests with a union router.
- Concurrency-graph/caching coverage: `use_cache` force-enabled on all parents referenced through a union (including nested parent stream instances), and parent/ancestor collection detects deadlocks through `UnionPartitionRouter` children, incl. a union nested inside `GroupingPartitionRouter`.
- Runtime round-trip: a read resuming from legacy state through a union-partitioned incremental stream asserts the migrated per-partition keys exactly match the runtime per-partition state keys emitted during the read (and that duplicated partitions across parents are read only once).

Note on generated models: `poetry run poe assemble` produces the same ~350-line style drift (`conint`/`confloat`, class reordering) on a clean `main` checkout in this environment, so `declarative_component_schema.py` was hand-patched with only the UnionPartitionRouter additions to preserve the committed generated style.

Link to Devin session: https://app.devin.ai/sessions/38753ed2d1f346e6a370888444676f4f

## Summary by CodeRabbit

- **New Features**
  - Added support for combining partitions from multiple routers with the new Union Partition Router.
  - Supports nested unions, grouping, list, substream, and custom routers.
  - Deduplicates partitions while preserving order and associated fields.
  - Enables legacy state migration and concurrent processing for union-based configurations.
  - Made the new router publicly available through the package API.

- **Bug Fixes**
  - Improved parent-stream discovery and state handling for nested partition-router configurations.

- **Tests**
  - Added comprehensive coverage for creation, validation, deduplication, nesting, state migration, and concurrent processing.

### DIFF
--- a/airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
+++ b/airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
@@ -19,6 +19,9 @@ from airbyte_cdk.sources.declarative.partition_routers.grouping_partition_router
 from airbyte_cdk.sources.declarative.partition_routers.substream_partition_router import (
     SubstreamPartitionRouter,
 )
+from airbyte_cdk.sources.declarative.partition_routers.union_partition_router import (
+    UnionPartitionRouter,
+)
 from airbyte_cdk.sources.message import MessageRepository
 from airbyte_cdk.sources.streams.concurrent.abstract_stream import AbstractStream
 from airbyte_cdk.sources.streams.concurrent.default_stream import DefaultStream
@@ -438,14 +441,18 @@ class ConcurrentReadProcessor:
         partition_router = (
             stream.get_partition_router() if isinstance(stream, DefaultStream) else None
         )
-        if isinstance(partition_router, GroupingPartitionRouter):
-            partition_router = partition_router.underlying_partition_router
-
-        if isinstance(partition_router, SubstreamPartitionRouter):
-            for parent_config in partition_router.parent_stream_configs:
-                parent_name = parent_config.stream.name
-                parent_names.add(parent_name)
-                parent_names.update(self._collect_all_parent_stream_names(parent_name))
+        routers = [partition_router] if partition_router else []
+        while routers:
+            router = routers.pop()
+            if isinstance(router, GroupingPartitionRouter):
+                routers.append(router.underlying_partition_router)
+            elif isinstance(router, UnionPartitionRouter):
+                routers.extend(router.partition_routers)
+            elif isinstance(router, SubstreamPartitionRouter):
+                for parent_config in router.parent_stream_configs:
+                    parent_name = parent_config.stream.name
+                    parent_names.add(parent_name)
+                    parent_names.update(self._collect_all_parent_stream_names(parent_name))
 
         return parent_names
 
--- a/airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
+++ b/airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
@@ -13,6 +13,9 @@ from airbyte_cdk.sources.concurrent_source.partition_generation_completed_sentin
 )
 from airbyte_cdk.sources.concurrent_source.stream_thread_exception import StreamThreadException
 from airbyte_cdk.sources.concurrent_source.thread_pool_manager import ThreadPoolManager
+from airbyte_cdk.sources.declarative.partition_routers.cartesian_product_stream_slicer import (
+    CartesianProductStreamSlicer,
+)
 from airbyte_cdk.sources.declarative.partition_routers.grouping_partition_router import (
     GroupingPartitionRouter,
 )
@@ -441,13 +444,15 @@ class ConcurrentReadProcessor:
         partition_router = (
             stream.get_partition_router() if isinstance(stream, DefaultStream) else None
         )
-        routers = [partition_router] if partition_router else []
+        routers = [partition_router] if partition_router is not None else []
         while routers:
             router = routers.pop()
             if isinstance(router, GroupingPartitionRouter):
                 routers.append(router.underlying_partition_router)
             elif isinstance(router, UnionPartitionRouter):
                 routers.extend(router.partition_routers)
+            elif isinstance(router, CartesianProductStreamSlicer):
+                routers.extend(router.stream_slicers)
             elif isinstance(router, SubstreamPartitionRouter):
                 for parent_config in router.parent_stream_configs:
                     parent_name = parent_config.stream.name
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -82,6 +82,9 @@ from airbyte_cdk.sources.declarative.partition_routers.grouping_partition_router
 from airbyte_cdk.sources.declarative.partition_routers.substream_partition_router import (
     SubstreamPartitionRouter,
 )
+from airbyte_cdk.sources.declarative.partition_routers.union_partition_router import (
+    UnionPartitionRouter,
+)
 from airbyte_cdk.sources.declarative.resolvers import COMPONENTS_RESOLVER_TYPE_MAPPING
 from airbyte_cdk.sources.declarative.spec.spec import Spec
 from airbyte_cdk.sources.declarative.types import Config, ConnectionDefinition
@@ -461,15 +464,18 @@ class ConcurrentDeclarativeSource(Source):
             inst = stream_name_to_instance.get(stream_name)
             if not isinstance(inst, DefaultStream):
                 return ancestors
-            router = inst.get_partition_router()
-            if isinstance(router, GroupingPartitionRouter):
-                router = router.underlying_partition_router
-            if not isinstance(router, SubstreamPartitionRouter):
-                return ancestors
-            for parent_config in router.parent_stream_configs:
-                parent_name = parent_config.stream.name
-                ancestors.add(parent_name)
-                ancestors.update(_collect_all_ancestor_names(parent_name))
+            routers = [inst.get_partition_router()]
+            while routers:
+                router = routers.pop()
+                if isinstance(router, GroupingPartitionRouter):
+                    routers.append(router.underlying_partition_router)
+                elif isinstance(router, UnionPartitionRouter):
+                    routers.extend(router.partition_routers)
+                elif isinstance(router, SubstreamPartitionRouter):
+                    for parent_config in router.parent_stream_configs:
+                        parent_name = parent_config.stream.name
+                        ancestors.add(parent_name)
+                        ancestors.update(_collect_all_ancestor_names(parent_name))
             return ancestors
 
         for stream in streams:
@@ -532,14 +538,23 @@ class ConcurrentDeclarativeSource(Source):
             elif stream_config.get("retriever", {}).get("partition_router", {}):
                 partition_router = stream_config["retriever"]["partition_router"]
 
-                if isinstance(partition_router, dict) and partition_router.get(
-                    "parent_stream_configs"
-                ):
-                    update_with_cache_parent_configs(partition_router["parent_stream_configs"])
-                elif isinstance(partition_router, list):
-                    for router in partition_router:
-                        if router.get("parent_stream_configs"):
-                            update_with_cache_parent_configs(router["parent_stream_configs"])
+                routers = (
+                    list(partition_router)
+                    if isinstance(partition_router, list)
+                    else [partition_router]
+                )
+                while routers:
+                    router = routers.pop()
+                    if not isinstance(router, dict):
+                        continue
+                    if router.get("parent_stream_configs"):
+                        update_with_cache_parent_configs(router["parent_stream_configs"])
+                    # Descend into composed partition routers (e.g. UnionPartitionRouter's
+                    # partition_routers or GroupingPartitionRouter's underlying_partition_router)
+                    # so nested parent streams also get caching enabled.
+                    routers.extend(router.get("partition_routers") or [])
+                    if router.get("underlying_partition_router"):
+                        routers.append(router["underlying_partition_router"])
 
         for stream_config in stream_configs:
             if stream_config["name"] in parent_streams:
--- a/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
+++ b/airbyte_cdk/sources/declarative/concurrent_declarative_source.py
@@ -76,6 +76,9 @@ from airbyte_cdk.sources.declarative.parsers.manifest_reference_resolver import 
 from airbyte_cdk.sources.declarative.parsers.model_to_component_factory import (
     ModelToComponentFactory,
 )
+from airbyte_cdk.sources.declarative.partition_routers.cartesian_product_stream_slicer import (
+    CartesianProductStreamSlicer,
+)
 from airbyte_cdk.sources.declarative.partition_routers.grouping_partition_router import (
     GroupingPartitionRouter,
 )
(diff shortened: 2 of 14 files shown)


airbyte_cdk/sources/declarative/partition_routers/union_partition_router.py
```diff
@@ -0,0 +1,136 @@
...
+
+from airbyte_cdk.sources.declarative.partition_routers.partition_router import PartitionRouter
+from airbyte_cdk.sources.types import StreamSlice, StreamState
+
+logger = logging.getLogger("airbyte")
+
+
+@dataclass
+class UnionPartitionRouter(PartitionRouter):
+    """
```
tolik0, line 17: This docstring is the generated-docs surface (`docs/generate.py`), and it's silent on the two things a manifest author most needs to know before reaching for this component:

- **Memory**: dedup requires retaining every distinct partition value for the whole of slice generation. Memory grows linearly with unique partition count.
- **Hashability**: `stream_slices` raises `ValueError` on unhashable partition values (L61-65), but nothing here or in the schema says so.

Suggested addition to the `Attributes:` block or just above it:

```
Partition values must be hashable — deduplication holds every distinct value in a set
for the duration of slice generation, so memory grows linearly with the number of
unique partitions. Note also that a union's partition count is the (deduplicated) sum
of its children's, which counts against ConcurrentPerPartitionCursor's per-partition
state limit.
```

airbyte_cdk/sources/declarative/partition_routers/union_partition_router.py
```diff
@@ -0,0 +1,136 @@
...
+    def __post_init__(self, parameters: Mapping[str, Any]) -> None:
+        self._parameters = parameters
+
+    def stream_slices(self) -> Iterable[StreamSlice]:
+        """
+        Iterate over the child partition routers in order, yielding each partition value once.
+
+        The first occurrence of a partition value wins; later duplicates from any child are skipped.
+        """
+        seen: set[Any] = set()
```
tolik0, line 56: Worth an explicit comment here, because this set is never cleared and — since the concurrent framework consumes `stream_slices()` lazily via the partition enqueuer — it stays live for essentially the whole sync, not just a slice-generation phase.

For source-github this is fine (a few thousand repo-name strings, order of 1 MB), so I'm not asking for a change in behaviour. But the more consequential downstream effect is that a union's partition count is the sum of its children's: `ConcurrentPerPartitionCursor.DEFAULT_MAX_PARTITIONS_NUMBER = 25_000` (`concurrent_partition_cursor.py:123`), past which per-partition cursors are evicted oldest-first with a warning and per-partition state silently degrades to global. Every partition is also an entry in the emitted `states: [...]` message.

A one-line comment here plus the docstring/schema notes would be enough.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -4475,6 +4479,47 @@ definitions:
       $parameters:
         type: object
         additionalProperties: true
+  UnionPartitionRouter:
+    title: Union Partition Router
+    description: >
+      A partition router that yields the deduplicated union of the partitions produced by its child
```
tolik0, line 4667: This `description` is what renders in the Connector Builder, so it's the surface most users will actually read. It covers dedup semantics well but has no scaling caveat.

`GroupingPartitionRouter` already sets the precedent for this at L4445-4446 ("Note that per-partition incremental syncs may not work as expected because..."). Suggest an analogous sentence:

```yaml
      Partition values must be hashable; deduplication holds every distinct value in memory
      for the duration of the sync, and the resulting partition count is the deduplicated sum
      of the child routers' partition counts, which affects per-partition state size.
```

unit_tests/sources/declarative/test_concurrent_declarative_source.py
```diff
@@ -5481,6 +5660,236 @@ def test_apply_stream_groups_raises_on_parent_child_in_same_group_with_grouping_
...
+        parameters={},
+    )
+
+    stream_slicer_router = (
+        GroupingPartitionRouter(
+            group_size=10,
+            underlying_partition_router=union_router,
+            config={},
+        )
+        if wrap_in_grouping
```
tolik0: This `wrap_in_grouping` path — and the corresponding `GroupingPartitionRouter → UnionPartitionRouter` branches at `concurrent_read_processor.py:447` and `concurrent_declarative_source.py:470` — isn't reachable from a manifest.

`declarative_component_schema.yaml:4467` (`GroupingPartitionRouter.underlying_partition_router`) still has `anyOf: [ListPartitionRouter, SubstreamPartitionRouter, CustomPartitionRouter]` and isn't touched by this PR; the generated `GroupingPartitionRouter` model is likewise unchanged. A manifest nesting a union inside a grouping router fails validation before it ever reaches this code. The test passes only because it constructs `GroupingPartitionRouter(...)` in Python directly, bypassing the schema.

Either add `UnionPartitionRouter` to that `anyOf` (and decide whether `GroupingPartitionRouter` should be permitted inside `UnionPartitionRouter.partition_routers` too — currently it isn't), or drop the runtime branch and this test parameter. Right now it reads as covered when it isn't.

airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
```diff
@@ -438,14 +441,18 @@ def _collect_all_parent_stream_names(self, stream_name: str) -> Set[str]:
...
         )
-        if isinstance(partition_router, GroupingPartitionRouter):
-            partition_router = partition_router.underlying_partition_router
-
-        if isinstance(partition_router, SubstreamPartitionRouter):
-            for parent_config in partition_router.parent_stream_configs:
-                parent_name = parent_config.stream.name
-                parent_names.add(parent_name)
-                parent_names.update(self._collect_all_parent_stream_names(parent_name))
+        routers = [partition_router] if partition_router else []
```
tolik0: Good change — the stack-based traversal is clearer than the previous chained `isinstance`. Two notes:

- This loop (and its twin at `concurrent_declarative_source.py:467`) doesn't descend into `CartesianProductStreamSlicer`, which is what `model_to_component_factory.py:2242` builds when a manifest declares `partition_router` as a **list**. Pre-existing for `SubstreamPartitionRouter`, but the union router makes it much more likely to bite: Step 8 of the source-github migration (`Commits`, repository × branch) will plausibly want `[repository_partition_router, branches_router]`, at which point deadlock detection goes blind to the union's parents. Since you're already restructuring this loop, adding `elif isinstance(router, CartesianProductStreamSlicer): routers.extend(router.stream_slicers)` here would close it cheaply.
- Non-blocking: `if partition_router` is truthiness rather than `is not None`. Fine for the current router types, but `is not None` is what's meant.

airbyte_cdk/sources/declarative/concurrent_declarative_source.py
```diff
@@ -461,15 +464,18 @@ def _collect_all_ancestor_names(stream_name: str) -> Set[str]:
...
-            router = inst.get_partition_router()
-            if isinstance(router, GroupingPartitionRouter):
-                router = router.underlying_partition_router
-            if not isinstance(router, SubstreamPartitionRouter):
-                return ancestors
-            for parent_config in router.parent_stream_configs:
-                parent_name = parent_config.stream.name
-                ancestors.add(parent_name)
-                ancestors.update(_collect_all_ancestor_names(parent_name))
+            routers = [inst.get_partition_router()]
```
tolik0: Same `CartesianProductStreamSlicer` gap as `concurrent_read_processor.py:444` — worth fixing in both places together, since they encode the same graph.

Minor: unlike the sibling, this seeds the stack with `[inst.get_partition_router()]` unconditionally, so a `None` can be pushed. Harmless (no `isinstance` branch matches), but it's an inconsistency between two loops that should stay in lockstep. Given they're now structurally identical, extracting a shared `_iter_composed_routers(router)` generator would prevent them drifting apart the next time a composite router is added.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4545,6 +4557,84 @@ def create_grouping_partition_router(
...
+                    InterpolatedString.create(
+                        child_model.cursor_field, parameters=child_model.parameters or {}
+                    ).eval(config)
+                )
+            elif isinstance(child_model, SubstreamPartitionRouterModel):
+                for parent_stream_config in child_model.parent_stream_configs:
+                    child_partition_fields.append(
+                        InterpolatedString.create(
+                            parent_stream_config.partition_field,
+                            parameters=child_model.parameters or {},
```
tolik0: Nit: `parent_stream_config.partition_field` is interpolated with `child_model.parameters` rather than the parent stream config's own `$parameters`. Harmless in practice because `propagate_types_and_parameters` has already pushed parameters down by this point, but it's inconsistent with how the runtime evaluates the same field (`substream_partition_router.py:199`, via the config-scoped `InterpolatedString`). Worth aligning so the build-time check can't diverge from runtime behaviour.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -3246,6 +3250,29 @@ class GroupingPartitionRouter(BaseModel):
...
 
+class UnionPartitionRouter(BaseModel):
+    type: Literal["UnionPartitionRouter"]
+    partition_field: str = Field(
+        ...,
+        description="The single partition key that all child partition routers' slices are normalized to. Each child router must emit this key in its partitions. Interpolation is evaluated once when the connector is built, using the connector config and $parameters.",
+        examples=["repository", "{{ config['partition_field'] }}"],
+        title="Partition Field",
+    )
+    partition_routers: List[
```
tolik0, line 3381: Re the note in the PR description about hand-patching this file to avoid the `poe assemble` style drift — the workaround is reasonable and I appreciate you calling it out explicitly. But it's already producing divergence: the YAML has `minItems: 2` on `partition_routers` (L4513) and this model has no corresponding constraint.

Not a bug today, since manifest validation enforces the JSON schema before models are built. But it's exactly the class of drift hand-patching invites. Could you open a follow-up issue for the generator drift so this file doesn't quietly become hand-maintained?

tolik0 (changes requested):

airbyte_cdk/sources/concurrent_source/concurrent_read_processor.py
```diff
@@ -438,14 +444,20 @@ def _collect_all_parent_stream_names(self, stream_name: str) -> Set[str]:
...
-                parent_names.add(parent_name)
-                parent_names.update(self._collect_all_parent_stream_names(parent_name))
+        routers = [partition_router] if partition_router is not None else []
+        while routers:
+            router = routers.pop()
+            if isinstance(router, GroupingPartitionRouter):
+                routers.append(router.underlying_partition_router)
+            elif isinstance(router, UnionPartitionRouter):
+                routers.extend(router.partition_routers)
+            elif isinstance(router, CartesianProductStreamSlicer):
```
tolik0, line 454: Non-blocking, but the one thing I'd still like: this branch and its twin in `concurrent_declarative_source.py` have no test coverage. Grepping the full PR diff for `CartesianProduct` returns 5 hits, all in these two source files (imports + `isinstance`) and none in tests.

That's two new branches in deadlock detection and parent-stream collection landing untested — and it's exactly the path Step 8 of the source-github migration (`Commits`, repository × branch) will exercise, which was the motivation for adding it. One test would cover both: a union nested inside a list-valued `partition_router`, asserting the union's parents are still collected.

Happy for this to be a follow-up rather than a blocker.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -4466,6 +4470,7 @@ definitions:
         anyOf:
           - "$ref": "#/definitions/ListPartitionRouter"
           - "$ref": "#/definitions/SubstreamPartitionRouter"
+          - "$ref": "#/definitions/UnionPartitionRouter"
```
tolik0, line 4654: Nice — this closes the reachability gap.

One thing worth making explicit: the relationship is now deliberately asymmetric. `GroupingPartitionRouter` may wrap a `UnionPartitionRouter`, but `UnionPartitionRouter.partition_routers` still excludes `GroupingPartitionRouter`.

I checked and that's the **correct** call — `_create_grouped_slice` builds `{key: [v1, v2, ...]}` (`grouping_partition_router.py:92-94`), so a grouped slice's partition value is a list, which would hit the union's own "can only deduplicate hashable partition values" `ValueError` at runtime. But right now that rule is enforced only by omission from this `anyOf`, which reads as an oversight rather than a decision.

A clause in the `UnionPartitionRouter` description — something like "child routers must emit scalar partition values, so `GroupingPartitionRouter` cannot be used as a child" — would make it intentional. Non-blocking.

tolik0 (approved):
Thanks — all eight items are addressed in 2468b394. I re-verified each against the head tree rather than the replies, and also checked the things that could plausibly have broken in the process:

- `ParentStreamConfig.parameters` does exist (`models/declarative_component_schema.py:2960`), so the `$parameters`-scope fix isn't an `AttributeError` waiting to happen. The `or` chain is semantically right too — `propagate_types_and_parameters` merges downward, so a populated parent-config params dict already subsumes the child's.
- `CartesianProductStreamSlicer.stream_slicers` is the correct attribute (`cartesian_product_stream_slicer.py:60`), and the new imports introduce no cycle — both modules already import siblings from that package.
- Grouping-over-union works at runtime, not just in the schema: `create_grouping_partition_router` dispatches the union child through the registered constructor, and `GroupingPartitionRouter.stream_slices` requires exactly one partition key after stripping `parent_slice` — a union slice is exactly `{partition_field: value}`, so it's a clean fit. The grouping router's request-option guards only match `Substream`/`List` underlying routers, but nothing slips through because the union's own factory already rejects request options on its children.

The alignment with airbytehq/airbyte#81428 is unchanged by this commit — neither union router's build path nor the `LegacyToPerPartitionStateMigration` behaviour the connector depends on was touched.

Approving. Two non-blocking notes inline; I'd like the first one but won't hold the PR for it. Declining the `_iter_composed_routers` extraction is fine by me as a follow-up.

**On CI** — not green, but neither failure looks related:
- Pytest 3.12: 1 failed / 4174 passed, and the failure is `unit_tests/legacy/sources/declarative/test_manifest_declarative_source.py::test_given_unmigrated_config_when_migrating_then_config_is_migrated` (`Expected 'open' to be called once. Called 6 times.`) — a legacy config-migration test that mocks builtin `open`. 3.10/3.11/3.13 pass on identical code and this PR touches nothing in that path; reads as ordering-dependent mock leakage.
- `destination-motherduck`: the same check you already re-triggered for an infra flake in ca6cc171.

Worth a re-run to confirm rather than a code change.

darynaishchenko:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/31008353273
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/31008380500

PR #1083, opened 2026-07-28
fix(declarative): allow bundled custom components without AIRBYTE_ENABLE_UNSAFE_CODE

### LINKED ISSUE #1082: Custom components in published manifest-only connectors are rejected in Cloud since 7.23.7
## Summary

CDK 7.23.7 broke every published manifest-only connector that bundles a `components.py`, in any environment that does not set `AIRBYTE_ENABLE_UNSAFE_CODE=true` — including Airbyte Cloud. `spec` still works, but `check`, `discover`, and `read` all fail immediately with:

```
airbyte_cdk.sources.declarative.parsers.custom_code_compiler.AirbyteCustomCodeNotPermittedError:
Custom connector code is not permitted in this environment.
```

## Root cause

https://github.com/airbytehq/airbyte-python-cdk/pull/1076 added an unconditional gate in `ModelToComponentFactory.create_custom_component`:

```python
if not custom_code_execution_permitted():
    raise AirbyteCustomCodeNotPermittedError
```

Before that change, the `AIRBYTE_ENABLE_UNSAFE_CODE` gate applied only to `components.py` code *injected through the config* (`__injected_components_py`), i.e. untrusted, user-supplied code from Connector Builder / a remote manifest. Code baked into a published connector image was trusted.

The new gate does not distinguish provenance, so a manifest-only connector shipping its own `manifest.yaml` + `components.py` inside its own image is now rejected too.

## Impact

Any manifest-only connector whose manifest references a `Custom*` component. In the monorepo, 56 connectors bundle a `components.py`; 4 are already published on base image `source-declarative-manifest:7.23.7` and are broken in Cloud today:

- `source-chargebee` (0.10.41)
- `source-bing-ads`
- `source-greenhouse`
- `source-apify-dataset`

27 more are on 7.23.6 and will break on their next automated dependency bump.

Example failure (`source-chargebee` 0.10.41, whose manifest uses `CustomTransformation` -> `source_declarative_manifest.components.CustomFieldTransformation` on ~27 streams):

```
File ".../parsers/model_to_component_factory.py", line 2108, in create_default_stream
    self._create_component_from_model(model=transformation_model, config=config)
File ".../parsers/model_to_component_factory.py", line 1822, in create_custom_component
    raise AirbyteCustomCodeNotPermittedError
```

Downstream destinations then fail with `Input was fully read, but some streams did not receive a terminal stream status message`, which obscures the real cause.

## Proposed fix

Restore the provenance distinction: apply the gate only when the manifest itself came from the config (`__injected_declarative_manifest` present), which is the untrusted case that PR #1076 set out to protect. Image-bundled manifests keep working without `AIRBYTE_ENABLE_UNSAFE_CODE`.

Affected connectors additionally need a rebuild onto the fixed CDK release once it ships.

---
Filed by Devin at the request of Syed Khadeer (syed.khadeer@airbyte.io) while triaging a broken production Chargebee -> BigQuery connection.

[Devin session](https://app.devin.ai/sessions/fbef868f12844f26b7cb6edbc1b3a8b0)

Airbyte-Support:
## Summary

Fixes https://github.com/airbytehq/airbyte-python-cdk/issues/1082: since 7.23.7, every published manifest-only connector that bundles a `components.py` fails `check`/`discover`/`read` in Airbyte Cloud with `AirbyteCustomCodeNotPermittedError`. Found while triaging a production Chargebee -> BigQuery connection that broke the moment `source-chargebee` 0.10.41 (base image `source-declarative-manifest:7.23.7`) rolled out.

#1076 added the `AIRBYTE_ENABLE_UNSAFE_CODE` gate to `create_custom_component` unconditionally, which lost the provenance distinction the gate has always had: it is meant to protect against *untrusted* code arriving from a caller (`__injected_components_py` / a caller-supplied manifest from Connector Builder), not against a connector's own code baked into its own image. Cloud never sets `AIRBYTE_ENABLE_UNSAFE_CODE`, so trusted first-party connectors started failing.

The gate now keys off manifest provenance instead of the environment alone:

```diff
-if not custom_code_execution_permitted():
+manifest_is_untrusted = not self._custom_components_trusted or bool(config.get(INJECTED_MANIFEST))
+if manifest_is_untrusted and not custom_code_execution_permitted():
     raise AirbyteCustomCodeNotPermittedError
```

Two provenance signals, because untrusted manifests reach the factory two different ways:

- `__injected_declarative_manifest` in the config.
- `custom_components_trusted=False`, a new kwarg threaded `ConcurrentDeclarativeSource` -> `ModelToComponentFactory` (and on to `substream_factory`).

A manifest read from `/airbyte/integration_code/source_declarative_manifest/manifest.yaml` inside a published image hits neither signal and may use its bundled components, as before 7.23.7. #1076's threat model (a manifest pointing `class_name` at any importable callable, e.g. `os.getcwd`) stays blocked everywhere it applied.

## Security follow-up: `custom_components_trusted=False` on *every* caller-supplied entry point

Review of the first cut found that the `config.get(INJECTED_MANIFEST)` signal alone is **bypassable**, so relying on it at the Connector Builder / `source-declarative-manifest` remote entry points reopened the #1076 RCE (GHSA-g35w-6r6v-h4r7). Two ways, both verified:

1. **Spec-level components dodge the config check.** `ConcurrentDeclarativeSource.__init__` builds the spec component with an empty dict config (`create_component(SpecModel, spec, dict())`), so a `CustomConfigTransformation`/`CustomValidationStrategy` in `spec.config_normalization_rules` reaches `create_custom_component` with no `INJECTED_MANIFEST` key.
2. **A manifest can delete its own provenance key.** A plain `ConfigRemoveFields` on `["__injected_declarative_manifest"]` runs during config normalization before streams are built, so later `Custom*` components see a config with the marker gone.

Both routes are the exact attack surface #1076 closed. Because `manifest_server.build_source` was the only caller passing the trust flag, the other two untrusted entry points were exposed. Fix: pass `custom_components_trusted=False` in both:

- `connector_builder/connector_builder_handler.py::create_source`
- `cli/source_declarative_manifest/_run.py::create_declarative_source` (remote-manifest path only; the bundled published connector uses the local-manifest path, which stays trusted)

The trust flag lives on the factory, not in the config, so it is immune to both bypasses. `config.get(INJECTED_MANIFEST)` is now redundant defense-in-depth rather than the primary control.

## Blast radius of the regression

56 monorepo connectors bundle a `components.py`. Per the live registry, **19 are published on 7.23.7 and broken in Cloud today** — apify-dataset, bing-ads, chargebee, google-search-console, greenhouse, instagram, klaviyo, linkedin-ads, monday, nexus-datasets, notion, outreach, paypal-transaction, pinterest, pylon, qualaroo, recurly, retently, slack (full table in https://github.com/airbytehq/airbyte-python-cdk/issues/1082#issuecomment-5101948949) — with 13 more on 7.23.6 that would break at their next base-image bump. All 19 need a rebuild once this ships.

## Tests

- #1076's gate test now injects a manifest into the config and is parametrized over a bundled component class and `os.getcwd`, asserting the gate fires *before* `class_name` is resolved; `test_create_custom_component_permitted_for_bundled_manifest` covers the image-bundled case with the env var unset; `test_create_custom_component_requires_custom_code_enabled_when_untrusted` covers the manifest-server flag. The two autouse `conftest.py` fixtures #1076 added to force the env var on across the declarative test suites are removed — they masked exactly this regression.
- New regression tests for the two bypasses, driven through the real Connector Builder `create_source` entry point with `AIRBYTE_ENABLE_UNSAFE_CODE` unset: `test_create_source_marks_manifest_untrusted` (invariant), `test_spec_level_custom_component_is_gated`, `test_config_strip_does_not_bypass_custom_code_gate`. All three fail on the pre-follow-up branch and pass with the fix.
- Verified beyond unit tests by recreating the published `source-declarative-manifest` image layout with source-chargebee 0.10.41's real `manifest.yaml` + `components.py` and running the connector CLI with `AIRBYTE_ENABLE_UNSAFE_CODE` unset: `discover` fails on `main` and returns 27 streams here, `CustomFieldTransformation` is genuinely instantiated, and a poisoned `class_name` is still blocked on every caller-supplied route.
- Also verified end-to-end with source-notion 4.0.18 (base image `source-declarative-manifest:7.23.7`) using real integration-test credentials, bundled `manifest.yaml` + `components.py`, and `AIRBYTE_ENABLE_UNSAFE_CODE` unset: `check` fails with `AirbyteCustomCodeNotPermittedError` on released CDK 7.23.7 and succeeds on this branch.

## `--components-path` gating

The remaining pre-existing gap flagged in review: `_register_components_from_file` executed a caller-supplied `--components-path` file via `exec_module` *before* any gate could fire, so the trust flag offered no protection on that CLI route. It is now gated: the helper checks `custom_code_execution_permitted()` and raises `AirbyteCustomCodeNotPermittedError` before the module is created or executed. Local-operator workflows using `--components-path` must set `AIRBYTE_ENABLE_UNSAFE_CODE=true` (same requirement as before #1076's regression for injected code). Covered by `test_register_components_from_file_is_gated`.

## Summary by CodeRabbit

- **Security**
  - Improved protection for custom components referenced by declarative manifests supplied at runtime.
  - Custom-component code from untrusted manifests is now blocked unless custom-code execution is explicitly enabled.
  - Bundled connector manifests still allow trusted custom components without requiring the unsafe-code setting.
- **Bug Fixes**
  - Ensured custom-component instantiation is gated before any class resolution/loading occurs for untrusted manifests.

Link to Devin session: https://app.devin.ai/sessions/20f99bb504b543fda97972948b0ceecb


darynaishchenko:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/30345566070
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/30345587161

darynaishchenko:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/30356998272
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/30357014888

pnilan (approved):

PR #1085, opened 2026-07-28
ci: test connector images against an SDM base image built from the current branch

darynaishchenko:
## Summary

Closes the third CI gap that let https://github.com/airbytehq/airbyte-python-cdk/pull/1076 ship the https://github.com/airbytehq/airbyte-python-cdk/issues/1082 regression without any check going red: the Docker image tests always built connector images `FROM` the *already-published* SDM base image from `metadata.yaml`, so the PR branch's CDK was never inside the container. A second gap compounded it — a failing in-container `check` exits 0 and only reports `CONNECTION_STATUS: FAILED`, which the harness never asserted on.

Three changes:

1. **Base-image override, plumbed end to end** — build a connector image `FROM` an arbitrary (e.g. locally built, unpublished) SDM image:
   ```text
   build_connector_image(..., base_image_override=...)   # utils/docker.py
   airbyte-cdk image build/test --base-image <image>     # cli/_image.py
   pytest --connector-base-image <image>                 # pytest_hooks.py -> connector_base_image_override fixture -> docker_base.py
   ```
   `base_image_override or metadata.data.connectorBuildOptions.baseImage` — metadata remains the default when the flag is absent.

2. **The existing "Test Connectors" matrix now tests the branch's CDK inside the container.** For manifest-only connectors (source-intercom, source-pokeapi), the matrix job builds `airbyte/source-declarative-manifest:dev` from the current branch and runs `image test --base-image` with it — previously those container tests ran the already-released CDK from the published base image, so they couldn't catch a CDK regression pre-release. Python connectors already get the branch CDK in their images: the `Bump to Dev Branch CDK` step pins the connector's CDK dependency to the branch ref, which the container image's `poetry install` then picks up.

3. **`test_docker_image_build_and_check` now asserts the check outcome** — previously a `check` returning `CONNECTION_STATUS: FAILED` (exit code 0, no trace error) *passed* the test. Now the reported outcome must match the scenario in both directions: scenarios expecting success must report `Status.SUCCEEDED`, scenarios expecting failure must *not* report `SUCCEEDED` (previously those were skipped entirely, so an expected-to-fail `check` that started succeeding in-container went unnoticed), and scenarios with no declared `status` map to `ALLOW_ANY` and keep their either-outcome-passes contract. The assertion logic lives in an `_assert_check_outcome` helper with unit tests covering the full expected-outcome × reported-status matrix.

Verified locally with the exact CI command:
- base image = published SDM `7.23.7` (the #1076 CDK): `test_docker_image_build_and_check` **fails** — check reports `Custom connector code is not permitted in this environment`.
- base image = SDM built from a branch carrying the #1083 fix: all image tests **pass**.

Also adds a parametrized unit test asserting `base_image_override` is threaded into the `BASE_IMAGE` build arg (and that metadata remains the default). The test runs against a `tmp_path` copy of the pokeapi fixture with Dockerfile-template resolution stubbed, so it is hermetic: no network fallback and no writes into the checked-in fixture directory.

## FAST standard tests also use the branch SDM base

The `Run FAST Standard Tests [Manifest-Only Connectors]` step also collects the docker image tests (`test_docker_image_build_and_{spec,check,read}`), which previously built the connector image `FROM` the published metadata base image — duplicating each image test against the released CDK and preventing layer reuse with the later `image test` step. The SDM build step now runs before it, and the FAST step passes `--pytest-arg=--connector-base-image --pytest-arg=airbyte/source-declarative-manifest:dev`, so both steps build `FROM` the branch-built SDM image. This is also the path that exercises the new pytest `--connector-base-image` option and `connector_base_image_override` fixture.

Also fixes a pre-existing crash in `airbyte-cdk connector test`: `--pytest-arg` (click `multiple=True`) yields a tuple, so combining it with `--collect-only` or `--no-creds` raised `AttributeError: 'tuple' object has no attribute 'append'`.

## CI state

While https://github.com/airbytehq/airbyte-python-cdk/pull/1083 was unmerged, the matrix container test for `source-intercom` (bundled `components.py`) correctly failed on this PR with `Custom connector code is not permitted in this environment` — proving the guard catches the #1082 regression when the branch-built SDM carries it. Now that #1083 is merged and this branch is updated with `main`, that check passes.

Note: the new `check`-outcome assertion also surfaced that `destination-motherduck`'s in-container `check` has been silently failing in CI (MotherDuck token auth error + a `PermissionError`), previously masked because the harness ignored `CONNECTION_STATUS: FAILED`. That looks like a connector-secrets/environment issue, independent of this PR's code.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `--base-image` override for connector image **build** and **test** flows.
  * CI now treats *manifest-only* connectors specially by building and using a local development base image for their connector tests and image checks.
* **Tests**
  * Extended connector Docker checks to assert successful connection status progression (not just container exit codes).
  * Added unit and pytest coverage for default vs overridden base-image selection.


airbyte_cdk/test/standard_tests/docker_base.py
```diff
@@ -252,6 +257,25 @@ def test_docker_image_build_and_check(
...
+            if message.connectionStatus is not None
+        ]
+        assert connection_statuses, (
+            f"`check` for connector '{connector_root.name}' emitted no CONNECTION_STATUS message. "
+            f"Logs: {check_result.logs}"
+        )
+        assert connection_statuses[-1].status == Status.SUCCEEDED, (
+            f"`check` for connector '{connector_root.name}' did not succeed: "
+            f"{connection_statuses[-1]}"
+        )
```
tolik0: `connector_root` here is `self.get_connector_root_dir()` without `.absolute()` (unlike `test_docker_image_build_and_spec` above), so `connector_root.name` is empty whenever the resolved root is a relative path — which is what happens in CI. From this PR's own `Check: destination-motherduck` run:

```
E       AssertionError: `check` for connector '' did not succeed: AirbyteConnectionStatus(status=<Status.FAILED: 'FAILED'>, message="An exception occurred: PermissionError(13, 'Permission denied')")
```

So the one message that tells you *which* connector failed drops the connector name. The class already exposes a `connector_name` classproperty (L57-60) that does `connector_root.absolute().name` — either that or an explicit `.absolute()` here works:

```suggestion
        assert connection_statuses, (
            f"`check` for connector '{connector_root.absolute().name}' emitted no "
            f"CONNECTION_STATUS message. Logs: {check_result.logs}"
        )
        assert connection_statuses[-1].status == Status.SUCCEEDED, (
            f"`check` for connector '{connector_root.absolute().name}' did not succeed: "
            f"{connection_statuses[-1]}"
        )
```

airbyte_cdk/test/standard_tests/docker_base.py
```diff
@@ -37,6 +38,44 @@
...
+            f"`check` for connector '{connector_name}' was expected to fail, but reported: "
+            f"{connection_statuses[-1]}"
+        )
+        return
+
+    assert connection_statuses, (
+        f"`check` for connector '{connector_name}' emitted no CONNECTION_STATUS message. "
+        f"Logs: {check_result.logs}"
+    )
+    if expected_outcome.expect_success():
```
tolik0, line 73: `ALLOW_ANY` is reachable here for scenarios that **do** declare `status: "succeed"`, because `_dedup_scenarios()` throws the declared status away.

`get_scenarios()` iterates `["spec", "connection", "basic_read"]`, and `_dedup_scenarios()` keeps the **first** scenario per `config_path` (merging only `empty_streams`). So a statusless entry in the `spec` section wins over the `connection` entry for the same config. And `connection` is the only section whose schema has a `status` field at all (`SpecTestConfig` has `config_path` but no `status`), so the one place the expectation can be declared is the place that gets discarded.

`destination-motherduck` is exactly that shape:

```yaml
  spec:
    tests:
      - spec_path: integration_tests/spec.json
        config_path: "integration_tests/config.json"   # no status -> ALLOW_ANY, kept
  connection:
    tests:
      - config_path: "integration_tests/config.json"
        status: "succeed"                              # deduped away
      - config_path: "secrets/config.json"
        status: "succeed"
```

Effect on this PR's own CI — same connector, same underlying failures, `connector test` step:

| head | result |
| --- | --- |
| e4dbd337 (assert unconditionally) | `2 failed, 3 passed, 2 skipped` |
| d525674d / 3b01ff8b (assert only on `expect_success`) | `1 failed, 4 passed, 2 skipped` |

The scenario that stopped failing is `integration_tests/config.json`. From e4dbd337, where the assertion was unconditional:

```
FAILED ...test_docker_image_build_and_check['config' Test Scenario0] - AssertionError: ... PermissionError(13, 'Permission denied')
FAILED ...test_docker_image_build_and_check['config' Test Scenario1] - AssertionError: ... Please check your MotherDuck token ...
```

Scenario0 is that config. Nothing about the connector or the runner environment has changed since, so its in-container `check` is still returning `FAILED` — it's just no longer asserted, despite the connector declaring `status: "succeed"` for it.

Worth noting CAT's own default here: `ConnectionTestConfig.status` defaults to `Status.Succeed`, so under CAT a connection entry with no `status` means "must succeed". `ConnectorTestScenario.status` defaults to `None` -> `ALLOW_ANY`, i.e. "either outcome passes". That divergence is what makes the dedup drop silent rather than loud.

The gating itself is reasonable and matches `_job_runner` — the problem is upstream. Suggestion: merge `status` in `_dedup_scenarios()`, preferring an explicitly declared status over `None` (and surfacing a conflict if two entries declare different ones). That keeps `ALLOW_ANY` meaning "genuinely undeclared" and fixes the in-process `test_check` path at the same time. Iterating `connection` before `spec` in `get_scenarios()` would also work, but reads as incidental rather than intentional.

Heads-up: this will put `destination-motherduck` back to 2 failures — which is the accurate state, since that check really is failing in the container.

Minor, visible in the same output: both configs have the stem `config`, so scenarios render as `['config' Test Scenario0]` / `[... Scenario1]` and the summary doesn't say which config failed. Worth deriving `_id` from the parent directory when stems collide.

airbyte_cdk/test/standard_tests/pytest_hooks.py
```diff
@@ -116,6 +128,18 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=None,
         help="Use this pre-built connector Docker image instead of building one.",
     )
+    parser.addoption(
+        "--connector-base-image",
```
tolik0, line 134: Two names for one concept, both introduced in this PR: the pytest option is `--connector-base-image` (here), while the CLI flag is `--base-image` (`_image.py:46` for `image build`, `_image.py:123` for `image test`).

The workflow ends up using both in adjacent steps, which is where it gets confusing:

```yaml
      - name: Run FAST Standard Tests [Manifest-Only Connectors]
        run: |
          poetry run airbyte-cdk connector test ${{ matrix.connector }} \
            --pytest-arg=--connector-base-image \
            --pytest-arg=airbyte/source-declarative-manifest:dev

      - name: Container Build and Test [All Connectors]
        run: |
          ...
            poetry run airbyte-cdk image test ${{ matrix.connector }} \
              --base-image airbyte/source-declarative-manifest:dev
```

Both flags are new in this PR, so there's no back-compat constraint on renaming either one — but once it ships, changing the CLI flag becomes a breaking change for anyone scripting it.

Aligning on `--connector-base-image` everywhere looks like the smaller edit, since the fixture (`connector_base_image_override`) and the plumbing arg (`base_image_override`) already read in that direction, and it disambiguates from `python-connector-base` / the connector's own base image. `--base-image` on both sides is equally fine — the point is just picking one.

tolik0 (approved):

Issue #1067, opened 2026-07-07
Bug - `incremental_dependency: true` triggers full refresh on child streams due to context loss in SubstreamPartitionRouter (Concurrent CDK)


leonmm2:
### [Bug] `incremental_dependency: true` triggers full refresh on child streams due to context loss in SubstreamPartitionRouter (Concurrent CDK)

*Note: This investigation stems from [Issue #61567](https://github.com/airbytehq/airbyte/issues/61567) already reported in the main Airbyte repository. To identify the root cause, I leveraged GitHub Copilot and Gemini Pro to analyze the current architecture of the `airbyte-python-cdk` repository. The findings point to a specific architectural gap causing the bug.*

#### Problem Summary
There is a severe performance degradation in the Concurrent CDK affecting any connector built with the Declarative CDK / Connector Builder that uses hierarchical relationships with `incremental_dependency: true`.

- **Initial Sync (Sync 1):** Works correctly. The `streamState` is persisted with the parent's cursor checkpoints.
- **Subsequent Syncs (Sync 2+):** The child stream loses its state context, forcing the parent stream to ignore its saved cursor and restart data extraction from the beginning (`start_date`).
- **Impact:** Delta syncs are 100x to 1000x slower. This affects production connectors such as `source-gmail`, `source-mixpanel`, `source-instagram`, `source-typeform`, and custom implementations (e.g., `source-dolibarr`).

#### Root Cause Analysis (Code Investigation)
The issue lies in how the concurrent engine fails to propagate the state context within the `SubstreamPartitionRouter`.

**Target File:** `airbyte_cdk/sources/declarative/partition_routers/substream_partition_router.py`

**1. Context loss in `cursor_slice` (Line ~214):**
Inside the `stream_slices` generator, when iterating over partitions (`partition.read()`), the `StreamSlice` is built and yielded with a hardcoded empty dictionary for the cursor:
```python
yield StreamSlice(
    partition={
        partition_field: partition_value,
        "parent_slice": parent_partition or {},
    },
    cursor_slice={},  # <-- Parent's cursor context is lost here
    extra_fields=extracted_extra_fields,
)
```
By yielding an empty `cursor_slice`, downstream components that rely on this slice to maintain the parent's checkpoint fail to associate the exact cursor bounds.

**2. Brittle state migration (`_migrate_child_state_to_parent_state` - Line ~261):**
[...]

PR #1099, opened 2026-07-31
fix(low-code): warn when incremental_dependency is set on a stream without incremental_sync

lazebnyi:
## What

Log a warning when a stream has `incremental_dependency: true` in its parent stream configuration but does not define `incremental_sync` of its own.

## Why

The parent stream's cursor is only persisted under the `parent_state` key of an incremental substream's state. On a full-refresh substream this setting is silently ignored — the parent is re-read from the beginning and all child partitions are re-fetched on every sync. This is exactly the trap behind https://github.com/airbytehq/airbyte/issues/61567 / #1067.

## How

`create_default_stream` now checks the stream's partition router (including routers wrapped in `GroupingPartitionRouter` and lists of routers) and logs a warning explaining the misconfiguration and how to fix it.

## Test

- New parametrized unit test in `test_model_to_component_factory.py`: a full-refresh substream with `incremental_dependency: true` emits the warning; the same substream with `incremental_sync` does not.
- Verified end-to-end with a mock-server read modeled on the Dolibarr manifest from the linked issue: the warning appears in the sync logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a warning when incremental dependency settings are configured on a stream without incremental synchronization, helping identify ineffective configuration.

* **Tests**
  * Added coverage to verify warnings appear only when the relevant incremental synchronization setting is missing.

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2034,6 +2034,7 @@ class ModelToComponentFactory:
     ) -> AbstractStream:
         primary_key = model.primary_key.__root__ if model.primary_key else None
         self._migrate_state(model, config)
+        self._warn_on_ineffective_incremental_dependency(model)
 
         partition_router = self._build_stream_slicer_from_partition_router(
             model.retriever,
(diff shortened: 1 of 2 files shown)


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2190,6 +2191,36 @@ def create_default_stream(
...
+                router = router.underlying_partition_router
+            if isinstance(router, SubstreamPartitionRouterModel) and any(
+                parent_stream_config.incremental_dependency
+                for parent_stream_config in router.parent_stream_configs
+            ):
+                LOGGER.warning(
+                    f"Stream `{model.name}` has `incremental_dependency: true` in its parent stream configuration but does not define `incremental_sync`. "
+                    "The parent stream's cursor is only persisted in the state of an incremental substream, so this setting has no effect and all parent records will be re-read on every sync. "
+                    "Define `incremental_sync` on this stream or remove `incremental_dependency`."
+                )
```
pnilan, line 2237: why warn instead of raise?

pnilan (approved):
approving, one non-blocking comment.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
lazebnyi: a warning is safer because raising would hard-break existing deployed connectors that already ship this combination

PR #1103, opened 2026-08-05
fix(low-code): make Spec.generate_spec idempotent and non-mutating

darynaishchenko:
## Summary

`Spec.generate_spec()` converted the `advanced_auth` enum fields (`auth_flow_type`, nested `scopes_join_strategy`) to their string values by assigning the converted values back onto the typed model. That in-place mutation meant:

- a second `generate_spec()` call in the same process raised `AttributeError` (`.value` on an already-converted `str`) — hit by any flow that generates the spec more than once (Connector Builder, tests);
- any other reader of `self.advanced_auth` afterwards saw a string where the type system promises an enum.

This change serializes the model to a dict first and normalizes enum values only in that throwaway copy, so repeated calls are idempotent and the typed model is never mutated. Normalization happens via a single JSON round-trip (`json.loads(self.advanced_auth.json())`) rather than a hand-written walk over the two known enum fields — `declarative_component_schema.py` is code-generated and already defines 12 `Enum` classes, so an explicit walk would silently pass an `Enum` through to `ConnectorSpecificationSerializer` the next time an enum field is added anywhere under the `advanced_auth` subtree.

## Behavior change: `advanced_auth` without an `auth_flow_type`

`AuthFlow` is a pydantic model, so `if self.advanced_auth:` is always truthy. A manifest that sets `advanced_auth` with only a `predicate_key` (no `auth_flow_type`) previously hard-failed with `AttributeError: 'NoneType' object has no attribute 'value'`; it now emits `AdvancedAuth(auth_flow_type=None, ...)`.

Passing `None` through is the more correct behavior — both `AuthFlow.auth_flow_type` (`declarative_component_schema.py:1618`) and the protocol `AdvancedAuth.auth_flow_type` (`airbyte_cdk/models/airbyte_protocol.py:119`) are `Optional`. This PR pins that semantic with a dedicated test rather than leaving it implicit.

## Context

Split out of #1066, where this fix was bundled with the (unrelated) `RateLimitedMultipleTokenAuthenticator` feature; review feedback there asked for it to be its own PR. The fix is being removed from #1066.

## Testing

`unit_tests/sources/declarative/spec/test_spec.py` — 15 passed. The idempotency test is parametrized over three `advanced_auth` shapes so both normalization branches and the `None` case are covered:

- `top_level_auth_flow_type_enum`
- `nested_scopes_join_strategy_enum` — asserts the emitted `scopes_join_strategy` is the plain string the protocol declares. That field is typed `Optional[str]`, so an un-normalized enum would otherwise slip past both the serializer and the `first == second` check.
- `no_auth_flow_type`

Verified the new assertions actually bite: reverting `generate_spec` to a plain `.dict()` with no normalization fails 4 tests, including both new parametrized cases.

`mypy`, `ruff check`, and `ruff format --check` are clean on both touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication specification generation for enum-based authentication flows and OAuth scope joining.
  * Ensured repeated specification generation produces consistent results without altering configured authentication settings.
  * Improved handling of authentication flows without a specified flow type.

* **Tests**
  * Added regression coverage for repeatable specification generation, plain protocol values, configuration immutability, and flow-type edge cases.

### DIFF
--- a/airbyte_cdk/sources/declarative/spec/spec.py
+++ b/airbyte_cdk/sources/declarative/spec/spec.py
@@ -3,6 +3,7 @@
 #
 
 from dataclasses import InitVar, dataclass, field
+from enum import Enum
 from typing import Any, List, Mapping, MutableMapping, Optional
 
 from airbyte_cdk.models import (
@@ -54,19 +55,17 @@ class Spec:
         if self.documentation_url:
             obj["documentationUrl"] = self.documentation_url
         if self.advanced_auth:
-            self.advanced_auth.auth_flow_type = self.advanced_auth.auth_flow_type.value  # type: ignore # We know this is always assigned to an AuthFlow which has the auth_flow_type field
-            # Convert scopes_join_strategy enum to its string value (same pattern as auth_flow_type above)
-            oauth_spec = getattr(self.advanced_auth, "oauth_config_specification", None)
-            if oauth_spec:
-                oauth_input = getattr(oauth_spec, "oauth_connector_input_specification", None)
-                if (
-                    oauth_input
-                    and hasattr(oauth_input, "scopes_join_strategy")
-                    and oauth_input.scopes_join_strategy is not None
-                ):
-                    oauth_input.scopes_join_strategy = oauth_input.scopes_join_strategy.value  # type: ignore
+            # Serialize to a dict and normalize enum values there so the typed model is
+            # never mutated and repeated calls produce the same result
+            advanced_auth = self.advanced_auth.dict()
+            if isinstance(advanced_auth.get("auth_flow_type"), Enum):
+                advanced_auth["auth_flow_type"] = advanced_auth["auth_flow_type"].value
+            oauth_spec = advanced_auth.get("oauth_config_specification") or {}
+            oauth_input = oauth_spec.get("oauth_connector_input_specification") or {}
+            if isinstance(oauth_input.get("scopes_join_strategy"), Enum):
+                oauth_input["scopes_join_strategy"] = oauth_input["scopes_join_strategy"].value
             # Map CDK AuthFlow model to protocol AdvancedAuth model
-            obj["advanced_auth"] = self.advanced_auth.dict()
+            obj["advanced_auth"] = advanced_auth
 
         # We remap these keys to camel case because that's the existing format expected by the rest of the platform
         return ConnectorSpecificationSerializer.load(obj)
--- a/airbyte_cdk/sources/declarative/spec/spec.py
+++ b/airbyte_cdk/sources/declarative/spec/spec.py
@@ -2,8 +2,8 @@
 # Copyright (c) 2025 Airbyte, Inc., all rights reserved.
 #
 
+import json
 from dataclasses import InitVar, dataclass, field
-from enum import Enum
 from typing import Any, List, Mapping, MutableMapping, Optional
 
 from airbyte_cdk.models import (
@@ -55,17 +55,13 @@ class Spec:
         if self.documentation_url:
             obj["documentationUrl"] = self.documentation_url
         if self.advanced_auth:
-            # Serialize to a dict and normalize enum values there so the typed model is
-            # never mutated and repeated calls produce the same result
-            advanced_auth = self.advanced_auth.dict()
-            if isinstance(advanced_auth.get("auth_flow_type"), Enum):
-                advanced_auth["auth_flow_type"] = advanced_auth["auth_flow_type"].value
-            oauth_spec = advanced_auth.get("oauth_config_specification") or {}
-            oauth_input = oauth_spec.get("oauth_connector_input_specification") or {}
-            if isinstance(oauth_input.get("scopes_join_strategy"), Enum):
-                oauth_input["scopes_join_strategy"] = oauth_input["scopes_join_strategy"].value
+            # Serialize through JSON so enum values are normalized at any depth and the typed
+            # model is never mutated — repeated calls produce the same result.
+            # Note: an AuthFlow without an auth_flow_type (e.g. only a predicate) is passed
+            # through as auth_flow_type=None, which the protocol AdvancedAuth allows.
+            advanced_auth_dict = json.loads(self.advanced_auth.json())
             # Map CDK AuthFlow model to protocol AdvancedAuth model
-            obj["advanced_auth"] = advanced_auth
+            obj["advanced_auth"] = advanced_auth_dict
 
         # We remap these keys to camel case because that's the existing format expected by the rest of the platform
         return ConnectorSpecificationSerializer.load(obj)
--- a/unit_tests/sources/declarative/spec/test_spec.py
+++ b/unit_tests/sources/declarative/spec/test_spec.py
@@ -162,6 +162,24 @@ def test_spec(spec, expected_connection_specification) -> None:
     assert spec.generate_spec() == expected_connection_specification
 
 
+def test_generate_spec_is_idempotent_and_does_not_mutate_the_model() -> None:
+    spec = component_spec(
+        connection_specification={"client_id": "my_client_id"},
+        parameters={},
+        advanced_auth=component_auth_flow(
+            auth_flow_type=component_auth_flow_type.oauth2_0,
+            predicate_key=None,
+            predicate_value=None,
+        ),
+    )
+
+    first = spec.generate_spec()
+    second = spec.generate_spec()
+
+    assert first == second
+    assert spec.advanced_auth.auth_flow_type is component_auth_flow_type.oauth2_0
+
+
 def test_given_list_of_transformations_when_transform_config_then_config_is_transformed() -> None:
     input_config = {"planet_code": "CRSC"}
     expected_config = {
--- a/unit_tests/sources/declarative/spec/test_spec.py
+++ b/unit_tests/sources/declarative/spec/test_spec.py
@@ -37,6 +37,9 @@ from airbyte_cdk.sources.declarative.models.declarative_component_schema import 
     OauthConnectorInputSpecification as component_declarative_oauth_connector_input_spec,
 )
 from airbyte_cdk.sources.declarative.models.declarative_component_schema import (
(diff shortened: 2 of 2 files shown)


tolik0 (commented):
Nice, well-scoped fix — thanks for splitting it out of #1066.

I verified this locally against the PR head:

- The bug reproduces on `main`: a second `generate_spec()` raises `AttributeError: 'str' object has no attribute 'value'`.
- On this branch, two calls compare equal, the model keeps its enum, and the nested `scopes_join_strategy` path also round-trips correctly (`'comma'` in the output, `ScopesJoinStrategy.comma` still on the model).
- `pytest unit_tests/sources/declarative/spec/test_spec.py` → 12 passed; `mypy`, `ruff check`, and `ruff format --check` all clean.

The approach is sound. Pydantic v1 `.dict()` preserves `Enum` members (no `use_enum_values` on these models) and builds fresh nested dicts, so the `isinstance(..., Enum)` guards fire and the in-place mutation of `oauth_input` correctly propagates into the parent dict. Nothing else in the CDK reads `advanced_auth.auth_flow_type`, so no caller depended on the old mutation side effect. Dropping the two `# type: ignore` comments is a nice bonus.

Two things worth addressing before merge (inline), plus one optional simplification.

airbyte_cdk/sources/declarative/spec/spec.py
```diff
@@ -54,19 +55,17 @@ def generate_spec(self) -> ConnectorSpecification:
...
-                if (
-                    oauth_input
-                    and hasattr(oauth_input, "scopes_join_strategy")
-                    and oauth_input.scopes_join_strategy is not None
-                ):
-                    oauth_input.scopes_join_strategy = oauth_input.scopes_join_strategy.value  # type: ignore
+            # Serialize to a dict and normalize enum values there so the typed model is
+            # never mutated and repeated calls produce the same result
+            advanced_auth = self.advanced_auth.dict()
+            if isinstance(advanced_auth.get("auth_flow_type"), Enum):
```
tolik0: Worth calling out in the PR description: this quietly changes behavior when `advanced_auth` is set but `auth_flow_type` is `None` (e.g. a manifest with only `predicate_key`). `AuthFlow` is a pydantic model so `if self.advanced_auth:` is always truthy — previously that path hard-failed with `AttributeError: 'NoneType' object has no attribute 'value'`, and now it emits `AdvancedAuth(auth_flow_type=None, ...)`.

Both `AuthFlow.auth_flow_type` and the protocol `AdvancedAuth.auth_flow_type` are `Optional`, so passing `None` through is arguably the more correct behavior. But it turns a loud failure into a quiet one for a malformed manifest, and no test pins the intended semantics either way. A one-line test (or a note in the description) would settle it.

airbyte_cdk/sources/declarative/spec/spec.py
```diff
@@ -54,19 +55,17 @@ def generate_spec(self) -> ConnectorSpecification:
...
+                advanced_auth["auth_flow_type"] = advanced_auth["auth_flow_type"].value
+            oauth_spec = advanced_auth.get("oauth_config_specification") or {}
+            oauth_input = oauth_spec.get("oauth_connector_input_specification") or {}
+            if isinstance(oauth_input.get("scopes_join_strategy"), Enum):
+                oauth_input["scopes_join_strategy"] = oauth_input["scopes_join_strategy"].value
```
tolik0: Optional simplification — consider letting pydantic normalize enums recursively instead of hand-walking the two known fields:

```python
# Serialize through JSON so enum values are normalized at any depth and the
# typed model is never mutated — repeated calls produce the same result
obj["advanced_auth"] = json.loads(self.advanced_auth.json())
```

I confirmed this yields `'space'` for the nested `scopes_join_strategy`. `declarative_component_schema.py` is code-generated and already has ~12 `Enum` classes, so the current form will silently pass an `Enum` through to `ConnectorSpecificationSerializer` the next time an enum field is added anywhere under the `advanced_auth` subtree.

Trade-off: `.json()` also coerces other non-JSON types (tuples → lists), none of which exist in this model tree, and it's marginally less explicit about *what* is normalized. Cost is negligible (once per spec call). Entirely your call — the explicit version is defensible; the generic one just won't need editing again.

Tiny nit while you're here: `advanced_auth` reads as the model when it's actually the serialized copy — `advanced_auth_dict` would make the `obj["advanced_auth"] = advanced_auth` line clearer.

unit_tests/sources/declarative/spec/test_spec.py
```diff
@@ -162,6 +162,24 @@ def test_spec(spec, expected_connection_specification) -> None:
...
+            predicate_key=None,
+            predicate_value=None,
+        ),
+    )
+
+    first = spec.generate_spec()
+    second = spec.generate_spec()
+
+    assert first == second
+    assert spec.advanced_auth.auth_flow_type is component_auth_flow_type.oauth2_0
```
tolik0: This covers the `auth_flow_type` branch but not `scopes_join_strategy` — which is the subtler half of the change, since it depends on the nested-dict mutation propagating back into `advanced_auth`. `test_declarative_oauth_flow` exercises that field, but only for a *single* `generate_spec()` call, so a regression in the nested path wouldn't be caught by anything here.

Suggest parametrizing this test over both `advanced_auth` shapes (or reusing the declarative-OAuth fixture from the case above) so both normalization branches get idempotency coverage. I verified the nested path does work today — this is purely about locking it in.

Also: `assert ... is component_auth_flow_type.oauth2_0` is the right check here (identity, not equality) — that catches a mutation to the string `"oauth2.0"` which `==` would not.

airbyte_cdk/sources/declarative/spec/spec.py
darynaishchenko: Good catch — agreed the `None` passthrough is the correct semantic, and agreed it shouldn't be left implicit. Done both: added a note to the PR description and pinned it with a test rather than relying on the description alone.

Confirmed the old behavior first — on `main`, an `AuthFlow` with only a predicate hits `AttributeError: 'NoneType' object has no attribute 'value'`, exactly as you described. Both sides are `Optional` (`declarative_component_schema.py:1618` and `airbyte_cdk/models/airbyte_protocol.py:119`), so `AdvancedAuth(auth_flow_type=None, ...)` is a legitimate output.

New test `test_generate_spec_without_auth_flow_type_emits_advanced_auth_with_none` asserts the exact emitted `AdvancedAuth`, and `no_auth_flow_type` is now one of the parametrized idempotency cases.

airbyte_cdk/sources/declarative/spec/spec.py
darynaishchenko: Took the generic version — the "won't need editing again" argument wins here, since the schema is regenerated every time a component is added and nothing would fail loudly if a new enum landed under this subtree.

```python
advanced_auth_dict = json.loads(self.advanced_auth.json())
```

Verified equivalence before switching: ran the explicit walk and the JSON round-trip against a full `AuthFlow` (both enums set, `State`, `access_token_headers`/`params`, `extract_output`, a `complete_oauth_output_specification` blob, plus `Extra.allow` extras on `OAuthConfigSpecification`) and the two dicts compare equal. On the tuple-coercion trade-off you flagged: no tuples in the tree today, and for a manifest-sourced model everything under here is already JSON, so the coercion is a no-op. If anything it's slightly safer — YAML can hand us a `datetime.date` from an unquoted date, which `.dict()` would pass straight through to the serializer.

Also renamed to `advanced_auth_dict` as suggested, and dropped the now-unused `from enum import Enum`.

unit_tests/sources/declarative/spec/test_spec.py
darynaishchenko: Parametrized as suggested — the idempotency test now runs over three `advanced_auth` shapes: `top_level_auth_flow_type_enum`, `nested_scopes_join_strategy_enum` (non-default `comma`), and `no_auth_flow_type`.

Chasing your point turned up something worth flagging: **the nested branch needs more than an idempotency assertion to be covered.** `first == second` plus a model-identity check would still pass on a regression there, because the protocol declares `scopes_join_strategy: Optional[str]` (`airbyte_cdk/models/airbyte_protocol.py:94`) — it isn't an enum-typed field, so an un-normalized `ScopesJoinStrategy` member sits in it happily and both calls produce the same wrong object. So the nested case also asserts on the emitted value:

```python
assert emitted == "comma" and type(emitted) is str
```

Mutation-tested the result: reverting `generate_spec` to a plain `.dict()` with no normalization fails 4 tests, including both new parametrized cases. Without that `type(...) is str` assertion the nested case passes against the broken implementation.

And noted on identity vs equality — kept that, generalized to `is advanced_auth.auth_flow_type` so it holds for the `None` case too.

darynaishchenko:
@coderabbitai addressing your two nitpicks — one applied, one doesn't hold:

**`scopes_join_strategy` coverage** — applied. Same point @tolik0 raised; the idempotency test is now parametrized over the top-level, nested and `None` shapes. Details in [that thread](https://github.com/airbytehq/airbyte-python-cdk/pull/1103#discussion_r3720050099).

**`dict()` → `model_dump(mode="python")`** — not applicable here. These aren't Pydantic v2 models. `declarative_component_schema.py` is generated against the v1 compat shim:

```python
# airbyte_cdk/sources/declarative/models/declarative_component_schema.py:9
from pydantic.v1 import BaseModel, Extra, Field
```

`pydantic.v1.BaseModel` has no `model_dump` — verified on the pinned version:

```
$ python -c "from pydantic.v1 import BaseModel; print([m for m in dir(BaseModel) if 'dump' in m])"
[]
```

So the proposed diff would be an `AttributeError` at runtime, not a deprecation cleanup. `.dict()` is the correct v1 API and emits no deprecation warning from a `pydantic.v1` model. Moot in the end — the call was replaced with `json.loads(self.advanced_auth.json())` (also v1 API) for unrelated reasons.

darynaishchenko:
I'll close this for now as this only needed for tests in source-github and they already fixed, generally we don't have a case in prod where spec can be called twice in the way described in the issue, we will re-open and finalize the change once it's really needed.

PR #1106, opened 2026-08-06
fix(low-code): data feed stop condition with client-side incremental

darynaishchenko:
## What

When a `DatetimeBasedCursor` sets both `is_data_feed: true` and `is_client_side_incremental: true`, the pagination stop condition silently never fires: incremental syncs re-fetch the stream's complete listing on every run and rely on the client-side filter to drop the stale records. This PR makes the two flags compose, so pagination stops on the first page that contains a record older than the cursor while output stays identical.

Found while reviewing the source-github `repositories` stream migration (airbytehq/airbyte#81428), where the legacy Python stream used the sorted-desc early exit that the manifest equivalent lost.

## How

Root cause: `SimpleRetriever._read_pages` takes `last_record` from the post-filter record pipeline, and `ClientSideIncrementalRecordFilterDecorator` drops exactly the records whose `should_be_synced` is false — so `CursorStopCondition`, which only sees records that survived filtering, can never observe a stale one.

The filter already evaluates `should_be_synced` on every raw record, so it now records that fact: it tracks whether the current page contained a record older than the cursor (per thread, because a single retriever — and therefore a single filter — is shared across partitions that are read concurrently; the flag is reset at the start of each `filter_records` call). A new `FilterAwareStopCondition` consults that flag instead of `last_record`, and the factory wires it in place of `CursorStopCondition` when both flags are set.

`StopConditionPaginationStrategyDecorator` now consults the stop condition even when the page yielded no records, since a page whose records were all filtered out must still stop the feed; `PaginationStopCondition.is_met` accordingly accepts `Optional[Record]` and `CursorStopCondition` treats "no record" as not met (same behavior as before).

Empirically verified on a 2-page sorted-desc stream with the state cursor falling inside page 1: before, both pages were fetched; after, only page 1 is fetched and the emitted records are unchanged. A first sync with no stale records still paginates to the natural end (covered by the new integration test).

## Changes

- `ClientSideIncrementalRecordFilterDecorator` tracks a thread-local `stale_record_seen_on_current_page` flag, reset on every `filter_records` call
- New `FilterAwareStopCondition` stops pagination when the filter observed a below-cursor record on the current page
- `StopConditionPaginationStrategyDecorator` evaluates the stop condition even when `last_record` is `None`; `CursorStopCondition` is `None`-safe
- `ModelToComponentFactory.create_default_paginator` wires `FilterAwareStopCondition` when both `is_data_feed` and `is_client_side_incremental` are set, keeping `CursorStopCondition` otherwise
- Unit tests for the filter flag (including per-thread isolation), the new stop condition, the decorator's no-record behavior, and factory wiring, plus an end-to-end regression test reading a mocked 2-page data feed

## Recommended Review Order

1. `airbyte_cdk/sources/declarative/requesters/paginators/strategies/stop_condition.py`
2. `airbyte_cdk/sources/declarative/extractors/record_filter.py`
3. `airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py`

## Summary by CodeRabbit

* **New Features**
  * Improved incremental data-feed syncing by filtering previously synced and future-dated records after pagination.
  * Added reliable stop conditions for descending cursor feeds.
  * Ensured independent cursor handling across concurrent partitions.
  * Preserved pagination tracking while excluding filtered records from emitted results.
  * Supported consistent filtering across standard and lazy data-feed retrieval.

* **Documentation**
  * Clarified when data-feed pagination filtering replaces client-side incremental filtering.


airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3422,6 +3422,16 @@ def _get_url(req: Requester) -> str:
...
             cursor = FinalStateCursor(name, None, self._message_repository)
 
+        # A data feed drops the records the cursor considers already synced in the retriever, which
+        # sits downstream of the paginator. Letting the record selector drop them as well would be
+        # redundant and would hide them from the pagination stop condition, so a data feed never
+        # delegates that filtering to the record selector, whether `is_client_side_incremental` is
+        # set or not.
+        data_feed_cursor = cursor if has_stop_condition_cursor else None
+        client_side_incremental_cursor = (
+            cursor if is_client_side_incremental_sync and not data_feed_cursor else None
```
tolik0: **Blocking — this silently flips `transform_before_filtering` to `False`.**

`create_record_selector` carries two unrelated things in one branch:

```python
transform_before_filtering = (
    False if model.transform_before_filtering is None else model.transform_before_filtering
)
if client_side_incremental_sync_cursor:
    record_filter = ClientSideIncrementalRecordFilterDecorator(...)
    transform_before_filtering = (
        True if model.transform_before_filtering is None else model.transform_before_filtering
    )
```

Since a data feed now never enters that branch, it loses the `True` default along with the decorator. The user's own `record_filter.condition` is still built as a plain `RecordFilter`, but it now runs *before* transformations instead of after.

Any stream with both flags whose `record_filter.condition` references a transformation-produced field now filters against a record that does not have that field yet. Reproduced with an `AddFields` adding `keep: "yes"` and `condition: "{{ record['keep'] == 'yes' }}"`:

```
base   4758b150 → transform_before_filtering = True   → EMITTED: [{'id': '1', ..., 'keep': 'yes'}]
branch 16d72e9d → transform_before_filtering = False  → EMITTED: []
```

Total record loss with no error, and the affected population is exactly the one this PR targets — `is_client_side_incremental` is the flag those connectors set.

Suggested fix: decouple the default from *which component* performs the cursor filtering. Default `transform_before_filtering` to `True` when the stream is client-side incremental **or** a data feed, regardless of whether `client_side_incremental_sync_cursor` is passed. Worth a factory test asserting `retriever.record_selector.transform_before_filtering is True` for a data feed stream that also sets `is_client_side_incremental`.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -457,7 +461,15 @@ def read_records(
...
         )
-        yield from self._read_pages(record_generator, _slice)
+        # A data feed paginates until it reaches a record older than the cursor, so the page that
+        # triggers the stop condition still holds already-synced records. Those are dropped here
+        # rather than in the record selector so that the paginator keeps seeing the whole page: the
+        # stop condition is evaluated on the last record of the page, which is precisely one of the
+        # records being dropped.
+        data_feed_cursor = self.data_feed_cursor
+        for record in self._read_pages(record_generator, _slice):
+            if data_feed_cursor is None or data_feed_cursor.should_be_synced(record):
```
tolik0: `should_be_synced` is two-sided:

```python
# airbyte_cdk/sources/streams/concurrent/cursor.py:573
return self.start <= record_cursor_value <= self._end_provider()
```

With no `end_datetime` — the normal data feed shape — `_end_provider()` is `now()`. So this drops records dated *ahead* of the connector's clock as well as the already-synced ones at the tail of the boundary page.

Reproduced with `is_data_feed` only (no `is_client_side_incremental`), page 1 leading with a record dated `2099-01-01`:

```
base   4758b150 → PAGES ['1','2']  IDS [1, 2, 3]
branch 16d72e9d → PAGES ['1','2']  IDS [1, 2]      ← id 3 dropped
```

Not severe — pagination is not truncated, only the record is lost, and a few seconds of clock skew self-heals next sync once `now()` catches up. But for APIs with genuinely forward-dated cursor values (scheduled or pending items) the record stays dropped every sync until its date arrives.

Two things worth settling:

1. Whether the drop should be one-sided. The behaviour being modelled — "the boundary page still holds records from a previous sync" — is a lower-bound statement; filtering on the upper bound is a side effect of reusing `should_be_synced`.
2. Either way this changes emitted output for **every existing `is_data_feed` connector**, not only ones that opted into client-side filtering. That is broader than "make the two flags compose" and deserves an explicit callout in the PR description and release notes, plus a test pinning whichever semantics you pick.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1295,11 +1295,11 @@ definitions:
...
         type: boolean
       is_client_side_incremental:
         title: Client-side Incremental Filtering
-        description: Set to True if the target API endpoint does not take cursor values to filter records and returns all records anyway. This will cause the connector to filter out records locally, and only emit new records from the last sync, hence incremental. This means that all records would be read from the API, but only new records will be emitted to the destination.
+        description: Set to True if the target API endpoint does not take cursor values to filter records and returns all records anyway. This will cause the connector to filter out records locally, and only emit new records from the last sync, hence incremental. This means that all records would be read from the API, but only new records will be emitted to the destination. This is not needed when Data Feed API is enabled, as a data feed already filters out the records that were synced during a previous sync.
```
tolik0: Documenting the interaction is the right instinct, but a YAML description is the weakest place to put a rule the code enforces silently — the factory now ignores `is_client_side_incremental` on a data feed without saying so anywhere the connector developer will look.

And the combination is not actually inert: the blocking comment on `model_to_component_factory.py` shows setting both flags today changes `transform_before_filtering`. A rule that only exists in prose is exactly how that slipped through.

Suggest backing it with a `LOGGER.warning` when both flags are set, alongside these descriptions, so it surfaces in sync logs rather than only in the manifest schema:

```python
if has_stop_condition_cursor and is_client_side_incremental_sync:
    LOGGER.warning(
        f"Stream {name}: `is_client_side_incremental` is ignored when `is_data_feed` is set, "
        "as a data feed already filters out records synced during a previous sync."
    )
```

Separately, the wording here ("those are filtered out as well") describes one-sided filtering, while the implementation is two-sided — see the comment on `simple_retriever.py`.

tolik0 (changes requested):
The rework in `16d72e9d` is a real improvement over the first approach. Filtering downstream of `_read_pages` means the paginator sees the page exactly as the API returned it, so `CursorStopCondition` needs no modification at all — no thread-local, no coupling between `requesters.paginators` and `extractors`, no widening of a public ABC. `test_given_data_feed_cursor_when_read_records_then_paginator_still_sees_the_whole_page` asserts the invariant that actually matters (`last_page_size == 3`, `last_record == page[-1]`) rather than an observable side effect, and the multi-partition integration test is a genuine concurrency test.

Two problems, one blocking. Both reproduced against `4758b150`.

**1. Blocking — `transform_before_filtering` silently flips to `False`.** Routing the cursor away from `create_record_selector` also drops the `True` default that branch carried. A data feed stream with `is_client_side_incremental` and a `record_filter.condition` over a transformed field now filters against untransformed records. With an `AddFields` adding `keep: "yes"` and `condition: "{{ record['keep'] == 'yes' }}"`:

```
base   4758b150 → transform_before_filtering = True   → EMITTED: [{'id': '1', ..., 'keep': 'yes'}]
branch 16d72e9d → transform_before_filtering = False  → EMITTED: []
```

Total record loss, no error. Details inline.

**2. `is_data_feed` alone now drops records above the cursor's end boundary.** `should_be_synced` is two-sided and `_end_provider()` is `now()` when there is no `end_datetime`, so forward-dated records are discarded too — a change in emitted output for every existing data feed connector, not just ones using client-side filtering. Details inline.

Two smaller things that have no line in the diff to hang off:

- **`file_uploader` now runs before the drop.** `RecordSelector.filter_and_transform` calls `file_uploader.upload(record)` while building records, and the data feed drop is downstream in the retriever. A data feed stream with a `file_uploader` will fetch and upload files for already-synced boundary-page records and then discard the records pointing at them. Under the old client-side-incremental path the filter ran before the upload.
- **`PaginationTracker.observe` sees records the sync never emits.** Harmless today — I checked, the tracker only holds a cursor for `pagination_reset: SPLIT_USING_CURSOR` and it is a `copy_without_state()` clone, so real stream state is untouched. But if `pagination_reset` and `is_data_feed` are ever combined, slice reduction would be computed from records that were never emitted. Worth a one-line comment noting the drop happens after `observe()`.

For what it's worth on the design question: the "a retriever shouldn't know about cursors" objection doesn't really hold — `stream_slicer` already *is* the concurrent cursor for incremental streams and the paginator already holds the same object via `CursorStopCondition`, so `data_feed_cursor` makes existing knowledge explicit rather than introducing new coupling. The field is fine.

The thing I would file as a follow-up rather than fix here: cursor-based filtering now has two implementations of the same `should_be_synced` call in two layers, selected by a factory conditional — and that duplication is precisely what produced problem 1. The underlying invariant is "the paginator must see the raw page", and `OffsetIncrement` already carries an optional `extractor` as a manual escape hatch for exactly that (`offset_increment.py:80-86`), because any `record_filter` that shortens a page makes offset pagination stop early today. Solving that once — hand the paginator the raw page's last record and size — would cover this case with no new plumbing and fix the latent bug too.

Local verification: `ruff` and `mypy` clean; 268 passed / 2 failed across `retrievers/`, `paginators/` and `test_model_to_component_factory.py`, and those 2 (`test_lazy_simple_retriever.py`) fail identically on `4758b150`, so they are pre-existing. CI's `destination-motherduck` failure is a connection `check`, unrelated to this diff, but worth a re-run rather than an assumption.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
darynaishchenko: Good catch — fixed in `ef4ffc13`, and your repro is exactly right.

The default was tied to *which component* performs the cursor comparison. It now follows the flag instead:

```python
# create_record_selector
default_transform_before_filtering = bool(
    client_side_incremental_sync_cursor or is_client_side_incremental_sync
)
```

`create_simple_retriever` passes `is_client_side_incremental_sync` straight through, so the data feed's missing cursor no longer changes when the condition runs.

Verified with your scenario (`AddFields` adding `keep: "yes"`, `condition: "{{ record['keep'] == 'yes' }}"`):

```
3e8c4878  transform_before_filtering: False  EMITTED: []
ef4ffc13  transform_before_filtering: True   EMITTED: [{'id': '1', 'updated_at': '2022-06-01T00:00:00Z', 'keep': 'yes'}]
```

One deliberate difference from your suggested fix: the default keys off `is_client_side_incremental` alone, not "client-side incremental **or** a data feed". Defaulting to `True` for data feeds too would flip the default for data-feed-only streams that set a `record_filter.condition`, which is a behaviour change `main` doesn't have. Two tests pin both sides:

- `test_given_data_feed_and_client_side_incremental_then_transform_before_filtering` → `True`
- `test_given_data_feed_and_record_filter_then_condition_stays_in_the_record_selector` → `False` for data-feed-only

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
darynaishchenko: Investigated this one carefully and decided to keep the two-sided filtering. Reasoning, with the evidence:

**It matches what `is_client_side_incremental` already does.** Same feed, same forward-dated record, `is_client_side_incremental` only, run against `main`:

```
MAIN CSI-ONLY EMITTED: ['recent', 'older']
MAIN CSI-ONLY STATE:   2024-06-01T00:00:00Z
```

Identical to what a data feed now does, and for the same reason — `ClientSideIncrementalRecordFilterDecorator` calls `should_be_synced`. So this isn't a new semantic, it's alignment with the sibling flag. Connectors setting *both* flags see no change at all; only data-feed-only connectors do.

**One correction to the comment:** it does not self-heal. Next sync the feed still returns the record at the top, `should_be_synced` still rejects it, and it is dropped again — every sync until its date arrives.

**But one-sided is worse in the other direction.** `cursor.observe` runs in `partition_reader.py:78-81` on records that come *out* of `partition.read()`, i.e. post-filter (`PaginationTracker`'s cursor is `None` unless `pagination_reset: SPLIT_USING_CURSOR`, and then it's a `copy_without_state()` clone). So dropping the record also keeps it out of state:

```
branch   EMITTED ['recent', 'older']             STATE 2024-06-01
main     EMITTED ['future', 'recent', 'older']   STATE 2099-01-01
```

Replaying a sync with that 2099 state:

```
SECOND SYNC (state=2099) EMITTED: []  requests=0
```

Zero requests — `start` is past `now()`, so the cursor emits no slices and the stream syncs nothing, permanently. Going one-sided would restore that failure mode to win back a single record.

Pinned in `test_given_record_dated_in_the_future_then_filter_it_out`, which asserts the forward-dated record is dropped without truncating pagination, and documents why.

The forward-dated record advancing `most_recent_cursor_value` past `now()` is a real bug, but it is `main`'s and predates this PR — same family as #1014. Happy to file it separately, and to call the data-feed-only output change out in the PR description and release notes if you'd like.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
darynaishchenko: Agreed on the substance — added the warning in `create_simple_retriever` (`ef4ffc13`):

```python
if post_pagination_filter and is_client_side_incremental_sync:
    LOGGER.warning(
        f"Stream {name}: `is_client_side_incremental` is ignored when `is_data_feed` is set, "
        "as a data feed already filters out the records that were synced during a previous sync."
    )
```

It shows up in a real sync:

```
{"type":"LOG","log":{"level":"WARN","message":"Stream items: `is_client_side_incremental` is ignored when `is_data_feed` is set, as a data feed already filters out the records that were synced during a previous sync."}}
```

Your point that the combination was not inert was right, and the `transform_before_filtering` bug is fixed in the same commit.

On the wording: the filtering is staying two-sided (reasoning in the `simple_retriever.py` thread — it matches `is_client_side_incremental`, and one-sided reintroduces state poisoning). So the description is indeed narrower than the behaviour. Leaving this thread open — say the word and I will reword both descriptions to talk about the cursor window rather than only previously-synced records, and regenerate the models.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3422,6 +3422,26 @@ def _get_url(req: Requester) -> str:
...
+                condition=None,
+                cursor=cursor,
+            )
+            if has_stop_condition_cursor
+            else None
+        )
+        client_side_incremental_cursor = (
+            cursor if is_client_side_incremental_sync and not post_pagination_filter else None
+        )
+
```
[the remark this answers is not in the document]
darynaishchenko: Fixed in `ef4ffc13`. `create_record_selector` now derives the default from the flag rather than from the cursor it receives:

```python
default_transform_before_filtering = bool(
    client_side_incremental_sync_cursor or is_client_side_incremental_sync
)
```

One deviation from the proposed direction: the default is not raised for *every* data feed, only for streams that set `is_client_side_incremental`. Keying it on `post_pagination_filter` would flip the default for data-feed-only streams with a `record_filter.condition`, which differs from `main`. Both cases are pinned:

- `test_given_data_feed_and_client_side_incremental_then_transform_before_filtering` → `True`
- `test_given_data_feed_and_record_filter_then_condition_stays_in_the_record_selector` → `False` for data-feed-only

That second test also covers the earlier nitpick about the `not isinstance(...)` assertion being trivially true — it uses a manifest that actually sets `record_filter.condition`, and asserts the selector keeps a plain `RecordFilter`.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
tolik0: Confirmed fixed in `ef4ffc13` — re-ran my repro against this head:

```
transform_before_filtering = True | EMITTED: [{'id': '1', 'updated_at': '2026-01-01T00:00:00Z', 'keep': 'yes'}]
```

And your deviation is the better call, not a compromise. Raising the default for data-feed-only streams would have flipped behaviour `main` doesn't have, purely to make my suggested rule read more symmetrically — keying it on the flag is narrower and correct. Good that both directions are pinned rather than just the one that broke.

Resolving.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
tolik0: You're right and I'm withdrawing this. I reproduced your argument independently rather than taking it on trust — data-feed-only stream, `2099-01-01` record at the top of a descending page, sync 1's state threaded into sync 2:

```
branch ef4ffc13 → SYNC2 emitted=['recent'] requests=1
origin/main    → SYNC2 emitted=[]          requests=0
```

Zero requests on `main`, exactly as you described. I also confirmed the mechanism you pointed at: `cursor.observe` runs in `partition_reader.py:78-81` on records coming *out* of `partition.read()`, so on `main` the forward-dated record is emitted, observed, and pushes state to 2099 — after which `start > now()` and the cursor emits no slices at all. Permanently, not until the next sync.

So two-sided filtering is buying protection from a stream-killing failure at the cost of one record, and going one-sided would trade in the wrong direction. Withdrawn.

Your correction on self-healing is also right — I had that wrong. The record is re-rejected every sync until its date arrives, not recovered on the next run.

Please do file the `main` state-poisoning bug separately; the forward-dated record advancing `most_recent_cursor_value` past `now()` is worth fixing on its own, and it's the thing that makes two-sided filtering load-bearing here rather than merely conservative.

Resolving.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
tolik0: Warning landed and reads well in the log output — thanks.

Taking you up on the reword: yes please, on both descriptions, plus regenerating `models/declarative_component_schema.py`. Now that two-sided filtering is a deliberate decision rather than a side effect, "records that were synced during a previous sync" is the one place left that still describes it as one-sided, and it's the text connector developers actually read. Framing it as the cursor window covers both ends.

Not blocking the approval — happy for it to land in this PR or a follow-up.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3422,6 +3426,31 @@ def _get_url(req: Requester) -> str:
         if cursor is None:
             cursor = FinalStateCursor(name, None, self._message_repository)
 
+        # A data feed drops the records the cursor considers already synced in the retriever, which sits downstream of
+        # the paginator. Letting the record selector drop them as well would be redundant and would hide them from the
+        # pagination stop condition, so a data feed never delegates that filtering to the record selector, whether
+        # `is_client_side_incremental` is set or not. The `condition` from `record_filter` is intentionally left out of
+        # the post-pagination filter: it stays in the record selector so that the records it rejects keep being counted
+        # by the paginator.
```
tolik0: This rationale is inverted, and it's the one bit worth fixing before merge.

> it stays in the record selector so that the records it rejects keep being counted by the paginator

I measured the opposite. Instrumenting `DefaultPaginator.next_page_token` on a data feed whose API page returned 3 records, one rejected by the `condition`:

```
PAGES: ['1'] | EMITTED: ['a']
paginator saw (last_page_size, last_record_id): [(2, 'c')]
```

`last_page_size` is 2, not 3. The record selector sits *upstream* of `_read_pages`, so a condition reject never reaches the paginator's accounting — keeping the condition there is exactly what stops those records from being counted. Moving it into the post-pagination filter is what would make them counted.

The decision is right: leaving the condition in the selector preserves `main`'s behaviour and is the conservative choice. Only the stated reason needs rewriting. Something like: the condition stays in the record selector to preserve existing behaviour, and is deliberately not moved into the post-pagination filter, which would change what the paginator counts.

Flagging it rather than letting it slide because a comment is what the next change gets justified from — the round-2 `transform_before_filtering` bug came from a default whose reason lived in the wrong place.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -457,7 +463,21 @@ def read_records(
...
-        yield from self._read_pages(record_generator, _slice)
+        records: Iterable[Mapping[str, Any]] = self._read_pages(record_generator, _slice)
+        if self.post_pagination_filter:
+            # A data feed paginates until it reaches a record older than the cursor, so the page that triggers the stop
+            # condition still holds already-synced records. Those are filtered here rather than in the record selector
+            # so that the paginator keeps seeing the whole page: the stop condition is evaluated on the last record of
+            # the page, which is precisely one of the records being dropped. Two consequences of filtering this late:
+            # the pagination tracker observes the dropped records, and a `file_uploader` on the record selector has
+            # already uploaded their files by the time they are dropped.
+            records = self.post_pagination_filter.filter_records(
```
tolik0, line 474: Minor consequence of reusing the selector's filter here: the retriever already holds real `Record` objects, and `filter_records` wraps each one again for the cursor call (`record_filter.py:71-77`):

```python
Record(data=record, associated_slice=stream_slice, stream_name="")
```

What the cursor actually receives:

```
cursor received: [('Record', 'Record', '', '2026-01-01')]   # (type, type(.data), stream_name, cursor value)
output type: Record | same object: True
```

So `.data` is a `Record` rather than a mapping, and `stream_name` is `""`. It works — `Record` is a `Mapping`, so cursor-value extraction recurses fine, and the objects yielded downstream are the originals, so `DeclarativePartition.read()` is unaffected. Not a regression against `main` either; this is long-standing behaviour for `is_client_side_incremental` streams.

But the inline version in `16d72e9d` passed the real record with its real stream name and slice, and reusing the selector-oriented filter gives that up. Any cursor keying on `record.stream_name` — including a custom `should_be_synced` — now sees `""` on data feeds too. A cheap guard in `filter_records` (skip the re-wrap when the item already is a `Record`) would fix it for both call sites.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3422,6 +3426,31 @@ def _get_url(req: Requester) -> str:
...
+                cursor=cursor,
+            )
+            if has_stop_condition_cursor
+            else None
+        )
+        client_side_incremental_cursor = (
+            cursor if is_client_side_incremental_sync and not post_pagination_filter else None
+        )
+        if post_pagination_filter and is_client_side_incremental_sync:
+            LOGGER.warning(
```
tolik0, line 3493: Cosmetic: this fires per stream construction rather than per sync, and `streams()` can be called more than once in a process (check and read paths), so the line can repeat in the logs. Not worth restructuring for — just noting it in case the duplication looks like a bug to someone reading a support log.

tolik0 (approved):
Approving. The blocking `transform_before_filtering` regression is fixed, and it's fixed at the right level — the default now follows the `is_client_side_incremental` flag rather than whichever component happens to receive the cursor, with both directions pinned by tests.

The bigger improvement is `post_pagination_filter` being a real `ClientSideIncrementalRecordFilterDecorator` owned by the retriever instead of an inline `should_be_synced` loop. That collapses the duplication that caused the round-2 bug in the first place: there is now one implementation of cursor filtering, positioned differently, rather than two selected by a factory conditional. Removing the failure mode rather than the instance.

**I've withdrawn the two-sided filtering objection** — reproduced your state-poisoning argument independently and it's decisive; details in that thread. Two-sided filtering is load-bearing here, not merely conservative, and my "self-heals next sync" note was wrong.

Three comment-level items below, none blocking. The first is one sentence and worth fixing before merge — a wrong rationale in a comment is what a future change gets argued from, which is precisely how the round-2 regression happened.

Also worth filing separately, beyond the `main` state-poisoning bug: `last_page_size` is post-filter. The comment thread below measures 2 for a 3-record page, and `OffsetIncrement` already carries an optional `extractor` as a manual escape hatch for exactly this (`offset_increment.py:80-86`) — which is why the same probe paginates correctly under `OffsetIncrement` but not `PageIncrement`. Handing the paginator the raw page's size and last record once would remove the need for that hatch and for any future component to work around filtered pages.

Verification on `ef4ffc13`: `ruff` and `mypy` clean; 272 passed / 2 failed across `retrievers/`, `paginators/` and `test_model_to_component_factory.py`, and those 2 (`test_lazy_simple_retriever.py`) fail identically on current `origin/main`, so they're pre-existing.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
darynaishchenko: Thanks for reproducing it independently rather than taking it on trust — and for the correction on your side too.

Filing the `main` state-poisoning bug separately, as you asked: the fix belongs in `ConcurrentCursor.observe`, clamping `most_recent_cursor_value` to `_end_provider()` so a forward-dated record cannot push state past `now()`. That is a cursor-level change affecting every incremental stream, so it does not belong in this PR. Once it lands, one-sided filtering here becomes strictly better than what we have — the record gets emitted *and* state stays sane — so it is worth revisiting this decision then.

The behaviour is pinned by `test_given_record_dated_in_the_future_then_filter_it_out`, which I have since restructured (per CodeRabbit) so the already-synced record sits on a second page: reaching page 2 now proves the forward-dated record is not what stops the pagination.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
darynaishchenko: Reworded both descriptions in `44165b99`, framed as the cursor window:

> **is_data_feed** — … The last page fetched still holds records that fall outside the cursor window, and those are filtered out as well, so Client-side Incremental Filtering does not need to be enabled alongside this field. Records are kept when their cursor value is within the window that starts at the previous sync's cursor value (or the start date) and ends at the end date, defaulting to the current time, so records dated in the future are filtered out too.

> **is_client_side_incremental** — … This will cause the connector to filter out records locally, keeping only the ones whose cursor value falls within the window that starts at the previous sync's cursor value (or the start date) and ends at the end date, defaulting to the current time. … This is not needed when Data Feed API is enabled, as a data feed already filters on the same window.

One note on the model regeneration. `poe assemble` produces a much larger diff than the two descriptions: it drops the `ScopesJoinStrategy` enum, adds reordering, and churns ~450 lines. That enum is in the committed `models/declarative_component_schema.py` but not in `declarative_component_schema.yaml`, and nothing references it except a comment in `airbyte_protocol.py` — so the committed model file appears to have drifted from the schema independently of this PR. Rather than fold that cleanup in here, I hand-edited the two description strings to match exactly what the generator emits for them. Happy to run the full regeneration in a separate PR if you would like the drift cleaned up.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
darynaishchenko: You're right, the rationale was backwards — the selector runs inside `_read_pages`, so a condition reject never reaches `last_page_size`/`last_record` at all. Rewritten in `44165b99`:

```python
# ... The `condition` from `record_filter` is intentionally left out of
# the post-pagination filter and stays in the record selector, which preserves the existing behaviour: the
# selector runs inside the page loop, so the records the condition rejects never reach the paginator's
# accounting. Moving it downstream would start counting them.
```

Agreed on the reason for flagging it. The decision was made from the correct model of the pipeline but written up from the inverted one, which is exactly the kind of comment the next change gets justified from.

airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
darynaishchenko: Good catch — applied in `44165b99`. `filter_records` now skips the re-wrap when the item already is a `Record`:

```python
if self._cursor.should_be_synced(
    record
    if isinstance(record, Record)
    else Record(data=record, associated_slice=stream_slice, stream_name="")
)
```

Measured on both call sites — `(type, type(.data), stream_name)` as received by the cursor:

```
record input  -> [('Record', 'dict', 'items')]  | same object out: True
mapping input -> [('Record', 'dict', '')]       | type out: dict
```

So the retriever path now hands the cursor the real record with its real stream name, the selector path is unchanged, and both still yield the objects that were passed in.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
darynaishchenko: Confirmed, and leaving it as is. The warning sits in `create_simple_retriever`, so it fires once per stream construction rather than once per sync, and `streams()` does get called more than once in a process.

Deduplicating would mean holding state on the factory keyed by stream name, which is more machinery than a repeated log line justifies. Noting it here so the duplication is on record for anyone reading a support log.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -3422,6 +3426,31 @@ def _get_url(req: Requester) -> str:
...
+                f"Stream {name}: `is_client_side_incremental` is ignored when `is_data_feed` is set, "
+                "as a data feed already filters out the records that were synced during a previous sync."
+            )
```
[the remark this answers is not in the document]
darynaishchenko: Good point — the flag is not fully ignored, it still drives the `transform_before_filtering` default. Reworded in `44165b99`, slightly differently from the suggestion so the log line leads with the user-visible effect:

```python
LOGGER.warning(
    f"Stream {name}: `is_client_side_incremental` adds no record filtering when `is_data_feed` is set, "
    "as a data feed already filters on the cursor value. It still makes the record selector apply the "
    "transformations before the `record_filter` condition."
)
```

As it appears in a sync:

```
{"type":"LOG","log":{"level":"WARN","message":"Stream items: `is_client_side_incremental` adds no record filtering when `is_data_feed` is set, as a data feed already filters on the cursor value. It still makes the record selector apply the transformations before the `record_filter` condition."}}
```

unit_tests/sources/declarative/retrievers/test_data_feed_integration.py
```diff
@@ -0,0 +1,267 @@
...
+                {"id": "future", "updated_at": "2099-01-01T00:00:00Z"},
+                {"id": "fresh", "updated_at": "2022-06-01T00:00:00Z"},
+                {"id": "already_synced", "updated_at": "2020-06-01T00:00:00Z"},
+            ]
+        },
+    )
+
+    # the forward-dated record does not stop the pagination, it is only left out of the emitted records
+    assert pages_fetched == ["1"]
+    assert record_ids == ["fresh"]
```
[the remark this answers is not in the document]
darynaishchenko: Applied in `44165b99`, with one adjustment. Splitting the pages exactly as suggested made the test pass for the wrong reason: the paginator is `PageIncrement(page_size=4)`, so a two-record page 1 ends pagination on a short page rather than on the stop condition, and `pages_fetched` stayed `['1']`.

Page 1 is now full, so reaching page 2 is meaningful:

```python
"1": [
    {"id": "future", "updated_at": "2099-01-01T00:00:00Z"},
    {"id": "fresh_1", "updated_at": "2022-06-01T00:00:00Z"},
    {"id": "fresh_2", "updated_at": "2022-05-01T00:00:00Z"},
    {"id": "fresh_3", "updated_at": "2022-04-01T00:00:00Z"},
],
"2": [
    {"id": "already_synced", "updated_at": "2020-06-01T00:00:00Z"},
],
...
assert pages_fetched == ["1", "2"]
assert record_ids == ["fresh_1", "fresh_2", "fresh_3"]
```

Page 1 now holds nothing already synced, so the forward-dated record is the only thing on it that could have stopped the pagination — which is the property you wanted isolated.

tolik0:
/prerelease

> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/31496780606
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/31496796422

PR #1121, opened 2026-08-18
feat(low-code): let a quota status endpoint report that rate limiting is off

darynaishchenko:
## What

`RateLimitedMultipleTokenAuthenticator` seeds its per-token counters from `quota_status_url` on the first signed request, through a per-token `HttpClient` built with no `error_handler`. The default mapping therefore applies, and **any non-2xx fails the connection before the connector issues a single stream request**.

That is the right behaviour when the endpoint is expected to work. It is the wrong behaviour when the endpoint is optional — which it is on GitHub Enterprise Server, where HTTP API rate limiting is disabled by default and an instance in that state answers `GET /rate_limit` with:

```
404 Not Found
{"message": "Rate limiting is not enabled."}
```

Every command dies at seeding with a generic *"Resource not found"*, and the connector cannot work around it: `QuotaStatusSource` exposes only `url`, `http_method` and `request_headers`, and a stream's `error_handler` does not apply to the authenticator's own client.

Evidence for the GHES behaviour: [vermiculus/magithub#104](https://github.com/vermiculus/magithub/issues/104) quotes the response verbatim, and [actions/stale#1227](https://github.com/actions/stale/issues/1227) reports the same message still on GHES 3.15.2.

## How

`QuotaStatusSource` gains an opt-in `unavailable_status_codes: [int]`.

- Listed statuses are mapped to `ResponseAction.IGNORE` **for the quota request only**, so `send_request` returns the response instead of raising and `_fetch_quota_states` can interpret it. Any status not listed keeps failing exactly as before, and the handler is otherwise identical to the default one (`max_retries=5`, `max_time=600s`), so nothing else about the quota request changes.
- Every pool of a token that answered with a listed status is then seeded **untracked**. An untracked pool skips exhaustion waits and proactive throttling; requests are still signed, and the tokens are still rotated round-robin, because nothing about the quota endpoint being unavailable implies the other credentials should sit idle.
- One INFO summary per authenticator, emitted once every token has been seeded, so it can state the scope of the consequence rather than guess at it.

The field is deliberately narrow in three ways.

**It never excuses a missing quota path.** A path the response does not contain still fails the connection, whether or not the field is set. The two cases look similar and are not: an endpoint that answers with an error is telling you it does not track quotas, while an endpoint that answers with a body does track them, so a path absent from that body is a wrong path. Letting the opt-in cover both would mean a typo in `remaining_path` silently switches quota tracking off for the whole sync.

**A list, not a hardcoded 404.** GitHub documents `404 — Resource not found` as a possible response for this endpoint, but nowhere states that it *means* rate limiting is disabled; that mapping comes from captured responses in the wild. A proxy in front of an instance can also answer differently. The undocumented assumption belongs in the connector that makes it, not in the CDK for everyone.

**Only the authenticator's own bookkeeping is suppressed.** GHES exposes HTTP API and secondary rate limiting as independent toggles, so an instance can 404 the quota endpoint and still answer 403 when pushed. Those responses stay with the stream error handler, and a test pins that separation — with one honest limit: on an untracked pool the retry rotates onto the next token but still pays the backoff the response asks for, rather than the shortened one `has_alternative_token` grants a tracked pool. The shortcut exists because a tracked pool can prove the sending token is spent; an untracked pool cannot, and overriding the server's own reset header on a guess would burn every retry in under a second whenever the limit turns out to be shared across credentials.

### Also in here: `config_error` → `system_error` for a missing quota path

The failure above already existed; this PR only changes how it is classified and worded. `config_error` tells a user their configuration is wrong and they must fix it, but the quota paths come from the manifest, so there is nothing in their config to correct. It now raises `system_error`, and the message names the pool and which of `remaining`/`reset`/`limit` was not found instead of saying "an expected field":

```
Quota status response does not contain the configured remaining path for token quota "rest".
```

### Implementation notes

`tracked: bool` on `_QuotaState`, rather than a very large `remaining`. Six call sites read that state — `_acquire_call`, `_compute_budget_delay`, `_get_budget_reserve`, `_refresh_after_exhaustion`, `update_from_response`, `has_alternative_token` — and a sentinel would have to satisfy all of them by arithmetic accident. It would also need a far-future `reset_at`, which makes *both* branches of `update_from_response` unreachable and would silently discard response headers on deployments that do send them.

Round-robin on the untracked path, rather than the "use a token until it is spent, then move on" rule tracked pools follow. Every token hits the same `quota_status_url` and therefore gets the same status, so on the deployment this targets the untracked branch is the *only* one `_acquire_call` ever takes. Without advancing the active token there, one credential would serve the whole sync and the rest of a multi-token configuration would go unused — and the server may still enforce limits it declines to report.

Untracking is per token and holds for the rest of the sync: the exhaustion wait is the only thing that reseeds after startup, and it becomes unreachable as soon as one token is untracked, so the endpoint is not consulted again. If only some tokens are untracked the others stay tracked and keep throttling, but they are not refreshed either — once their counters are locally spent, traffic moves onto the untracked tokens. The field's description says all of this, because it is the argument for listing only the codes the endpoint uses to report that rate limiting is not enabled.

## Review guide

1. `airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py` — `_quota_status_error_handler`, `_untracked_states`, `_log_untracked_tokens`, the `_fetch_quota_states` branch, and the six `tracked` guards
2. `declarative_component_schema.yaml` — the new field's description
3. `unit_tests/sources/declarative/auth/test_rate_limited_multiple_token.py`

## Testing

`166 passed` in the authenticator suite; `795 passed` across `declarative/auth`, `declarative/parsers` and `streams/http`. `ruff check` clean, `ruff format` applied, `mypy` clean repo-wide.

New tests: untracked seeding never blocks or throttles · 404 **without** the opt-in still fails · a 500 still fails when only 404 is excused · `has_alternative_token` returns False when nothing is tracked · an untracked pool ignores response quota headers, asserted on the counters the guard protects rather than on the flag · a missing quota path raises `system_error` with or without the opt-in · the untracked summary is logged once and states its scope · the field is threaded through the factory and participates in the instance cache key.

Untracked tokens are pinned to share the load round-robin, an untracked pool is pinned never to be reseeded (so it cannot silently flip back to tracked), duplicate status codes are rejected by the schema, and two definitions listing the same codes in a different order are pinned to share one set of counters.

Four tests cover the mixed state — one token untracked, another healthy — which is the only state in which four of the `tracked` guards are reachable: an exhausted tracked token rotates onto the untracked one instead of waiting, an untracked peer suppresses the budget delay, `_refresh_after_exhaustion` issues no request, and an untracked sender reports no alternative token even though the other token has quota. Each guard was checked by deleting it and confirming the suite fails.

## Note for the reviewer

The generated model was updated **by hand** to add only the new field. Running `poe assemble` on `main` produces unrelated churn — it renames the public `OAuthScope` model to `Scope`/`OptionalScope` and reorders ~300 lines — which is pre-existing drift between the checked-in file and what the generator currently emits, and renaming an exported model is not something this PR should carry. Worth fixing separately.

Logging note: the one INFO summary is per authenticator, but `HttpClient` also logs its own `Ignoring response for 'GET' request to '…' with response code '404'` line once per token, because `ResponseAction.IGNORE` always logs. That is the low-level event and belongs per request; it cannot be suppressed from here.

## User Impact

None for any connector that does not set `unavailable_status_codes`. Behaviour is unchanged when the field is absent or empty.

`unavailable_status_codes` now carries `uniqueItems: true`, so a manifest repeating a status code is a validation error rather than a second, non-counter-sharing authenticator. No released manifest sets the field at all.

One classification change affects connectors that do not set the field: a quota status response missing a configured path already failed the connection, and still does, but now as a `system_error` rather than a `config_error`, with a message naming the pool and the field. No connector in the registry has hit this — the paths are always present on github.com — but a user who saw the old error would now see one that does not tell them to go fix their configuration.

- [x] YES 💚


tolik0 (commented):
Reviewed `07e1f7c8` against `main`. The problem is real and the shape of the fix is right: a GHES instance with HTTP API rate limiting disabled answers `GET /rate_limit` with `404 {"message": "Rate limiting is not enabled."}`, seeding goes through an `HttpClient` built with no `error_handler`, and the default mapping kills the connection before the connector issues a single stream request — with no connector-side workaround. Scoping `ResponseAction.IGNORE` to the authenticator's own client, for named statuses only, is the correct blast radius, and I agree with `tracked: bool` over a sentinel `remaining` for exactly the reason you give.

Two things I'd like to sort out before this goes in, plus some smaller notes inline.

**1. The diff and the PR description disagree on the central design decision.** The description says, twice, that a missing quota path still fails the connection whether or not the field is set, and that letting the opt-in cover both would mean "a typo in `remaining_path` silently switches quota tracking off for the whole sync". It also describes a `config_error` → `system_error` reclassification with a message naming the pool and field. Neither is in the code: `_extract_path` returns `_MISSING` as soon as `unavailable_status_codes` is non-empty, and the raise is still `config_error` with the generic "missing an expected field". The commit message and the tests both match the code, so I think the description is what moved on — but the description's argument against the shipped behaviour is the one I'd act on. Details inline.

**2. In the GHES case the untracked path makes multi-token configuration inert.** I went looking for the behaviour rather than reading the description, and measured two things on this branch:

- Three tokens, `404` seed, nine signed requests → `{'token token_1': 9}`. Every pool of every token is untracked (they all hit the same URL and get the same status), so `_acquire_call`'s early return is the only path it ever takes and `_active_token` is never advanced. Someone configuring five PATs on GHES to spread load gets one PAT doing everything, silently.
- `has_alternative_token` on a secondary-limit 403 → `False`. Your own narrowness argument is that GHES exposes primary and secondary limiting as independent toggles and that a 403 "remains the stream error handler's job" — but the handler's tool for that job is `_can_retry_on_another_token`, which is off for untracked pools. So on the target deployment a 403 sleeps out the full backoff on one credential instead of rotating across the pool.

Both look fixable cheaply and I've suggested how inline.

**On the guard coverage claim.** The description says "Each guard was checked by deleting it and confirming the suite fails". I couldn't reproduce that for four of the six — deleting them one at a time leaves `test_rate_limited_multiple_token.py` at 59/59 green. Table inline on `_compute_budget_delay`. I'd also note the four mixed-state tests the description describes ("one token untracked, another healthy") aren't in the diff; the new tests either untrack every token or untrack one pool of a single token.

**Non-blocking, and I think you're right to keep it out of scope:** hand-editing the generated model rather than carrying the `OAuthScope` → `Scope`/`OptionalScope` rename of a public model is the right call. That drift will keep blocking `poe assemble` for everyone though — worth filing as its own issue rather than leaving it in a PR body, wdyt?

Everything I claim to have measured is reproducible on this branch with the snippets in the inline comments. CI: only `destination-motherduck` is red, and that's an expired MotherDuck token in CI, not this change.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -342,6 +452,8 @@ def _extract_path(self, response_body: Mapping[str, Any], path: List[str]) -> An
         value: Any = response_body
         for key in path:
             if not isinstance(value, Mapping) or key not in value:
+                if self._unavailable_status_codes:
+                    return _MISSING
```
tolik0: This is the line I'd most like to talk about, because the PR description argues against it better than I can:

> **It never excuses a missing quota path.** A path the response does not contain still fails the connection, whether or not the field is set. […] Letting the opt-in cover both would mean a typo in `remaining_path` silently switches quota tracking off for the whole sync.

As written, a missing path on a healthy `200` produces an untracked pool as soon as `unavailable_status_codes` is non-empty. The two failures really are different, the way the description says: a status code is the endpoint telling you it doesn't track quotas, while a `200` with a body is it telling you it does — so a path absent from that body is a wrong path. Concretely, on github.com a typo in `remaining_path` fails loudly today; with `unavailable_status_codes: [404]` set for the GHES case, that same typo silently disables quota tracking for every sync on every deployment, announced by one INFO line.

The description's `system_error` reclassification isn't here either — the raise below is still `failure_type=FailureType.config_error` with "Quota status response is missing an expected field.", rather than the `Quota status response does not contain the configured remaining path for token quota "rest".` the body quotes.

My suggestion: drop this branch, keep the raise unconditional, and do the `config_error` → `system_error` change the description describes (the paths come from the manifest, so there's nothing in the user's config to fix). That also resolves two of my other comments — the `limit_path` asymmetry below and the durability question on `_refresh_after_exhaustion` — since both are downstream of `_MISSING` existing at all.

If per-pool tolerance is genuinely wanted for deployments that report some pools and not others, could it be its own field (`allow_missing_quota_paths` or similar)? Nothing about the name, type, or description of a list of HTTP statuses suggests it also relaxes JSON path validation, so as-is the coupling is undiscoverable from the manifest side.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -331,6 +432,15 @@ def _fetch_quota_states(self, token: str) -> dict[str, _QuotaState]:
                 if quota.limit_path
                 else remaining
             )
+            if remaining is _MISSING or reset is _MISSING or limit is _MISSING:
```
tolik0: Two things here.

**Declaring `limit_path` makes a pool more fragile than omitting it.** Just above, an absent `limit_path` falls back to `limit = remaining`. But a *declared* `limit_path` that the response doesn't contain untracks the whole pool through this check — even though `remaining` and `reset`, the two values every decision actually needs, were both present:

```python
# response has resources.core.{remaining,reset} but no resources.core.limit
a = auth(quota_status_unavailable_status_codes=[404])
a(prepared_request())
# -> rest tracked: False, rest remaining: 0
```

So adding `limit_path` to a manifest can only make things worse. Could the `_MISSING` check cover `remaining` and `reset` only, and fall back to `remaining` for the limit — matching the branch right above it? (Moot if you take the suggestion to revert `_extract_path` to raising, since then all three are strict again.)

**Small one:** `self._untracked_states()[quota.name]` builds a state for every pool and calls `ab_datetime_now()` to keep one of them. A single-pool `_untracked_state(quota)` that `_untracked_states` maps over would read better.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -210,10 +293,19 @@ def _acquire_call(self, quota: TokenQuota) -> str:
             with self._lock:
                 token = self._active_token
                 state = self._states[token][quota.name]
+                if not state.tracked:
+                    # Nothing to spend and nothing to wait for. Rotation still works on demand
+                    # (`update_from_response`, an explicit exhaustion signal), it just is not
+                    # driven by counters that do not exist.
+                    return token
```
tolik0, line 338: This early return fires before any rotation, and I think that's a bigger deal than the description treats it as. The description frames untracked-token traffic concentration as a mixed-state consequence ("once their counters are locally spent, traffic moves onto the untracked tokens"), but on a real GHES instance there is no mixed state: every token hits the same `quota_status_url` and gets the same `404`, so every pool of every token is untracked and this return is the *only* path `_acquire_call` ever takes. `_active_token` is then never advanced for the rest of the sync.

Measured on this branch:

```python
requests_mock.get(URL, status_code=404, json={"message": "Rate limiting is not enabled."})
a = auth(tokens=("token_1", "token_2", "token_3"),
         quota_status_unavailable_status_codes=[404])
used = [a(prepared_request()).headers["Authorization"] for _ in range(9)]
# Counter(used) -> {'token token_1': 9}
```

An operator who configures five PATs on GHES to spread load gets one PAT doing everything, with nothing in the log saying so. Nothing about the quota endpoint being unavailable implies the other credentials should go unused.

Would you consider advancing the token before returning here?

```python
if not state.tracked:
    self._active_token = next(self._tokens_iter)
    return token
```

One line, spreads load exactly as the round-robin did before this PR, and it changes nothing else about the untracked path. It's also a prerequisite for the `has_alternative_token` suggestion I left below — without it, a retry would re-sign with the same unrotated token.

Separately: the second half of the mechanism deserves a comment here. Once any token is untracked, `_refresh_after_exhaustion` is unreachable for that pool, so a tracked token's quota is never reused even after its window resets. That argument currently lives only in the PR body.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -438,10 +556,12 @@ def has_alternative_token(self, request: requests.PreparedRequest) -> bool:
         with self._lock:
             if not self._states or sender is None:
                 return False
-            if self._states[sender][quota.name].remaining > 0:
+            sender_state = self._states[sender][quota.name]
+            if not sender_state.tracked or sender_state.remaining > 0:
                 return False
             return any(
                 self._states[token][quota.name].remaining > 0
+                and self._states[token][quota.name].tracked
```
tolik0: This guard means an untracked pool can never report an alternative token, which I think undercuts the separation the PR is built on. The description's argument for narrowness is that GHES exposes HTTP API and secondary rate limiting as independent toggles, so "an instance can 404 the quota endpoint and still answer 403 when pushed", and that such responses "remain the stream error handler's job". But the handler's tool for that job is `_can_retry_on_another_token` → `has_alternative_token`, which shortens a rate-limit backoff to `TOKEN_ROTATION_BACKOFF` when another credential could serve the request.

Measured on this branch, two tokens, `404` seed:

```python
a = auth(tokens=("token_1", "token_2"), quota_status_unavailable_status_codes=[404])
r = prepared_request(); a(r)
a.has_alternative_token(r)   # -> False
```

So on the exact deployment this PR targets — primary limiting off, secondary limiting on — a 403 sleeps out the full computed backoff on one credential instead of rotating across the pool the user configured.

`test_untracked_pool_reports_no_alternative_token` asserts this as intended, and the stated reasoning is sound in isolation ("with no counters it cannot claim a token is spent"). But I don't think the conclusion follows: rotating to another untracked token can't be worse than sleeping, because there are no counters to protect on either side. Worth reconsidering — with the ordering caveat that it's only useful after the `_acquire_call` rotation fix, or the retry re-signs with the same token and a backoff becomes a tight loop.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -295,7 +389,9 @@ def _refresh_after_exhaustion(self, quota: TokenQuota) -> None:
         with self._refresh_lock:
             with self._lock:
                 still_exhausted = all(
-                    self._states[token][quota.name].remaining <= 0 for token in self._tokens
+                    self._states[token][quota.name].remaining <= 0
+                    and self._states[token][quota.name].tracked
```
tolik0, line 437: The description says:

> Untracking is per token and holds for the rest of the sync: the exhaustion wait is the only thing that reseeds after startup, and it becomes unreachable as soon as one token is untracked, so the endpoint is not consulted again.

That holds for per-*token* untracking, but not for the per-*pool* untracking that the `_MISSING` branch creates, because this check is per quota. If pool `rest` is tracked and exhausts across all tokens, it waits, calls `_seed_all_tokens()`, and that refetches every pool of every token — including the ones that were untracked.

Measured on this branch, one token, seed response omits `graphql` and reports `rest` already exhausted, second response includes `graphql`:

```
quota endpoint calls:            2
graphql tracked after reseed:    True
_untracked_logged still latched: True
```

So the endpoint *is* consulted again, `graphql` silently becomes tracked mid-sync, and nothing is logged — `_seed_all_tokens` resets `_budget_logged` but not `_untracked_logged`. The reverse transition happens too: a pool tracked at startup becomes untracked after a reseed if the endpoint stops reporting it.

I don't think either direction is obviously wrong — re-reading the endpoint is arguably better than latching a stale verdict — but the "holds for the rest of the sync" reasoning is load-bearing for the design, so it'd be good to decide this deliberately and log the transition either way. Disappears entirely if `_extract_path` goes back to raising.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -171,6 +206,54 @@ def __call__(self, request: requests.PreparedRequest) -> Any:
...
+        return {
+            quota.name: _QuotaState(remaining=0, reset_at=now, limit=0, tracked=False)
+            for quota in self._quotas
+        }
+
+    def _log_untracked_once(self, reason: str) -> None:
+        if self._untracked_logged:
+            return
+        self._untracked_logged = True
+        self._logger.info(
```
tolik0: This message states a per-pool cause and a global consequence in the same breath. Actual output when one pool is missing from the response:

```
Quota status endpoint did not report quota 'graphql'; that pool is untracked. Token quotas are
untracked: the connector will not wait for quota resets, throttle proactively, or rotate tokens
on exhaustion. Responses that report a rate limit are still handled by the stream's error handler.
```

An operator reading that can't tell whether `rest` is still throttling — it is. Could the consequence clause be phrased to match the scope of the reason it's given?

Also, being once per authenticator *lifetime* means only the first reason is ever reported: a token that answered with a listed status and a pool missing from another token's response are two different facts, and the second is swallowed. Once per distinct reason would fix both that and the silent mid-sync transition I mentioned on `_refresh_after_exhaustion`.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -171,6 +206,54 @@ def __call__(self, request: requests.PreparedRequest) -> Any:
...
+        return HttpStatusErrorHandler(
+            self._logger,
+            error_mapping={
+                **DEFAULT_ERROR_MAPPING,
+                **{
+                    status_code: ErrorResolution(
+                        response_action=ResponseAction.IGNORE,
+                        failure_type=FailureType.transient_error,
+                        error_message=(
+                            "Quota status endpoint reports that rate limiting is unavailable; "
```
tolik0: Heads up that the logging note in the PR description is inverted here. `http_client.py` does:

```python
self._logger.info(error_resolution.error_message or log_message)
```

so because this sets `error_message`, the generic `Ignoring response for 'GET' request to '…' with response code '404'` line the description says can't be suppressed is in fact never emitted — `HttpClient` logs *this* string instead, once per token, and `_log_untracked_once` then logs a second, near-identical sentence.

Net effect is two INFO lines saying the same thing per authenticator, not one summary plus a low-level per-request line. Worth picking one: drop `error_message` here to get the generic per-request line alongside the summary, or keep it and drop the summary.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -56,11 +64,33 @@ def is_response_aware(self) -> bool:
         )
 
 
+class _Missing:
+    """Marks a quota path the response did not contain. Distinct from `None`, which a server is
+    free to send as a legitimate value."""
+
+    def __repr__(self) -> str:
+        return "<missing>"
```
tolik0: `__repr__` is never called — nothing formats the sentinel into a message, it's only ever compared with `is`. Fine to keep as debugging affordance, just flagging it as currently dead.

The docstring premise is also a bit stronger than the code: "Distinct from `None`, which a server is free to send as a legitimate value" is true of the sentinel, but a literal `null` at `remaining_path` still reaches `int(None)` in `_fetch_quota_states` and raises an unhandled `TypeError`. That's pre-existing and not this PR's job to fix — I'd just not have the comment imply it's handled.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
```diff
@@ -262,6 +354,8 @@ def _acquire_call(self, quota: TokenQuota) -> str:
     def _compute_budget_delay(self, quota: TokenQuota) -> Optional[float]:
         """Compute the proactive throttling delay. Must be called while holding the lock."""
         states = [self._states[token][quota.name] for token in self._tokens]
+        if any(not state.tracked for state in states):
```
tolik0, line 395: I tried to reproduce the description's "Each guard was checked by deleting it and confirming the suite fails" and couldn't, for four of the six. Deleting one guard at a time and running `unit_tests/sources/declarative/auth/test_rate_limited_multiple_token.py` (baseline: 59 passed):

| Guard | Suite after deleting it |
| --- | --- |
| `and … .tracked` in `_acquire_call`'s `all(...)` | 59 passed |
| `if any(not state.tracked …): return None` (this one) | 59 passed |
| `and … .tracked` in `_refresh_after_exhaustion`'s `still_exhausted` | 59 passed |
| `and … .tracked` on the peer tokens in `has_alternative_token` | 59 passed |
| `if not state.tracked: return` in `update_from_response` | suite hangs (>150s) |
| `if not state.tracked: return token` in `_acquire_call` | suite hangs (>150s) |

Only the two early returns are exercised, and both by hang rather than assertion — removing the `_acquire_call` one leaves the rotation `else` branch spinning forever when every token is untracked.

Worth noting the four uncovered guards are the four that only fire in a mixed per-token state, which is close to unreachable in production: all tokens hit the same `quota_status_url`, so a `404` for one credential and a `200` for another needs a proxy behaving per-credential. So either the mixed-state tests are worth adding, or the guards that only defend an unreachable state are worth dropping — I'd just rather not leave untested branches behind a claim they were verified.

`has_alternative_token`'s peer guard is a specific case: `test_untracked_pool_reports_no_alternative_token` trips the *sender* guard on the first condition, so the `any(...)` clause it looks like it's testing is never reached.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -444,6 +444,14 @@ definitions:
         type: object
         additionalProperties:
           type: string
+      unavailable_status_codes:
+        title: Unavailable Status Codes
+        description: Status codes from the quota status endpoint that mean quota tracking is unavailable rather than broken, such as a self-hosted deployment with rate limiting turned off. Every pool is then treated as untracked, so the authenticator stops waiting for quota resets, stops throttling proactively and stops rotating on exhaustion, while still signing requests. Rate limiting reported by ordinary responses is unaffected, so an error handler that retries 429 or 403 keeps working. Any status not listed still fails the connection, so list only the codes the endpoint uses to report that rate limiting is not enabled.
```
tolik0: Two additions I'd suggest for this description.

**The missing-path consequence isn't mentioned.** As implemented, setting this field also makes a missing quota path produce an untracked pool instead of failing the connection. A connector author reading this description learns nothing about that, so the footgun is invisible from the manifest side. (My preference is to remove that coupling in the code instead — see the comment on `_extract_path` — but if it stays, it needs to be documented here.)

**Nothing stops an auth status from being listed.** `unavailable_status_codes: [403]` or `[401]` would turn a revoked token into "quota tracking unavailable" rather than a credentials failure. The last sentence already says to list only codes the endpoint uses to report that rate limiting is off, which is the right guidance — one explicit "don't list authentication or authorization statuses" would make it harder to get wrong, or reject them at parse time.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -4693,6 +4693,11 @@ def create_rate_limited_multiple_token_authenticator(
             key: str(InterpolatedString.create(value, parameters={}).eval(config))
             for key, value in (model.quota_status_source.request_headers or {}).items()
         }
+        # Normalized the same way as the quota specs above, so an omitted field and an explicit
+        # `[]` key identically and keep sharing one set of counters.
+        quota_status_unavailable_status_codes = sorted(
```
tolik0, line 4700: `sorted(...)` preserves duplicates, so `[404]` and `[404, 404]` produce different `cache_key`s and therefore two authenticators that don't share quota counters, even though both become `{404}` at runtime. `sorted(set(...))` here, plus `uniqueItems: true` on the schema field, would make the two spellings equivalent on both sides. (CodeRabbit flagged this one too.)

Also a typo in the comment just above: "Normalized the same way as the quota specs above" → "Normalize".

unit_tests/sources/declarative/auth/test_rate_limited_multiple_token.py
```diff
@@ -997,3 +997,167 @@ def test_factory_does_not_share_instances_across_differing_header_config():
...
+    tolerant = build(definition)
+    assert tolerant._unavailable_status_codes == {404}
+
+    without = {
+        **definition,
+        "quota_status_source": {"type": "QuotaStatusSource", "url": QUOTA_STATUS_URL},
+    }
+    strict = build(without)
+    assert strict._unavailable_status_codes == set()
+    assert strict is not tolerant, "the cache key must include the new field"
```
tolik0, line 1357: The new tests here all either untrack every token (a `404` seeding response) or untrack one pool of a single token. The description says:

> Four tests cover the mixed state — one token untracked, another healthy — which is the only state in which four of the `tracked` guards are reachable

but I can't find a test that constructs one untracked token alongside a healthy one, which is why the four guards in my table on `_compute_budget_delay` survive deletion. Did those get lost in a rebase?

Two more that would be worth having, both derived from behaviour I measured rather than from the guards: the token actually used when every pool is untracked (currently `token_1` for all requests), and what happens to a previously-untracked pool after `_seed_all_tokens` runs again (currently it can flip back to tracked, silently).

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Done in d501e4f6 — dropped the `_MISSING` branch entirely and made the raise unconditional, so `unavailable_status_codes` gates status codes only.

Your framing is the one I ended up using: an endpoint answering with an error is telling you it doesn't track quotas, an endpoint answering with a body is telling you it does, so a path absent from that body is a wrong path.

The reclassification is in too:

```
type    : FailureType.system_error
message : Quota status response does not contain the configured remaining path for token quota "rest".
internal: Quota status response did not contain the remaining path ['resources','core','remaning'] configured for quota 'rest'
```

The test is parameterized over both configurations and asserts the failure type, so the classification is pinned rather than incidental.

On giving per-pool tolerance its own field — not adding one. The trigger this PR is for is a 404, not a partial 200, so there's no evidence behind the per-pool case; it can land later with its own field and its own reason if a deployment turns up that needs it. As you predicted, this also closed the `limit_path` asymmetry and the `_refresh_after_exhaustion` durability question.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Both resolved by dropping `_MISSING`. A declared-but-absent `limit_path` now raises like the other two rather than untracking a pool whose `remaining` and `reset` were both present, so the asymmetry where adding `limit_path` could only make things worse is gone.

The `_untracked_states()[quota.name]` nit went with it — that was the only per-pool call site, so the method has one caller now and returns exactly what it wants. I kept the part of your point that survives, as a docstring note: `remaining=0` there is load-bearing rather than incidental, because it is what makes every `remaining > 0` test in the class correct for an untracked pool without also consulting `tracked`. That let me delete a redundant guard — see the thread on `_compute_budget_delay`.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Applied in d501e4f6, your line as written. This is the one I'd missed: I had measured `['token_1', 'token_1', 'token_1']` in my own pass and read it as "nothing to spend", which walked straight past the fact that the tokens are still there to spread load over.

Three tokens, 404 seed, nine signed requests:

```
before: {'token token_1': 9}
after : {'token_1': 3, 'token_2': 3, 'token_3': 3}
```

`test_untracked_tokens_still_share_the_load` pins the round-robin, and I checked the new line is load-bearing by deleting it and confirming the suite fails.

Also added the comment you asked for on the second half of the mechanism — once any token is untracked the exhaustion branch below can never fire, so `_refresh_after_exhaustion` is unreachable and a tracked token's quota is never picked up again even after its window resets. That argument was only in the PR body; it's at the branch now.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Keeping it `False`, but your comment did change something: the docstring's stated reason stopped being true the moment the rotation fix landed. It claimed `_acquire_call` returns the active token unchanged, so promising rotation would retry on the same credential — with round-robin, the retry does rotate.

Rewrote it with the real reason, which is narrower and is a trade-off rather than a clear win. What the guard withholds is only the *skipped wait*. The backoff it would skip is computed from the server's own reset or `Retry-After` header, and an untracked pool has no counters with which to argue the rejection was about that credential specifically. If the limit turns out to be shared across credentials — the same user's PATs, or a per-IP secondary limit — returning `True` burns all five retries in about half a second and fails a request that waiting would have completed.

So: rotate, but still pay the backoff the response asks for. I also dropped the claim from the PR body that these responses "remain the stream error handler's job" full stop, because you're right that it promised a rotation shortcut it doesn't deliver.

If you think the per-token case is common enough to be worth the shared-limit risk, happy to revisit. The middle option would be a larger substituted backoff instead of `TOKEN_ROTATION_BACKOFF` rather than a boolean flip, but that constant is in `HttpClient` from #1117 and would want its own tests.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Gone with `_MISSING`, as you expected. Untracking is per token and all-or-nothing now, so `still_exhausted` can never be true while any token is untracked, which makes the reseed unreachable and both transitions you measured impossible.

Verified rather than assumed — spent the tracked token and kept going well past the point a reseed would have happened:

```
further quota-status requests: 0
```

`test_untracked_tokens_are_never_reseeded` pins it. "Holds for the rest of the sync" is load-bearing for the design, so I'd rather it fail a test than a re-read.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Fixed in d501e4f6, and your diagnosis pointed at the right cause — the call site couldn't know the scope. Moved the log to `_seed_all_tokens`, the first point that sees every token, so it reports what is actually true:

```
Quota status endpoint reports that rate limiting is unavailable for 1 of 2 tokens. Those tokens are
used without quota tracking. The other 1 keep proactive throttling until their counters are locally
spent, after which traffic moves onto the untracked tokens: the connector no longer waits for a
quota reset, so it never refreshes them.
```

The once-per-lifetime swallowing is moot now: with `_MISSING` gone there is one untracking reason and it is identical across tokens, so nothing is hidden. I checked the flag can't latch wrongly either — it is only set when something is actually untracked, so an all-tracked first seeding followed by a 404 on a later reseed still reports.

Worth flagging that my first attempt at this overclaimed in a subtler way: it said resets and throttling "stay in effect for the other N", which measurement contradicted, because the tracked tokens are never refreshed once any token is untracked. The wording above is the corrected one.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: You're right, and the PR body had it backwards — thanks. Dropped `error_message` from the `IGNORE` resolution, so `HttpClient` logs its own per-request line and the summary is the only sentence that states the consequence.

Three tokens on a 404 now give three `Ignoring response for 'GET' request to '…' with response code '404'` lines plus one summary, where before it was the custom string once per token plus a near-identical summary.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: First half is moot — the sentinel is deleted, so the dead `__repr__` and the docstring premise go with it.

Confirmed the `int(None)` part is still live, and agreed it isn't this PR's job:

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Left alone rather than folded in, and no comment now implies it's handled.

airbyte_cdk/sources/declarative/auth/rate_limited_multiple_token.py
darynaishchenko: Reproduced your table and got the same result. I'd found the same thing in my own pass and had the fix sitting in my working tree unpushed, which is why the diff you read didn't have it — sorry for the wasted effort.

Took the "add the tests" branch rather than dropping the guards, on the reachability point: a mixed state needs one token's fetch to fail where another's succeeds, which a proxy can do but so can a transient failure on a single fetch, so I'd rather the guards stay. A `_mixed_authenticator` helper seeds one token from a 404 and the other from a healthy body, and four tests pin the four outcomes.

Your `has_alternative_token` observation was exactly right, and I resolved it the other way: the sender guard trips first, so the `any()` clause was never reached — and it turned out that clause couldn't change the answer at all, since an untracked pool always holds `remaining=0` and `remaining > 0` already excludes it. Deleted it, and wrote the invariant down on `_untracked_states` instead of duplicating it as a guard.

All seven guards now fail the suite when deleted, five of them within 0.13s. Two still detect by non-termination rather than assertion and I don't have a clean fix for that: removing either early return leaves the rotate branch spinning, with no sleep and no iterator to assert against. The four fast ones got there by giving the `time.sleep` patches a raising `side_effect` instead of `assert_not_called`.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
darynaishchenko: First half is moot — the coupling is removed, so there's nothing to document.

Second half added as guidance rather than a parse-time rejection:

> Do not list authentication or authorization statuses, since a 401 or 403 from a revoked credential would then be read as quota tracking being unavailable rather than as a credentials failure.

Guidance rather than rejection because rejecting outright would block a legitimate case: an endpoint can answer 403 when the credential lacks the scope to read quota, and untracking is arguably right there. I confirmed the footgun is real first — `unavailable_status_codes: [403]` plus a 403 response does untrack.

The description also gained the mixed-token consequence and the "holds for the rest of the sync" clause, both from other threads here.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
darynaishchenko: Both applied. `sorted(set(...))`, plus `uniqueItems: true` on the field and `unique_items=True` in the generated model — following `HttpResponseFilter.http_codes`, which is the schema's existing precedent for a status-code list, so this isn't a new pattern.

Confirmed the finding first: `[404]` and `[404, 404]` did build two authenticators that didn't share counters. Duplicates are a `ValidationError` at the schema layer now, `sorted(set(...))` still canonicalizes the programmatic path, and a test pins that two definitions listing the same codes in a different order share one instance.

Typo fixed — and the same one in the comment just above it that this was copied from.

unit_tests/sources/declarative/auth/test_rate_limited_multiple_token.py
darynaishchenko: Not a rebase — they were in my working tree unpushed when you reviewed. They're in d501e4f6 now, with both of the tests you suggested.

The first is `test_untracked_tokens_still_share_the_load`, which asserts round-robin rather than `token_1` for everything, since that behaviour changed on your `_acquire_call` comment.

The second turned out to have a different answer than when you measured it: with `_MISSING` gone a previously-untracked pool can't flip back, because the reseed is unreachable while any token is untracked. So `test_untracked_tokens_are_never_reseeded` pins the impossibility rather than the transition.

darynaishchenko:
Thanks @tolik0 — pushed `d501e4f6`, which addresses everything from your review. Replies are on each thread; the three items from the review body that weren't inline:

**The description/diff disagreement was my fault, and you diagnosed it correctly.** The description had moved on: I'd found the `_extract_path` and guard-coverage problems in my own pass, written the fixes and the description, and then not pushed the code. So you reviewed the old diff against the new prose. The description now matches `d501e4f6`, and your instinct — that the description's argument against the shipped behaviour was the one to act on — is what the code does.

**Multi-token configuration being inert was a genuine miss on my side.** I'd measured `{'token token_1': 9}` myself and read it as "nothing to spend, so nothing to rotate", which skipped the point that load spreading is the reason someone configures five PATs in the first place. Your one-liner is in, verified before and after. The `has_alternative_token` half I've kept as-is and explained on that thread — the short version is that with rotation fixed, what the guard withholds is only the skipped wait, and skipping a backoff the server asked for is a guess that fails requests outright when the limit is shared across credentials. Open to revisiting if you think the per-token case dominates.

**`poe assemble` drift** — agreed it shouldn't live in a PR body. It needs regenerating plus a CI check that fails when the committed model diverges; I'll take it out of this PR's description and into its own issue rather than leaving it as tribal knowledge.

@coderabbitai both outside-diff findings were valid and are fixed in `d501e4f6`: `sorted(set(...))` with `uniqueItems: true` on the schema field, and the `_MISSING` branch removed so a successful response omitting a configured path raises `system_error` with the pool and field named. One deviation from the suggestion — the error message does not include the token, because the token is the credential.

Also verified as still live but deliberately left out of scope, per @tolik0's note on it: a literal `null` at `remaining_path` reaches `int(None)` and raises an unhandled `TypeError`. Pre-existing from #1117.

`166` passed in the authenticator suite, `795` across `declarative/auth`, `declarative/parsers` and `streams/http`; ruff, format and mypy clean. `destination-motherduck` is the expired MotherDuck token, as you said — it fails identically on #1119.

tolik0 (approved):

PR #1124, opened 2026-08-20
fix(declarative): detect gzip payloads in GzipParser instead of trusting headers

Airbyte-Support:
## Summary

Requested by Syed Khadeer (Airbyte support) off Zendesk ticket 18622: a customer's Connector Builder **Test** read of a stream that downloads a gzipped CSV fails with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1` while a real sync of the same stream succeeds. `0x1f 0x8b` is the gzip magic — compressed bytes are reaching the UTF-8 CSV parser.

Resolves https://github.com/airbytehq/oncall/issues/13362:
- https://github.com/airbytehq/oncall/issues/13362

Cause: `create_gzip_decoder` builds two *different* decoders depending on mode, and neither actually looks at the payload.

```python
if self._emit_connector_builder_messages:      # Builder test read
    return CompositeRawDecoder(gzip_parser.inner_parser, False)   # gunzip step dropped entirely
return CompositeRawDecoder.by_headers(          # real sync
    [({"Content-Encoding", "Content-Type"}, _compressed_response_types, gzip_parser)],
    stream_response=True,
    fallback_parser=gzip_parser.inner_parser,   # no gunzip unless a header matched
)
```

The Builder branch assumed `requests` had already decompressed `response.content`, which only holds for *transport-level* `Content-Encoding: gzip`. When the body is a gzip *payload* (e.g. an S3 `*.csv.gz` download served as `application/gzip` or `binary/octet-stream`, as Apple App Store Connect analytics report segments are), nothing gunzips it in Builder mode. Measured on `main` with a single `GzipDecoder(CsvDecoder)`, against an undecoded raw stream (`decode_content=False`, matching how `requests` constructs `response.raw` in production sync mode):

| payload / headers | sync | Builder test read |
| --- | --- | --- |
| gzip, `Content-Type: application/gzip` | OK | `UnicodeDecodeError` 0x8b |
| gzip, `Content-Type: binary/octet-stream` | `UnicodeDecodeError` 0x8b | `UnicodeDecodeError` 0x8b |
| gzip, `Content-Encoding: gzip` | OK | OK |
| gzip×2, `Content-Encoding: gzip` + `Content-Type: application/gzip` | `UnicodeDecodeError` 0x8b (only one layer stripped) | `UnicodeDecodeError` 0x8b |

(An earlier revision of this table claimed the `Content-Encoding: gzip` sync path raised `BadGzipFile`; that measurement was taken with `decode_content=True`, which pre-decompressed the fixture body — the production raw path succeeded. Corrected per review.)

That asymmetry is why manifests in the wild carry double-nested `GzipDecoder(GzipDecoder(CsvDecoder))` workarounds (`source-amazon-ads` ships one next to a `# TODO Fix me`) — and why the single- vs. double-nesting workarounds are mutually exclusive depending on which header the server returns. Previously reported in airbytehq/oncall#7739, #11173, #11809 and airbytehq/airbyte#56988.

Fix, in two parts:

1. Make `GzipParser` self-detecting — the behavior its own docstring already promised (*"If the data is not gzipped, reset the pointer and pass the data to the inner parser as is"*) but never implemented. `GzipParser.parse` reads the 2-byte header (looping, since `response.raw.read(2)` may short-read), gunzips when it is `\x1f\x8b`, and otherwise hands the inner parser the untouched stream. Because the stream may be non-seekable (`response.raw` in sync mode), the consumed header is restored by wrapping it in a private `_PrefixedStream` (`io.RawIOBase`) inside an `io.BufferedReader`, which keeps it usable by both `gzip.GzipFile` and `CsvParser`'s `TextIOWrapper`.
2. Use it in both modes of `create_gzip_decoder`, treating the HTTP transport layer and the gzip representation as separate layers — a response can validly carry *both* `Content-Encoding: gzip` and `Content-Type: application/gzip`, in which case the sync raw stream holds two gzip layers:

```python
if self._emit_connector_builder_messages:   # requests strips the transport layer via response.content
    return CompositeRawDecoder(gzip_parser, False)

transport_gzip_parser = GzipParser(inner_parser=gzip_parser)   # strips the Content-Encoding layer, then the payload layer
return CompositeRawDecoder.by_headers(
    [
        ({"Content-Encoding"}, {"gzip"}, transport_gzip_parser),
        ({"Content-Type"}, _compressed_response_types, gzip_parser),
    ],
    stream_response=True,
    fallback_parser=gzip_parser,
)
```

A single `GzipDecoder` now works for every combination in the table above, in both modes (pass-through makes the nested parser safe when only one layer is present). Existing double-nested manifests keep working: the inner `GzipParser` sees already-decompressed data and passes it through instead of raising `BadGzipFile`.

## Test plan

- New `GzipParser` unit tests in `unit_tests/sources/declarative/decoders/test_composite_decoder.py`: gzipped payload and non-gzip pass-through, each over both a seekable and a truly non-seekable stream (`seek()`/`tell()` raise `io.UnsupportedOperation`), a stream that returns one byte per `read()`, empty payload, and nested `GzipParser(GzipParser(CsvParser))` over a singly-gzipped body.
- New `create_gzip_decoder` tests in `unit_tests/sources/declarative/parsers/test_model_to_component_factory.py` covering `emit_connector_builder_messages` `True` and `False` against each header shape above, with `preload_content=False, decode_content=False` fixtures matching the production raw path — including the double-layer `Content-Encoding` + `Content-Type` regression case. Builder mode had no coverage at all before.
- `poetry run pytest unit_tests/sources/declarative/decoders/ unit_tests/sources/declarative/parsers/test_model_to_component_factory.py`, `poetry run ruff format --check . && poetry run ruff check .`, `poetry run mypy --config-file mypy.ini airbyte_cdk`.

Link to Devin session: https://app.devin.ai/sessions/3b5024f8f9bf4de88ec1abad6fb688fb

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2783,15 +2783,12 @@ class ModelToComponentFactory:
         gzip_parser: GzipParser = ModelToComponentFactory._get_parser(model, config)  # type: ignore  # based on the model, we know this will be a GzipParser
 
         if self._emit_connector_builder_messages:
-            # This is very surprising but if the response is not streamed,
-            # CompositeRawDecoder calls response.content and the requests library actually uncompress the data as opposed to response.raw,
-            # which uses urllib3 directly and does not uncompress the data.
-            return CompositeRawDecoder(gzip_parser.inner_parser, False)
+            return CompositeRawDecoder(gzip_parser, False)
 
         return CompositeRawDecoder.by_headers(
             [({"Content-Encoding", "Content-Type"}, _compressed_response_types, gzip_parser)],
             stream_response=True,
-            fallback_parser=gzip_parser.inner_parser,
+            fallback_parser=gzip_parser,
         )
 
     @staticmethod
--- a/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -2785,8 +2785,12 @@ class ModelToComponentFactory:
         if self._emit_connector_builder_messages:
             return CompositeRawDecoder(gzip_parser, False)
 
+        transport_gzip_parser = GzipParser(inner_parser=gzip_parser)
         return CompositeRawDecoder.by_headers(
-            [({"Content-Encoding", "Content-Type"}, _compressed_response_types, gzip_parser)],
+            [
+                ({"Content-Encoding"}, {"gzip"}, transport_gzip_parser),
+                ({"Content-Type"}, _compressed_response_types, gzip_parser),
+            ],
             stream_response=True,
             fallback_parser=gzip_parser,
         )
--- a/unit_tests/sources/declarative/decoders/test_composite_decoder.py
+++ b/unit_tests/sources/declarative/decoders/test_composite_decoder.py
@@ -73,6 +73,42 @@ def generate_csv(
     return csv_data.encode(encoding)
 
 
+class NonSeekableBytesIO(BytesIO):
+    def seekable(self) -> bool:
+        return False
+
+
+def test_gzip_parser_decompresses_gzip_payload():
+    parser = GzipParser(inner_parser=CsvParser())
+
+    assert list(parser.parse(BytesIO(compress_with_gzip("date,units\n2026-08-01,42\n")))) == [
+        {"date": "2026-08-01", "units": "42"}
+    ]
+
+
+@pytest.mark.parametrize("stream_class", [BytesIO, NonSeekableBytesIO])
+def test_gzip_parser_passes_through_non_gzip_payload(stream_class):
+    parser = GzipParser(inner_parser=CsvParser())
+
+    assert list(parser.parse(stream_class(b"date,units\n2026-08-01,42\n"))) == [
+        {"date": "2026-08-01", "units": "42"}
+    ]
+
+
+def test_gzip_parser_handles_empty_payload():
+    parser = GzipParser(inner_parser=CsvParser())
+
+    assert list(parser.parse(BytesIO())) == []
+
+
+def test_nested_gzip_parser_decompresses_single_gzip_payload():
+    parser = GzipParser(inner_parser=GzipParser(inner_parser=CsvParser()))
+
+    assert list(parser.parse(BytesIO(compress_with_gzip("date,units\n2026-08-01,42\n")))) == [
+        {"date": "2026-08-01", "units": "42"}
+    ]
+
+
 @pytest.mark.parametrize("encoding", ["utf-8", "utf", "iso-8859-1"])
 def test_composite_raw_decoder_gzip_csv_parser(requests_mock, encoding: str):
     requests_mock.register_uri(
--- a/unit_tests/sources/declarative/decoders/test_composite_decoder.py
+++ b/unit_tests/sources/declarative/decoders/test_composite_decoder.py
@@ -78,12 +78,27 @@ class NonSeekableBytesIO(BytesIO):
         return False
 
 
+class OneByteAtATimeBytesIO(BytesIO):
+    def read(self, size: int = -1) -> bytes:
+        if size > 0:
+            size = 1
+        return super().read(size)
+
+
 def test_gzip_parser_decompresses_gzip_payload():
     parser = GzipParser(inner_parser=CsvParser())
 
     assert list(parser.parse(BytesIO(compress_with_gzip("date,units\n2026-08-01,42\n")))) == [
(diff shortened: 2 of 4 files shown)


unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -261,6 +270,42 @@ def test_create_check_stream():
...
+):
+    csv_data = b"date,units\n2026-08-01,42\n"
+    response = requests.Response()
+    response.status_code = 200
+    response.headers.update(headers)
+    response.raw = HTTPResponse(
+        body=io.BytesIO(gzip.compress(csv_data)),
+        headers=headers,
+        status=200,
+        preload_content=False,
```
pnilan, line 295: > 🤖 *This PR review comment was generated by an AI Agent.*

`urllib3.HTTPResponse` defaults `decode_content=True`, but Requests creates `response.raw` with `decode_content=False`. As written, the sync test receives an already-decompressed body and does not reproduce the production raw-response path.

Please configure the fixture explicitly:

```suggestion
        preload_content=False,
        decode_content=False,
```

Builder mode will still decompress `Content-Encoding: gzip` through `response.content`, while sync mode will leave the gzip bytes for `GzipParser`. Please also update the PR/issue matrix: a singly gzipped `Content-Encoding: gzip` response should already have succeeded in the old sync path because the selected `GzipParser` received compressed raw bytes.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2783,15 +2783,12 @@ def create_gzip_decoder(
...
-            # CompositeRawDecoder calls response.content and the requests library actually uncompress the data as opposed to response.raw,
-            # which uses urllib3 directly and does not uncompress the data.
-            return CompositeRawDecoder(gzip_parser.inner_parser, False)
+            return CompositeRawDecoder(gzip_parser, False)
 
         return CompositeRawDecoder.by_headers(
             [({"Content-Encoding", "Content-Type"}, _compressed_response_types, gzip_parser)],
             stream_response=True,
-            fallback_parser=gzip_parser.inner_parser,
+            fallback_parser=gzip_parser,
```
pnilan, line 2795: > 🤖 *This PR review comment was generated by an AI Agent.*

A response may validly have both `Content-Encoding: gzip` and `Content-Type: application/gzip`: the former is the HTTP encoding layer and the latter describes the gzip representation. Builder removes the HTTP layer through `response.content`, but sync reads undecoded `response.raw`, so the current single `GzipParser` removes only one of the two layers.

Please select a nested transparent parser for `Content-Encoding` and retain the model parser for the representation/fallback paths:

```python
transport_gzip_parser = GzipParser(inner_parser=gzip_parser)

return CompositeRawDecoder.by_headers(
    [
        ({"Content-Encoding"}, {"gzip"}, transport_gzip_parser),
        ({"Content-Type"}, _compressed_response_types, gzip_parser),
    ],
    stream_response=True,
    fallback_parser=gzip_parser,
)
```

Add a Builder/sync regression case with both headers, `decode_content=False`, and `body=io.BytesIO(gzip.compress(gzip.compress(csv_data)))`. That proves both modes remove the HTTP and representation layers and return the same CSV records.

unit_tests/sources/declarative/decoders/test_composite_decoder.py
```diff
@@ -73,6 +73,57 @@ def generate_csv(
     return csv_data.encode(encoding)
 
 
+class NonSeekableBytesIO(BytesIO):
+    def seekable(self) -> bool:
+        return False
```
pnilan, line 79: > 🤖 *This PR review comment was generated by an AI Agent.*

Overriding `seekable()` alone does not make this stream non-seekable: inherited `seek()` and `tell()` still work. Please import `UnsupportedOperation` from `io` and make those operations fail:

```python
class NonSeekableBytesIO(BytesIO):
    def seekable(self) -> bool:
        return False

    def seek(self, pos: int, whence: int = 0) -> int:
        raise UnsupportedOperation("seek")

    def tell(self) -> int:
        raise UnsupportedOperation("tell")
```

Then parameterize `test_gzip_parser_decompresses_gzip_payload` over `[BytesIO, NonSeekableBytesIO]`, just like the pass-through test. This directly proves that both compressed and uncompressed inputs work without seeking.

pnilan (changes requested):
> 🤖 *This PR review was generated by an AI Agent.*

Requesting changes for two gzip parity/test-validity gaps and one test-hardening issue. The previously discussed `typing-extensions` concern is intentionally omitted because Pydantic already guarantees a sufficient version.

pnilan (approved):

PR #1126, opened 2026-08-20
fix(http): decide token rotation before a backoff strategy can refuse to wait, and resolve the wait cap at construction

darynaishchenko:
## What

Two changes to when `HttpClient` rotates credentials instead of waiting out a rate limit. The first is the bug in the title; the second rides on the same reorder and is a deliberate widening, not a fix — both are described below and both are pinned by tests.

1. **A capped strategy can no longer pre-empt rotation.** `HttpClient` can shorten a rate-limit wait when the authenticator holds a spare credential, and a backoff strategy can refuse a wait that exceeds a cap. Combining the two loses the first: a capped strategy ends the stream while a fully-quota'd credential sits idle.
2. **Rotation now also covers rate limits that produce no wait at all**, where the old order fell through to the default exponential retry. That changes which exception is raised and therefore which retry curve runs: `RateLimitBackoffException` → `UserDefinedBackoffException`, `backoff.expo` (1s, 2s, 4s, …) → `backoff.constant` at `0.1 + 1`s, and `max_time` — not passed to the rate-limit handler — now bounds the loop. On a credential with quota that is the better curve, but it is a semantic change to behaviour that shipped in 7.26.0, which is why the title carries it.
3. **Both capped strategies resolve `max_waiting_time_in_seconds` in `__post_init__`.** Follows from 1: if the strategies can be skipped, a cap that cannot be evaluated can go unreported. See below.

The two features shipped two days apart — the rotation shortcut in #1117 (7.26.0), the cap in #1123 (7.27.0) — and nothing covered them together.

## Why

`_handle_error_resolution` asks the strategies for a wait, then rewrites that wait when the response was rate-limited and another credential is available:

```python
user_defined_backoff_time = None
for backoff_strategy in self._backoff_strategies:      # step 1
    backoff_time = backoff_strategy.backoff_time(...)  #   <- a capped strategy raises here
    ...

if (                                                   # step 2
    user_defined_backoff_time
    and error_resolution.response_action == ResponseAction.RATE_LIMITED
    and self._can_retry_on_another_token(request)
):
    user_defined_backoff_time = self.TOKEN_ROTATION_BACKOFF
```

Step 2 is expressed as a modification of step 1's result, so it depends on step 1 returning. `WaitUntilTimeFromHeaderBackoffStrategy._capped` raises `AirbyteTracedException` instead of returning when the computed wait exceeds `max_waiting_time_in_seconds`, so control leaves the function at step 1 and rotation is never considered.

Measured on `source-github` (airbytehq/airbyte#81428), two tokens, one rejected with `X-RateLimit-Reset` an hour out:

| wait budget | before | after |
|---|---|---|
| 120 min (no cap hit) | rotates, retry in 0.1 s | unchanged |
| 30 min (cap hit) | **stream stops**, spare token idle | rotates, retry in 0.1 s |

The connector-side consequence is worse than a wasted wait: a user who *lowers* the wait budget — asking for less waiting — gets a failed sync where the connector would previously have carried on. A bound on waiting silently became a bound on the sync.

## How

Ask the rotation question first, and only fall through to the strategies when the answer is no.

When another credential can serve the retry, the strategies are not consulted at all. That is deliberate rather than an optimisation: the wait they compute is derived from the rejected credential's response headers, and the retry will not use that credential, so the number describes a window nobody is waiting for. Skipping them also means a strategy that refuses to wait cannot pre-empt the decision.

When there is no spare credential, the loop runs exactly as before and a cap still ends the stream — that path is untouched, and a test pins it.

Non-rate-limit resolutions (`RETRY`, `REFRESH_TOKEN_THEN_RETRY`) are unaffected: the rotation branch is gated on `RATE_LIMITED`, as it was before.

Two consequences of not calling the strategies, both deliberate and both now pinned by tests:

- **A rate limit that produces no backoff at all rotates too**, where the old order fell through to the default exponential retry. `has_alternative_token` only answers True when the sending credential is tracked and spent, so this is the same retry on a credential with quota rather than a blind shortcut.
- **A `max_waiting_time_in_seconds` the manifest got wrong is no longer reported from this path.** `evaluate_max_waiting_time` raises `system_error` for a cap it cannot resolve, and that check lives inside the strategy — so with rotation available a broken cap now stays quiet until the first rate limit that finds no spare credential. Resolving the cap once in `__post_init__` closes it on every path, and is the better fix, but it moves *when* the field raises and breaks nine tests that assert the current timing — that is #1123's contract, not this PR's ordering, so it is recorded in the comment and left for its own change.

## Resolving the cap at construction

Deciding rotation first means a strategy may never be asked for a wait — and the check that a manifest's `max_waiting_time_in_seconds` can be evaluated at all lived inside that call. Measured on an earlier commit of this branch, with `max_waiting_time_in_seconds: "{{ config['not_in_spec'] }}"` and a 429 an hour out:

| authenticator | before | after |
|---|---|---|
| spare credential available | `200` — the bound is not applied and nothing says so | raises at construction |
| no spare credential | `system_error` | raises at construction |

Two outcomes from one manifest, decided by the quota state of an unrelated credential. Caching the resolved value or storing the error to re-raise would not have helped: the rotation path never reaches `backoff_time()`, so detection had to leave it. `config` is a dataclass field and this cap interpolates over `config` alone, so the value is knowable as soon as the component exists, and `ModelToComponentFactory` already passes the real config.

**This is a contract change, and it is the reason the fix was initially deferred.** Nine tests pinned the old timing and now assert construction instead — five in `test_wait_time_from_header.py`, four in `test_wait_until_time_from_header.py`. What they assert otherwise is unchanged: still an `AirbyteTracedException` with `system_error`, still never a raw jinja or float error, still the connector's fault rather than the user's. Only the moment moves, from "the first retryable error that reaches a strategy" to "when the strategy is built".

**Fleet impact is nil today**, which is what makes this the cheap moment to do it — every user of the field resolves cleanly: `source-github` ([#81428](https://github.com/airbytehq/airbyte/pull/81428)) guards with `config.get(...) is not none`, `source-klaviyo` passes a float, `source-granola` hardcodes `60`. Not a breaking change: no field added, removed or renamed, no default changed, no manifest edit required of anyone. It changes when an already-invalid manifest reports itself.

The lazy check stays alongside it, so a strategy constructed directly in Python behaves as before.

## Tests

Four added to `unit_tests/sources/streams/http/test_http_client.py`, plus one guard on the schema itself:

| Test | Asserts |
|---|---|
| `test_rotation_is_preferred_over_a_strategy_that_refuses_to_wait` | with a spare credential, the request succeeds after a sub-second retry |
| `test_a_refusing_strategy_still_ends_the_stream_without_a_spare_credential` | with no spare credential, the cap still raises |
| `test_rotation_is_preferred_over_the_real_capped_strategy` | the same, driving the real `WaitUntilTimeFromHeaderBackoffStrategy` with a cap below the wait the response asks for |
| `test_rotation_also_covers_a_rate_limit_with_no_computed_wait` | a strategy returning `None` is not consulted at all on the rotation path |

The first fails on `main` (`AirbyteTracedException` instead of a 200), verified by reverting the source change with the tests in place. The third exists because the stub tests would survive the cap being changed to return instead of raise, which is the failure mode that produced this bug.

Both `max_waiting_time_in_seconds` schema descriptions, both class docstrings, `WaitUntilTimeFromHeader._capped` and `WaitTimeFromHeader`'s inline cap said "stop the stream" and "0 means never wait" without qualification. Neither holds once rotation preempts the strategy, so each now says the bound applies only when waiting is the only way forward.

`unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py` is new and unrelated to the behaviour change: it loads the shipped schema through the same helper every declarative source uses. Nothing asserted that before, and the first version of the description edit above put a `": "` inside a plain YAML scalar, which stopped the whole schema parsing and failed every low-code source. That is fixed here by quoting both scalars — the text is unchanged, so the generated models still match — and the new test catches the class of mistake in milliseconds.

## Checks

- `unit_tests/sources/streams/http`, `.../declarative/requesters/error_handlers`, `.../declarative/auth`, plus the new schema guard: 763 passed
- `unit_tests/sources/declarative`: the `test_concurrent_declarative_source` failures are pre-existing — same names, on an `origin/main` worktree (a `requests_cache` sqlite lock on this machine)
- `ruff format --check`, `ruff check`: clean
- `mypy` on all five changed modules (`http_client.py`, `error_handlers/backoff_strategy.py`, `max_waiting_time_helper.py` and both backoff strategies): clean
- `unit_tests/.../error_handlers`: 147 passed after the nine timing tests were moved to construction
- The three rotation tests still fail against `main`'s `http_client.py` after being refactored onto the `_rate_limited_client` helper

## Why it is worth taking now

The interaction was found while reviewing airbytehq/airbyte#81428, which needs it. At that PR's head the cap is applied on **both** wait paths with the user's configured budget — `max_waiting_time_in_seconds: "{{ (config['max_waiting_time'] ... else 120) * 60 + 1 }}"` on `WaitTimeFromHeader` and `WaitUntilTimeFromHeader` — and the manifest comment says so, pointing at a prerelease of this PR and at `test_short_wait_budget_does_not_cost_token_rotation` as the end-to-end proof. Without this fix, a budget shorter than the distance to the reset ends the stream with a healthy second token idle (measured on 7.28.0: a 30-minute budget against a 60-minute reset).

Still worth a maintainer's read on the ordering choice — specifically, whether skipping the strategies entirely on the rotation path is right, or whether the wait should still be computed and discarded. The trade-off is written up under "Two consequences" above: skipping means a `max_waiting_time_in_seconds` the manifest got wrong is not reported from this path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate-limit handling by prioritizing credential rotation before applying backoff strategies.
  * Retries using an alternative credential now bypass configured wait limits and use the standard rotation delay.
  * Clarified maximum-wait behavior for header-based retry strategies, including boundary conditions.

* **Documentation**
  * Updated retry and backoff guidance to accurately describe credential rotation, wait caps, and strategy behavior.
  * Expanded validation guidance for declarative schema documentation.

### DIFF
--- a/airbyte_cdk/sources/streams/http/http_client.py
+++ b/airbyte_cdk/sources/streams/http/http_client.py
@@ -594,16 +594,31 @@ class HttpClient:
             # raising (`max_waiting_time_in_seconds`), which would otherwise end the stream
             # before rotation was ever considered. Rotating is strictly the better outcome
             # there: it is the same retry, seconds from now, on a credential with quota.
+            #
+            # Two consequences of not calling the strategies, both deliberate. A rate limit
+            # that yields no backoff at all now rotates too, rather than falling through to
+            # the default exponential retry -- on a rotating credential that is the better
+            # behaviour, and `has_alternative_token` only answers True when the sending
+            # credential is tracked and spent. And a `max_waiting_time_in_seconds` the manifest
+            # got wrong -- one that cannot be evaluated -- is no longer reported here, since
+            # that error is raised from inside the strategy; it still surfaces on the first
+            # rate limit that finds no spare credential. Resolving the cap once at construction
+            # would close that gap on every path, but it moves when `max_waiting_time_in_seconds`
+            # fails, so it belongs in its own change rather than in this reorder.
             rotate_instead_of_waiting = (
                 error_resolution.response_action == ResponseAction.RATE_LIMITED
                 and self._can_retry_on_another_token(request)
             )
 
             if rotate_instead_of_waiting:
+                # Says that a wait was skipped without the number, which is no longer computed,
+                # and names the cap explicitly: a connector that configured one gets no other
+                # signal that the retry went ahead without consulting it.
                 self._logger.info(
-                    f"Rate limited on the current credential; retrying in "
-                    f"{self.TOKEN_ROTATION_BACKOFF}s with another one instead of waiting for "
-                    f"the rate limit to reset."
+                    "Rate limited on the current credential; retrying in "
+                    f"{self.TOKEN_ROTATION_BACKOFF}s with another one instead of waiting for the "
+                    "rate limit to reset. Any configured backoff, including a wait cap, is not "
+                    "evaluated for this retry."
                 )
                 user_defined_backoff_time = self.TOKEN_ROTATION_BACKOFF
             else:
--- a/airbyte_cdk/sources/streams/http/http_client.py
+++ b/airbyte_cdk/sources/streams/http/http_client.py
@@ -588,29 +588,33 @@ class HttpClient:
             ResponseAction.REFRESH_TOKEN_THEN_RETRY,
         ):
             user_defined_backoff_time = None
-            for backoff_strategy in self._backoff_strategies:
-                backoff_time = backoff_strategy.backoff_time(
-                    response_or_exception=response if response is not None else exc,
-                    attempt_count=self._request_attempt_count[request],
-                )
-                if backoff_time:
-                    user_defined_backoff_time = backoff_time
-                    break
-
-            if (
-                user_defined_backoff_time
-                and error_resolution.response_action == ResponseAction.RATE_LIMITED
+            # Asked before the strategies, not after. The backoff they compute describes the
+            # credential the server just rejected, so when another credential can serve the
+            # retry that wait is irrelevant -- and a strategy is allowed to refuse a wait by
+            # raising (`max_waiting_time_in_seconds`), which would otherwise end the stream
+            # before rotation was ever considered. Rotating is strictly the better outcome
+            # there: it is the same retry, seconds from now, on a credential with quota.
+            rotate_instead_of_waiting = (
+                error_resolution.response_action == ResponseAction.RATE_LIMITED
                 and self._can_retry_on_another_token(request)
-            ):
-                # The backoff was derived from this response's headers, which only describe the
-                # credential that was rejected. Another one has quota, and the retry re-signs the
-                # request, so waiting out this window would idle for nothing.
+            )
+
+            if rotate_instead_of_waiting:
                 self._logger.info(
                     f"Rate limited on the current credential; retrying in "
-                    f"{self.TOKEN_ROTATION_BACKOFF}s with another one instead of waiting "
-                    f"{user_defined_backoff_time:.0f}s for the rate limit to reset."
+                    f"{self.TOKEN_ROTATION_BACKOFF}s with another one instead of waiting for "
+                    f"the rate limit to reset."
                 )
                 user_defined_backoff_time = self.TOKEN_ROTATION_BACKOFF
+            else:
+                for backoff_strategy in self._backoff_strategies:
+                    backoff_time = backoff_strategy.backoff_time(
+                        response_or_exception=response if response is not None else exc,
+                        attempt_count=self._request_attempt_count[request],
+                    )
+                    if backoff_time:
+                        user_defined_backoff_time = backoff_time
+                        break
 
             error_message = (
                 error_resolution.error_message
--- a/airbyte_cdk/sources/streams/http/http_client.py
+++ b/airbyte_cdk/sources/streams/http/http_client.py
@@ -598,13 +598,13 @@ class HttpClient:
             # Two consequences of not calling the strategies, both deliberate. A rate limit
             # that yields no backoff at all now rotates too, rather than falling through to
             # the default exponential retry -- on a rotating credential that is the better
-            # behaviour, and `has_alternative_token` only answers True when the sending
-            # credential is tracked and spent. And a `max_waiting_time_in_seconds` the manifest
-            # got wrong -- one that cannot be evaluated -- is no longer reported here, since
-            # that error is raised from inside the strategy; it still surfaces on the first
-            # rate limit that finds no spare credential. Resolving the cap once at construction
-            # would close that gap on every path, but it moves when `max_waiting_time_in_seconds`
-            # fails, so it belongs in its own change rather than in this reorder.
+            # behaviour, and `has_alternative_token` only answers True when the retry will
+            # rotate -- the CDK's own authenticator narrows that further, to a sending credential
+            # that is tracked and spent. And a `max_waiting_time_in_seconds` the manifest got
+            # wrong -- one that cannot be evaluated -- is not reported from here, since that
+            # error is raised from inside the strategy. Both capped strategies therefore resolve
+            # the field once in `__post_init__` too, so a manifest mistake fails at startup
+            # rather than waiting for a rate limit that finds no spare credential.
             rotate_instead_of_waiting = (
                 error_resolution.response_action == ResponseAction.RATE_LIMITED
                 and self._can_retry_on_another_token(request)
--- a/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
+++ b/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
@@ -1,0 +1,40 @@
+#
+# Copyright (c) 2026 Airbyte, Inc., all rights reserved.
+#
+
+"""The shipped manifest schema has to parse, and nothing else asserted that it does.
+
+`_get_declarative_component_schema()` runs before any low-code source is built, so a malformed
+`declarative_component_schema.yaml` fails every declarative connector at startup. It is also easy
+to break from a documentation edit alone: the descriptions are plain YAML scalars, so a `": "` in
+prose ends the scalar and the file stops being a mapping. That happened in this file's history and
+surfaced as 194 unrelated test failures rather than as one obvious error.
+"""
+
+from airbyte_cdk.sources.declarative.concurrent_declarative_source import (
+    _get_declarative_component_schema,
+)
+
+
+def test_the_shipped_component_schema_parses():
+    schema = _get_declarative_component_schema()
+
+    assert schema["title"] == "DeclarativeSource"
+    assert "HttpRequester" in schema["definitions"]
+
+
+def test_every_description_survives_the_yaml_round_trip():
+    """A description that ends early still parses — it just silently loses its tail, or turns the
+    rest of the sentence into a key. Reading them all back catches the truncation too."""
+    schema = _get_declarative_component_schema()
+
+    for name, definition in schema["definitions"].items():
+        for field_name, field_schema in (definition.get("properties") or {}).items():
+            description = (
+                field_schema.get("description") if isinstance(field_schema, dict) else None
+            )
+            if description is not None:
+                assert isinstance(description, str), (
+                    f"{name}.{field_name} description is not a string"
+                )
+                assert description.strip(), f"{name}.{field_name} has an empty description"
--- a/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
+++ b/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
@@ -57,6 +57,9 @@ def test_no_plain_scalar_carries_yaml_punctuation():
     Every inline value is checked, not only `description`, because the hazard belongs to the
     scalar style rather than to the field: a `title`, an `error_message` or a value written
     under a sequence item all break the same way.
+
+    Line-anchored, so a plain scalar wrapped onto a continuation line is only checked on its first
+    line. The file has none today, and quoting a long value is the fix either way.
     """
     # The same bytes the loader reads, fetched the same way, so this cannot drift from what ships.
     raw_schema = pkgutil.get_data(
--- a/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
+++ b/unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
@@ -9,11 +9,29 @@
 to break from a documentation edit alone: the descriptions are plain YAML scalars, so a `": "` in
 prose ends the scalar and the file stops being a mapping. That happened in this file's history and
 surfaced as 194 unrelated test failures rather than as one obvious error.
+
+Two guards, because the two ways prose breaks a plain scalar look nothing alike. A `": "` is loud:
+the file stops parsing, so loading it is enough to catch it. A `" #"` is silent: YAML reads the
+rest of the line as a comment, the value keeps parsing as a shorter string, and nothing in the
+parsed tree records that a sentence went missing. Only the source text knows, so the second guard
+reads that instead of the loaded schema.
 """
+
+import pkgutil
+import re
 
 from airbyte_cdk.sources.declarative.concurrent_declarative_source import (
     _get_declarative_component_schema,
 )
+
+# `key: value` with the value on the same line. List items are excluded on purpose: `- key: value`
+# is a nested mapping, where a colon is structure rather than a truncated sentence.
+_INLINE_VALUE = re.compile(r"^\s*[\w$\"-]+:\s+(\S.*?)\s*$")
+
+# A plain scalar is one not opened with a quote, a block indicator or a flow collection, and it is
+# the only style these two characters are dangerous in: inside `'...'` or a `|` block they are
+# text like any other.
+_QUOTED_BLOCK_OR_FLOW = "'\"|>&*[{"
 
 
 def test_the_shipped_component_schema_parses():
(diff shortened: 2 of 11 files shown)


darynaishchenko:
@coderabbitai review

darynaishchenko:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/32473963287
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/32473979054

darynaishchenko:
@coderabbitai review

unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
```diff
@@ -0,0 +1,73 @@
...
+import pkgutil
+import re
+
+from airbyte_cdk.sources.declarative.concurrent_declarative_source import (
+    _get_declarative_component_schema,
+)
+
+# `key: value` with the value on the same line. List items are excluded on purpose: `- key: value`
+# is a nested mapping, where a colon is structure rather than a truncated sentence.
+_INLINE_VALUE = re.compile(r"^\s*[\w$\"-]+:\s+(\S.*?)\s*$")
```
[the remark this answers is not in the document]
darynaishchenko: Good catch — taken, in d7fcbab0.

The comment justifying the exclusion was wrong, and so was the number in it. Those 125 hits came from reading a whole list item as one scalar (`- inject_into: header` → value `inject_into: header`). Capturing only the value, the way the `key: value` form already did, reports **none** on the untouched file, and covers 106 more lines. Your regex is in there verbatim.

It leaves one gap, so the pattern became a pair of scans:

```python
_SCANS = (
    (re.compile(r"^\s*(?:-\s+)?[\w$\"-]+:\s+(\S.*?)\s*$"), (": ", " #")),
    (re.compile(r"^\s*-\s+(\S.*?)\s*$"), (" #",)),
)
```

The second one is for a sequence item that is a bare scalar (`- some example text`), which neither form matched before. It checks `" #"` only: a `": "` there does not truncate anything, it turns the item into a one-key mapping, which parses — and telling that apart from the nested mappings the file legitimately writes that way is not possible from the text alone.

Mutation-tested, six probes against the shipped schema:

| mutation | guard |
|---|---|
| untouched file | passes |
| unquoted `": "` in a description | fails |
| `" #"` in a description | fails |
| `" #"` in a `title` | fails |
| `" #"` in an unquoted `- inject_into: header` | fails — new, from your suggestion |
| `" #"` in a bare sequence item | fails — new |

One probe misfired first time on `- "$ref": "#/definitions/CheckStream"`: the value is quoted, so it is skipped, correctly. Re-ran it against an unquoted line. Also added a `break` after the first hazard on a line, since the two scans overlap on `- key: value` and would otherwise report it twice.

unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
```diff
@@ -0,0 +1,87 @@
...
+    offenders = []
+
+    for number, line in enumerate(raw_schema.decode().splitlines(), start=1):
+        for pattern, hazards in _SCANS:
+            match = pattern.match(line)
+            if match is None:
+                continue
+            value = match.group(1)
+            if value[0] in _QUOTED_BLOCK_OR_FLOW:
+                continue
```
tolik0: **P2 — a correctly quoted value under a sequence-item key is reported as an offender.**

These two `continue`s mean different things and only one of them is right. The first (`match is None`) correctly means "this scan did not match, try the next". This one means "this scan matched and the value is safely quoted" — which settles the *line*. But `continue` sends it on to scan 2, whose `^\s*-\s+(\S.*?)\s*$` re-captures the whole `key: 'value'` as if it were a bare sequence scalar. The quotes are then just characters inside the captured text, so a ` #` between them gets reported.

Measured by inserting one line into the packaged schema and running the guard:

| inserted line | expected | actual |
|---|---|---|
| `  - x_probe: 'safely quoted # hash'` | pass | **FAILED** — `line 75: x_probe: 'safely quoted # hash'` |
| `  - x_probe: "safely quoted # hash"` | pass | **FAILED** |
| `  x_probe: 'safely quoted # hash'` | pass | passed |
| `  - 'safely quoted # hash'` | pass | passed |

Row 3 is the tell: the identical value is safe under a plain key and an offender under a sequence-item key. It fails closed rather than open, and there are zero occurrences in the schema today — but the assertion message then reads "wrap each one in single quotes" at an author who already did.

```suggestion
                break
```

I verified the fix: with `break`, all four rows above behave as expected and every true positive is still caught — `- key: value with a # hash`, `- key: value with a: colon`, `- bare text with a # hash`, `key: text with a # hash` and `key: text with a: colon` all still fail the guard, and the untouched file still passes both tests. Nothing is lost, because a line whose value is quoted is settled by whichever scan read it first.

airbyte_cdk/sources/streams/http/http_client.py
```diff
@@ -588,29 +588,48 @@ def _handle_error_resolution(
...
+            # there: it is the same retry, seconds from now, on a credential with quota.
+            #
+            # Two consequences of not calling the strategies, both deliberate. A rate limit
+            # that yields no backoff at all now rotates too, rather than falling through to
+            # the default exponential retry -- on a rotating credential that is the better
+            # behaviour, and `has_alternative_token` only answers True when the sending
+            # credential is tracked and spent. And a `max_waiting_time_in_seconds` the manifest
+            # got wrong -- one that cannot be evaluated -- is no longer reported here, since
+            # that error is raised from inside the strategy; it still surfaces on the first
+            # rate limit that finds no spare credential. Resolving the cap once at construction
```
tolik0: **P3 — worth an issue rather than only a comment.**

The sequencing argument is right: moving cap resolution to `__post_init__` changes *when* `max_waiting_time_in_seconds` raises and breaks the nine tests pinning the current timing, which is #1123's contract rather than this PR's.

But the consequence stands — a manifest with a cap expression that cannot be evaluated now stays quiet until the first rate limit that finds no spare credential, and the only record of that anywhere is this comment. This is the third place it has been deferred (here, #1123, and my review of #1123). Could we open a CDK issue and link it from this line, so it survives the next person reading this block? wdyt

unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
```diff
@@ -0,0 +1,87 @@
+#
+# Copyright (c) 2026 Airbyte, Inc., all rights reserved.
+#
+
+"""The shipped manifest schema has to parse, and nothing else asserted that it does.
```
tolik0: **P3 — this line and the one nine lines down disagree.**

"nothing else asserted that it does" versus "surfaced as 194 unrelated test failures". 194 tests failing *is* an assertion that the schema parses — a bad one, at the wrong layer, with unreadable output. That is a stronger case for this file than the one the docstring makes, and it is the case the rest of the paragraph already argues: the guard replaces a 194-failure blast with one named failure in milliseconds.

Also worth weighing, though I would not hold the PR for it: the file is acknowledged scope creep, and this PR changes a retry path no connector exercises yet — so it is a plausible fast-revert candidate, and a revert would take the schema guard with it.

tolik0 (approved):
**Approve.** Non-breaking, and the core change is correct — this is the fix for the interaction I flagged as a design note on #1123 ("the cap raise pre-empts `_can_retry_on_another_token`, so a multi-token pool dies on a window it never needed to wait out").

Round 2 review of `1635fb86` (round 1 was `c8a491e4`).

### Verified rather than read

- **The bug reproduces.** In a worktree at the PR head with *only* `http_client.py` reverted to `origin/main`, three of the four rotation tests fail — `test_rotation_is_preferred_over_a_strategy_that_refuses_to_wait`, `..._over_the_real_capped_strategy`, `..._also_covers_a_rate_limit_with_no_computed_wait` — and all five pass with the PR's file restored. Re-ran after the `_rate_limited_client` refactor: identical, so the −65 lines cost no coverage.
- **Blast radius is nil for anyone who has not opted in.** `_can_retry_on_another_token` short-circuits on `isinstance(authenticator, TokenRotatingAuthenticator)`, so every other connector reaches the `else:` branch, which is the pre-PR loop verbatim. And `grep -rl RateLimitedMultipleTokenAuthenticator airbyte-integrations/connectors/` in the monorepo returns nothing — the only consumer is the still-open #81428.
- **No hammer-loop risk from `TOKEN_ROTATION_BACKOFF = 0.1`.** `has_alternative_token` answers `True` only when the *sending* token is tracked with `remaining == 0` and another has quota; in exactly that state `_acquire_call`'s final `else` advances `_tokens_iter`, and `_send` re-signs on every retry. The prompt retry is on a different credential by construction.
- **Generated model is in sync** with the YAML by exact string containment, and no comparison operator changed — `>=` in `_capped` is already `main`'s behaviour from #1123, so the description edits are documentation catching up.
- **The widened YAML guard works.** I mutation-tested it independently of the commit message: five hazard forms each fail the guard (`": "` and `" #"` on a plain key, both on a `- key: value`, and `" #"` on a bare sequence item), the documented `": "`-on-a-bare-item carve-out is correctly not flagged, and the untouched file passes.
- **Local suites:** 757 passed / 5 failed + 1 error across `streams/http`, `declarative/auth`, `error_handlers` and the new schema test — the same six names on an `origin/main` worktree, the known macOS `requests_cache` artefact. Byte-identical to round 1.

### Round 2 closed four round-1 items, and one further than asked

The base-contract docstring, the widened title and rewritten body, the trimmed Builder tooltip, and the `_rate_limited_client` reuse are all done. The `should_backoff()` the old docstring referenced genuinely does not exist anywhere in the package — good incidental catch. And taking CodeRabbit's regex, then finding the gap it left, then mutation-testing the result is the right way to handle a bot finding.

### Open — none blocking

Three inline comments: one P2 false positive introduced by the guard widening (one-keyword fix, verified), and two P3s.

### Caveat on CI

All five `Pytest` jobs were **pending** when I looked. They were green on round 1's head and the round-2 delta is docstrings plus test-only code that I ran locally, but CI has not confirmed this head. `Check: source-shopify` is also pending; `destination-motherduck` is red on 6 of 6 open PRs, so pre-existing; `source-google-drive` is skipped, not passed. None of the six `Test Connectors` exercises rotation or a wait cap, so that matrix proves only that the untouched branch did not regress — the real validation is a `/prerelease` pinned onto #81428, whose `test_short_wait_budget_does_not_cost_token_rotation` is an `xfail(strict)` tripwire.

unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
darynaishchenko: Fixed in a44e143e — `break`, exactly as suggested.

You were right about the mechanism and about the count. My first pass at reproducing it showed *three* false positives, but one was my own probe: I inserted the sequence item at a point where the YAML did not allow one, so the `parses` test failed rather than the guard. Separating the two failure modes gives your table exactly — two false positives, both under a sequence-item key, and `- 'safely quoted # hash'` correctly passing.

Verified after the change, inserting each probe line into the packaged schema at a valid sequence position:

| inserted line | before | after |
|---|---|---|
| `- x_probe: 'safely quoted # hash'` | FAILED | passes |
| `- x_probe: "safely quoted # hash"` | FAILED | passes |
| `- 'safely quoted # hash'` | passes | passes |
| `- x_probe: value with a # hash` | FAILED | FAILED |
| `- x_probe: value with a: colon` | FAILED | FAILED |
| `- bare text with a # hash` | FAILED | FAILED |
| untouched file | passes | passes |

Plus the true positives on existing lines, all still caught by the guard rather than by the parse test: `" #"` on a description, on a title, on an unquoted `- inject_into: header`, and on a bare `- type`.

Both exits are `break` now, with a comment recording why the distinction is not what it looks like — the first `continue` means "this scan did not match", the second meant "this line is settled", and only `break` says that.

unit_tests/sources/declarative/test_declarative_component_schema_is_loadable.py
darynaishchenko: Fixed in a44e143e. You are right that 194 failing tests *are* an assertion — a bad one, at the wrong layer, with unreadable output — and that this is the stronger case for the file.

The docstring now opens with "the shipped manifest schema has to parse, and only 194 unrelated failures said so", and the paragraph closes with what the guards actually add: not coverage, but a named failure in milliseconds.

On the revert point: taken, and noted rather than acted on. The file is still scope creep, and a revert of this PR would still take the guard with it. I left it here because the YAML break it guards against was caused by this PR, so the guard and the hazard have the same blast radius.

airbyte_cdk/sources/streams/http/http_client.py
darynaishchenko: Rather than file it, I fixed it — 69441bda, and the PR title and description now carry it as a third change.

Your framing is what changed my mind: three deferrals with nothing outside a comment is not a deferral, it is a leak. And the sequencing argument cuts the other way once you look at the exposure. Every user of the field in the monorepo resolves cleanly today — source-github guards with `config.get(...) is not none`, source-klaviyo passes a float, source-granola hardcodes `60` — so the change is a fleet no-op right now, and #1126 exists precisely to make the source-github pattern (a user-configurable budget interpolated into the cap) worth adopting. Interpolated values are the ones that can fail to resolve, so the cost of this fix only rises from here.

Both strategies now resolve the cap in `__post_init__`. Caching or storing-to-re-raise would not have worked: the rotation path never reaches `backoff_time()`, so detection had to leave that function, and construction is the only earlier point with everything it needs — `config` is a dataclass field, the cap interpolates over `config` alone, and the factory already passes the real config.

Measured, same probe as before:

| authenticator | before | after |
|---|---|---|
| spare credential available | `200` — bound not applied, nothing said so | raises at construction |
| no spare credential | `system_error` | raises at construction |

The nine tests assert construction now, which is the contract change you and I both flagged as #1123's. Everything else they pin is unchanged — `AirbyteTracedException`, `system_error`, never a raw jinja or float error, connector's fault rather than the user's. Only the moment moves. `error_handlers` is 147 passing.

I also updated the three places whose prose asserted the old timing, so nothing now contradicts the code: `evaluate_max_waiting_time`'s docstring, the deferral comment on this very block, and — while there — the `has_alternative_token` claim you did not raise but which overstated the protocol's guarantee. It now says the protocol promises only that the retry will rotate, and that the CDK's own authenticator narrows that to a tracked, spent sender.

Happy to split this into its own PR if you would rather keep the retry-ordering change reviewable on its own — say the word and I will lift the commit out.

PR #1135, opened 2026-08-26
feat(low-code): fetch complete nested lists when RecordExpander input is truncated, warn when unrecoverable

ZaneHyattAB:
## Overview

<table><tr><td>

**:point_right: TL;DR:** When an API embeds only the first page of a nested list inside a parent record, the low-code record expander can now fetch the rest of the list from its own endpoint, and when no such endpoint exists it logs a warning instead of silently dropping the missing items.

Specifically, this adds two optional fields to `RecordExpander`: `truncation_indicator_path` and `truncated_list_retriever` (`SimpleRetriever | CustomRetriever`), plus an optional `message_repository` for Connector Builder visibility.

</td></tr></table>

Pairs with (connector consumer, stays in draft until a CDK release contains this change):
- https://github.com/airbytehq/airbyte/pull/85087

Requested by Zane Hyatt (@ZaneHyattAB); investigation in https://github.com/airbytehq/oncall/issues/12975.

```yaml
record_expander:
  type: RecordExpander
  expand_records_from_field: [data, object, lines, data]
  truncation_indicator_path: [data, object, lines, has_more]   # NEW
  truncated_list_retriever:                                    # NEW, optional
    type: SimpleRetriever
    requester:
      path: invoices/{{ stream_slice['parent_record']['data']['object']['id'] }}/lines
      use_cache: true
    paginator: { ... }
```

## Changes

- `RecordExpander`: when `truncation_indicator_path` is truthy and `truncated_list_retriever` is set, the retriever is invoked with the parent record as `stream_slice['parent_record']` and its records (all pages via its own paginator) replace the embedded ones. Fetched items are handled like embedded ones: mappings get parent context, scalars are yielded as-is or wrapped as `{"value": item, "original_record": ...}` under `remain_original_record`.
- `RecordExpander`: truthy indicator with **no** retriever expands the embedded items as before and logs one WARNING per stream instance naming the expansion path, indicator path, embedded count and (when an integer `total_count` sibling exists) the expected total. No payload values. The sync never fails and no records are dropped.
- `RecordExpander`: a **separate** once-per-stream WARNING when a configured retriever returns fewer records than `total_count` (points at a missing `paginator` / wrong endpoint). Not emitted in Connector Builder test reads when the nested retriever's paginator is capped: `create_record_expander` sets `suppress_incomplete_fetch_warning` when the constructed `SimpleRetriever.paginator` is a `PaginatorTestReadDecorator` (a nested retriever without a `paginator` gets `NoPagination`, which is never capped, so its shortfall is real and still warns); for a `CustomRetriever` the paginator is opaque, so the flag falls back to `bool(self._limit_pages_fetched_per_slice)`. The flag is inert outside test reads, and the no-retriever warning is unaffected.
- `RecordExpander`: warnings are emitted through one channel: as `AirbyteLogMessage(level=WARN)` via the `MessageRepository` when one is wired in (the factory always supplies one, so they reach the platform log stream and Connector Builder `StreamRead.logs`), otherwise through the stdlib `airbyte` logger. Never both, so a warning is not printed twice. The no-retriever warning is emitted *before* the embedded children are yielded, so a consumer that stops iterating early (e.g. the Connector Builder record limit) still sees it; fetched children stay streamed, so the shortfall warning is emitted once the retriever is exhausted (a consumer that stops early cut the fetch short itself, so no shortfall can be reported).
- `RecordExpander`: items yielded by the nested retriever are unwrapped from `Record` or `AirbyteMessage(type=RECORD)`; other protocol messages a `CustomRetriever` may yield (LOG, TRACE, STATE, CONTROL, as `AirbyteMessage` envelopes or as bare `AirbyteLogMessage`-style payloads) are skipped rather than treated as child records.
- `RecordExpander`: falsy/missing indicator is unchanged (no request, no warning); an empty retriever result falls back to the embedded items; retriever request failures propagate through its `error_handler` and fail the stream.
- Validation: `truncated_list_retriever` requires `truncation_indicator_path`; dpath glob metacharacters (`*`, `?`, `[`) are rejected in the indicator path always and in `expand_records_from_field` when a retriever is set, checked on the interpolated values at construction time; `ValueError` from dpath on a non-mapping segment is treated as not truncated.
- `ModelToComponentFactory.create_record_expander`: builds the nested `SimpleRetriever` with name `record_expander_truncated_list`, an auxiliary `log_formatter` (`is_auxiliary=True`), and `message_repository`; rejects `partition_router` / `pagination_reset` on it; builds `CustomRetriever` from `model` + `config` plus the same auxiliary `log_formatter` so manifest values survive and nested requests are still classified as auxiliary.
- Schema (`declarative_component_schema.yaml`) and generated model: field descriptions cover the warnings, glob restriction, request amplification and `use_cache`, paginator semantics, `$parameters` propagation, failure semantics, unsupported nested fields, Builder auxiliary/test-read behavior, and the older-CDK compatibility note.
- Tests: expander unit tests, factory tests, and a new `HttpMocker` end-to-end test module.

## Review Spotlight

Reviewers with limited time, please review first:
- [`record_expander.py`](https://github.com/airbytehq/airbyte-python-cdk/blob/devin/1787781944-record-expander-truncated-list/airbyte_cdk/sources/declarative/expanders/record_expander.py): `expand_record`, `_fetch_complete_list`, and the two warning paths
- [`model_to_component_factory.py` `create_record_expander`](https://github.com/airbytehq/airbyte-python-cdk/blob/devin/1787781944-record-expander-truncated-list/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py): nested retriever wiring and rejections
- [`test_record_expander_http.py`](https://github.com/airbytehq/airbyte-python-cdk/blob/devin/1787781944-record-expander-truncated-list/unit_tests/sources/declarative/expanders/test_record_expander_http.py): end-to-end nested fetch with pagination

## Risks

- One nested fetch per truncated parent record: request amplification on streams with many truncated parents. Mitigation documented: `use_cache: true` on the nested requester.
- Connector Builder test reads apply the page cap to each nested fetch independently, so the nested list can look truncated in a test read even though full syncs read every page. The incomplete-fetch warning is suppressed in test reads for that reason when the nested paginator is capped, so a test read does not surface a real shortfall from a paginated nested retriever either; a nested retriever without a `paginator` is not capped and still warns.
- Older CDK versions silently ignore the new fields; adopting connectors must pin a CDK version that includes them.
- Behavior with the new fields unset is unchanged; `source-stripe` is the only current `RecordExpander` consumer.

## Open questions for maintainers

- **Retriever failure tolerance**: fail-loud is the default here. If opt-in tolerance (e.g. `error_handler` with `IGNORE` falling back to embedded items) is wanted, it should ship together with the incomplete-fetch warning so it cannot degrade into silent loss.
- **Builder test-read cap**: `create_default_paginator` wraps the nested paginator in `PaginatorTestReadDecorator`. A proper fix (feeding nested page counts into `test_read_limit_reached` or exempting this paginator) touches Builder plumbing outside this PR; this PR only suppresses the incomplete-fetch warning under the cap.
- **Generated model**: `declarative_component_schema.py` is hand-edited here, scoped to the `RecordExpander` additions plus `RecordExpander.update_forward_refs()`; the hand-edited `RecordExpander` class body is byte-identical to codegen output. Regenerating the file at the merge base (`4855c2da`) and at head (`eba756f7`) with `bin/generate_component_manifest_files.py` (`datamodel_code_generator==0.26.3`) reorders the same 9 top-level classes in both runs (`BlockSimultaneousSyncsAction`, `DpathExtractor`, `HttpMethod`, `QuotaStatusSource`, `RateLimitedMultipleTokenAuthenticator`, `RecordExpander`, `RecordSelector`, `ScopesJoinStrategy`, `TokenQuota`), and both runs emit `OptionalScope` + `Scope` where the checked-in file has `OAuthScope`. All of that divergence is pre-existing on `main`; this PR introduces none of it (the ~65 extra diff lines at head are this PR's own new content). Regenerating the file belongs in a separate PR against `main`. A `poe assemble && git diff --exit-code` gate cannot pass on this PR alone because the drift predates it; such a gate needs the regeneration PR to land first.
- **Extend vs. new component**: see design notes below; open to disagreement.

## Follow-ups

- Thread the real stream name into the nested retriever's Builder log messages (needs `create_record_selector` plumbing).
- Optional per-sync cap / LRU on nested fetches, pending a policy for what happens when the cap is hit.

## Test plan

- `poetry run pytest unit_tests/sources/declarative/expanders/` (44 tests): retriever fetch / fallback / no-call paths; validation (missing indicator path, every glob metacharacter in both paths, interpolated-glob rejection, safe interpolated path accepted, wildcards still allowed without a retriever); dpath `ValueError` treated as not truncated; scalar fetched items with and without `remain_original_record`; retriever errors propagate; incomplete-fetch warning (once, counts only) and its no-warning cases (count matches, `total_count` missing/bool/string); no-retriever warning (once, counts/paths, no total when absent, none when falsy or when a retriever is configured); incomplete-fetch warning suppressed with `suppress_incomplete_fetch_warning=True` and still emitted without it, no-retriever warning unaffected by the flag; warnings through `MessageRepository`; `HttpMocker` end-to-end: truncated parent triggers `GET /invoices/{id}/lines` across two `starting_after` pages, one call per page, 15 fetched + untruncated parents' embedded items; outer `request_parameters` do not leak into the nested request.
- `poetry run pytest unit_tests/sources/declarative/parsers/test_model_to_component_factory.py`: `SimpleRetriever` wiring (name, `message_repository`, `suppress_incomplete_fetch_warning` false by default, and with `limit_pages_fetched_per_slice=2` true for a capped `DefaultPaginator` and false for `NoPagination`), `CustomRetriever` preserves manifest `name`/`primary_key`, `partition_router` and `pagination_reset` rejected.
- `poetry run ruff check .`, `poetry run ruff format --check .`, `poetry run mypy --config-file mypy.ini airbyte_cdk` clean locally.
- End-to-end against a Stripe-shaped fixture lives in the dependent source-stripe PR's integration tests.

## Design notes: why, survey results, extend vs. new component

<details>
<summary>Show/Hide Content</summary>

### Why

Stripe's `/v1/events` payloads embed only the first page (10 items) of nested list objects, with `lines.has_more: true` and `total_count` reflecting the real size, verified by live measurement (invoices with 15/16/20 lines each embed exactly 10). `RecordExpander` (introduced in https://github.com/airbytehq/airbyte-python-cdk/pull/859) had no way to follow that, so source-stripe's events-based `invoice_line_items` incremental path silently drops line items 11+ of any invoice with more than 10 lines. Stripe rejects `expand[]=data.data.object.lines` on `/v1/events`, so there is no request-side workaround.

### Survey results (generality)

A survey of certified/GA connectors for the same embedded-list-truncation pattern found **no other connector that can adopt the retriever path today**: source-stripe is the only connector with an embedded nested list, a truncation flag, and a dedicated complete-list endpoint. The closest real data-loss cousin is `source-intercom` `conversation_parts`: Intercom embeds at most the 500 most recent parts of a conversation and exposes **no** endpoint to fetch the rest, so it can never use `truncated_list_retriever`. That unrecoverable case is what the warn-on-truncation path covers: it converts silent data loss into visible data loss, and it is what makes this change generally useful beyond Stripe. Safe contrasts (nested pagination correctly followed): source-monday `items_page.cursor`, source-github GraphQL `pageInfo.hasNextPage`, and any `lazy_read_pointer` / `LazySimpleRetriever` user.

### Extend vs. new component

Patrick Nilan raised whether giving `RecordExpander` (previously a pure in-memory dpath transform) a retriever, and therefore HTTP capability, fundamentally changes what the component is, and whether a brand-new component would be better. This PR takes the **extend** position; that decision is open to disagreement:

- The component's contract ("given a parent record, yield the complete set of child records from its nested list") is unchanged; the retriever is a fallback for honoring it when the payload is incomplete. Both fields are optional and unset behavior is identical.
- The CDK already has declarative components composing retrievers/streams (`SubstreamPartitionRouter`, `AsyncRetriever`), so a component owning a retriever is not novel.
- Exactly one connector uses `RecordExpander` today (source-stripe), so a new component would duplicate nearly all of `RecordExpander`'s surface for one user and add a deprecation/migration burden; the truncation config is generic (arbitrary dpath indicator + standard `SimpleRetriever`/`CustomRetriever`), not Stripe-specific.

If reviewers prefer a new component anyway, the natural shape is a `TruncatedListExpander` superset of `RecordExpander` in the same `record_expander` slot.

The alternative of rerouting the events path through `SubstreamPartitionRouter`/`lazy_read_pointer` (which does follow nested pagination) was rejected: `lazy_read_pointer` is coupled to parent-stream partitioning and full-refresh child reads, and forcing the events stream into that shape would break the events cursor/state semantics and `remain_original_record` transformations.

</details>

## Summary by CodeRabbit

- **New Features**
  - Added support for detecting truncated nested lists and retrieving complete contents through configurable paginated requests.
  - Preserved request parameters and supported auxiliary handling for expanded records.
  - Added validation for unsupported truncated-list retrieval options.

- **Bug Fixes**
  - Improved handling of empty or incomplete retrievals, protocol messages, and fallback logging.
  - Prevented duplicate warnings during concurrent record expansion.
  - Suppressed incomplete-fetch warnings during Connector Builder test reads only when a paginator is configured.

- **Documentation**
  - Clarified truncation behavior and Connector Builder test-read limitations.

Link to Devin session: https://app.devin.ai/sessions/66b1cf8a1cf04d9387166364ae684e12
Open in Devin Desktop: https://app.devin.ai/desktop/session/66b1cf8a1cf04d9387166364ae684e12?variant=devin


ZaneHyattAB:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/33128872969
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/33128883500

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -116,10 +166,72 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
                     else:
                         yield item
                     expanded_any = True
+                    embedded_count += 1
+
+        if truncated and not self.truncated_list_retriever:
```
tolik0: **P1 — configuring a retriever makes truncation unreportable (silent data loss in the feature built to prevent it)**

The warning is gated on the retriever's *absence*, so once one is configured truncation is unreportable **regardless of what the retriever returned**. Nothing compares the fetched count against the API's own completeness signal — even though `_get_sibling_total_count` (`:207`) already reads exactly that, and is called only from the warning path.

This is easy to hit. `truncated_list_retriever` does not require a `paginator`, and when `model.paginator` is absent `create_simple_retriever` falls through to `NoPagination` (`SimpleRetriever.__post_init__`: `self._paginator = self.paginator or NoPagination(...)`). A retriever without a paginator fetches the sub-endpoint's default page and stops.

Reproduced at this head — parent with `total_count: 100`, 10 embedded items, retriever returning one 10-item page:

```
records emitted: 10   warnings: 0
API-reported total_count: 100
-> 90 line items lost, silently
```

Before this PR the same manifest with an indicator and no retriever emitted 10 records **and** a WARNING. Adding the retriever removes the warning and keeps the loss.

Suggestion: count what `_fetch_complete_list` yields and warn once when it falls short of `total_count`, reusing the helper you already have:

```python
expected = self._get_sibling_total_count(parent_record, indicator_path)
if expected is not None and fetched_count < expected:
    self._warn_incomplete_fetch(expand_path, fetched_count, expected)
```

Also worth requiring a `paginator` on `truncated_list_retriever`, or at minimum calling out in the field description that omitting one silently truncates.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -70,12 +86,33 @@ class RecordExpander:
...
     def __post_init__(self, parameters: Mapping[str, Any]) -> None:
         self._expand_path: list[InterpolatedString] = [
             InterpolatedString.create(path, parameters=parameters)
             for path in self.expand_records_from_field
         ]
+        if self.truncated_list_retriever and not self.truncation_indicator_path:
+            raise ValueError(
+                "`truncation_indicator_path` is required when `truncated_list_retriever` is configured."
+            )
+        if any(path == "*" for path in (self.truncation_indicator_path or [])):
```
tolik0: **P1 — the wildcard guard rejects only the literal `"*"`; a glob that matches twice crashes the sync**

`dpath.get` honours full glob syntax, so `has_*`, `h?` and `h[ab]` all pass this guard (and the one at `:106`) and then glob at read time. When a glob matches **two** leaves `dpath.get` raises `ValueError` — and `_is_truncated` at `:216` catches only `KeyError`, while its sibling `_get_sibling_total_count` at `:208` catches `(KeyError, ValueError)` on the same call shape.

The `ValueError` escapes `expand_record` → `DpathExtractor.extract_records` → `RecordSelector` → `SimpleRetriever._read_pages` → the partition, killing the stream *after* N records were emitted. Because it depends on the **record shape**, a Builder test read on well-behaved data passes and production breaks later.

A second bypass is total: these guards run on the **raw** strings in `__post_init__`, before interpolation, and the schema declares `interpolation_context: [config]` on both fields — so an interpolated `*` is never checked. That contradicts the description added at `declarative_component_schema.yaml:2263`.

Reproduced at this head (dpath 2.2.0):

```
guard ACCEPTED ['a', 'has_*']
guard ACCEPTED ['a', 'h?']
guard ACCEPTED ['a', 'h[ab]']
RAISED from expand_record: ValueError: dpath.get() globs must match only one leaf: ['a', 'h[ab]']
interpolated '*' construction: ACCEPTED
RAISED from expand_record: ValueError: dpath.get() globs must match only one leaf: ['a', '*', 'has_more']
```

List traversal, scalar traversal, unclosed `[`, and `None`/int segments all raise `KeyError` and are safe.

Suggestion:

```python
# :216 — match the sibling at :208
-        except KeyError:
+        except (KeyError, ValueError):
             return False

# here — reject every dpath glob metacharacter, not just "*"
+        _GLOB_CHARS = ("*", "?", "[")
+        if any(any(ch in path for ch in _GLOB_CHARS)
+               for path in (self.truncation_indicator_path or [])):
```

Interpolated paths can't be checked at construction — validate the evaluated path on first use, or say in the description that interpolated segments are unchecked. Neither the multi-match glob nor the interpolated-`*` bypass is covered by the 12 new tests.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,16 @@ def create_record_expander(
         config: Config,
         **kwargs: Any,
     ) -> RecordExpander:
+        truncated_list_retriever = None
+        if model.truncated_list_retriever:
+            truncated_list_retriever = self._create_component_from_model(
```
tolik0: **P1 — the enclosing stream's `$parameters` are injected into this nested retriever's requester**

`ManifestComponentTransformer.propagate_types_and_parameters` walks every nested dict and applies `propagated_component[parameter_key] = propagated_component.get(parameter_key) or parameter_value` (`manifest_component_transformer.py:203-206`). Because `truncated_list_retriever` is a full component subtree under the stream's retriever, the stream's `$parameters` land on the **sub-requester** for every field it doesn't set explicitly — so the parent endpoint's query params get sent to a different endpoint.

Built from a manifest mirroring source-stripe:

```
sub request PATH:   invoices/in_1/lines
sub request PARAMS: {'types[]': ['invoice.created']}      <-- inherited from the events stream
```

and end-to-end against mocked HTTP:

```
GET .../v1/events?types%5B%5D=invoice.created
GET .../v1/invoices/in_1/lines?types%5B%5D=invoice.created   <-- stray param
```

This is *standard* CDK propagation, not a regression — `ParentStreamConfig.stream` leaks identically, and `use_parent_parameters` doesn't help (it only flips precedence, never suppresses). So the mechanism isn't the finding. What makes it P1 is that the intended consumer already hit it: airbytehq/airbyte#85087 moves `request_parameters` off `$parameters` and onto the events requester, and its integration test asserts the exact sub-request query string (`.with_limit(100)`, no `with_any_query_params()`), so the restructure was mandatory rather than stylistic. Nothing in the schema description, the docstring, or the PR body records that constraint.

Note `subscription_items` (`source-stripe/manifest.yaml:1645`) still uses the `$parameters` form and would hit this the moment a retriever is added there.

Suggestion: don't change the transformer. Add a sentence to the `truncated_list_retriever` description saying the enclosing component's `$parameters` propagate into it and that request-shaping `$parameters` must move onto the outer requester's `request_parameters` — plus the regression test noted on the test file.

unit_tests/sources/declarative/expanders/test_record_expander.py
```diff
@@ -0,0 +1,227 @@
...
+import pytest
+
+from airbyte_cdk.sources.declarative.expanders.record_expander import RecordExpander
+from airbyte_cdk.sources.types import Record, StreamSlice
+
+config = {}
+parameters = {}
+
+
+def _make_retriever(records):
```
tolik0, line 19: **P1 — no CDK test can observe the sub-request this feature exists to make**

All 12 expander tests replace the retriever with `MagicMock()`, and `test_create_record_expander_with_truncated_list_retriever` asserts only `truncation_indicator_path` and `isinstance(..., SimpleRetriever)` without ever calling the requester. So the entire HTTP half of the change — request URL, query params, pagination, error path, and the `NoPagination` trap — has no test in this repo that can see it. The `$parameters` leak is the proof: it's live in the component the factory builds, and 13 green tests don't notice.

"End-to-end verification lives in the dependent source-stripe PR" is true, but it means the CDK can't regression-test its own feature — a future change to this wiring would be caught only in another repo, by a draft PR.

Mutation testing at this head: 13 mutants, 10 killed, **3 survived** — the `isinstance(total, bool)` guard (`record_expander.py:210`), the `Record`-unwrap (`:228`, untestable as written since `Record` *is* a `Mapping`), and the non-`Mapping` filter (`:229`, flagged separately).

Suggestion — one `HttpMocker`/`requests_mock` test closes most of this. Build the selector from a manifest mirroring Stripe (stream-level `$parameters.request_parameters`, a `truncated_list_retriever` with an `HttpRequester` and a `DefaultPaginator`), call `expand_record`, assert the recorded requests. With no leak it produces:

```
-> .../invoices/in_1/lines
-> .../invoices/in_1/lines?starting_after=il_2
```

and adding the stream `$parameters` flips it to the failing form. That single test covers the request URL, the currently-untested pagination claim, and the `$parameters` leak at once.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,16 @@ def create_record_expander(
...
+                model=model.truncated_list_retriever,
+                config=config,
+                name="record_expander_truncated_list",
```
tolik0: **P1 — the constant name makes every sub-request vanish from Connector Builder and manifest-server test-read**

In test-read mode this `name` reaches `create_simple_retriever` → `log_formatter=self._get_log_formatter(log_formatter, name)` → `format_http_message(..., name)`, which sets `log_message["airbyte_cdk"]["stream"]["name"] = "record_expander_truncated_list"` (`sources/http_logger.py:49`) and leaves `is_auxiliary` unset. `connector_builder/test_reader/message_grouper.py:101-102` then drops it:

```python
if is_page_http_request_for_different_stream(json_message, stream_name):
    continue
```

because the constant never equals the real stream name (`helpers.py:302`).

Measured in Builder mode, one parent record, `has_more: true`:

```
HTTP requests actually made: 6
    .../v1/events
    .../v1/invoices/in_1/lines?limit=2          (x5)
BUILDER slice 0: pages=1 aux=0
   page 0: url=.../v1/events   records shown = 10
```

Ten records attributed to a page whose displayed body holds three embedded items; the five requests that produced the other seven appear neither as a page nor in `auxiliary_requests`. The manifest server is identical (`manifest_server/command_processor/processor.py:42` uses the same `TestReader`).

The Builder author is the audience for this feature and gets a debugging dead end. Combined with the warning not reaching Builder logs and the per-parent pagination cap, a Builder user configuring this sees no requests, no warning, and a silently short list.

Secondary: `HttpClient.cache_filename` is `f"{self._name}.sqlite"` (`streams/http/http_client.py:161`), so every `truncated_list_retriever` in a manifest shares one cache DB — source-stripe already has two `record_expander` sites.

Suggestion: smallest fix is `is_auxiliary=True` so it lands in the slice's `auxiliary_requests` panel. Threading the stream name through is better but needs plumbing — `create_record_selector` calls `_create_component_from_model(model=model.extractor, decoder=decoder, config=config)` without `name`, and `create_dpath_extractor` likewise, so the stream name isn't reachable here today. (This is the consequence that wasn't part of the earlier CodeRabbit exchange on this line.)

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,16 @@ def create_record_expander(
         config: Config,
         **kwargs: Any,
     ) -> RecordExpander:
+        truncated_list_retriever = None
+        if model.truncated_list_retriever:
+            truncated_list_retriever = self._create_component_from_model(
+                model=model.truncated_list_retriever,
```
tolik0: **P2 — `partition_router` on the sub-retriever is schema-legal and silently dropped**

The new `anyOf` `$ref`s `SimpleRetriever`, whose model has a `partition_router` field — but this call passes `name`, `primary_key`, `stream_slicer` and `transformations` and **not** `partition_router=`, which is the only way `create_simple_retriever` turns it into a `request_options_provider`.

Verified with a sub-retriever declaring `ListPartitionRouter(values: [A, B], request_option: {field_name: cur, inject_into: request_parameter})`: exactly **1** request, `cur` param present **False**, no error and no warning.

Worst-case edge (reasoned from the code, not executed): `SubstreamPartitionRouter` + `lazy_read_pointer` satisfies `create_simple_retriever`'s lazy branch, since `get_stream_state("record_expander_truncated_list", None)` is empty — returning a `LazySimpleRetriever` whose `_read_pages` reads `stream_slice.extra_fields["child_response"]`, and the `StreamSlice(partition=..., cursor_slice={})` built at `record_expander.py:224` has no `extra_fields`, so `KeyError`.

Suggestion: reject `partition_router` (and `pagination_reset`) on `truncated_list_retriever` here, or narrow the schema `$ref` to a reduced retriever definition.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -116,10 +166,72 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
...
+        self, parent_record: Mapping[str, Any]
+    ) -> Iterable[MutableMapping[str, Any]]:
+        if not self.truncated_list_retriever:
+            return
+        stream_slice = StreamSlice(partition={"parent_record": parent_record}, cursor_slice={})
+        for item in self.truncated_list_retriever.read_records(
+            records_schema={}, stream_slice=stream_slice
+        ):
+            data = item.data if isinstance(item, Record) else item
+            if not isinstance(data, Mapping):
```
tolik0: **P2 — non-`Mapping` fetched items are dropped, the HTTP call is wasted, and the truncated list is emitted instead**

The embedded path handles non-dict items — it yields the scalar, or `{"value": item, "original_record": ...}` when `remain_original_record` (`:160-166`). This path just drops them.

Verified end-to-end through a full `ConcurrentDeclarativeSource.read` with a sub-endpoint returning `{"data": ["il_a","il_b","il_c","il_d"]}` (an extractor pointed at a list of ids or URLs — legal, and legal on the embedded path):

```
records: 6   <- the truncated embedded items
HTTP calls: 4  <- 3 sub-fetches paid for and thrown away
```

The schema says "*If the retriever returns no records, the embedded items are expanded as a fallback*" — here the retriever returned four records. And if *every* item is non-`Mapping`, the `first_fetched is None` peek makes a malformed sub-response look like a successful truncated read. No log either way.

This is also one of the three surviving mutants: deleting both lines leaves all 50 tests green.

Suggestion: mirror the embedded path's scalar handling, or at minimum log the drop instead of `continue`.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -89,8 +126,20 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
             return
 
         parent_record = record
+
+        truncated = bool(self._truncation_indicator_path) and self._is_truncated(parent_record)
+
+        if truncated and self.truncated_list_retriever:
+            fetched_records = iter(self._fetch_complete_list(parent_record))
```
tolik0: **P2 — a failing retriever loses the embedded items too, and aborts the parent stream**

`_fetch_complete_list` is consumed with no `try`, so any exception escapes **after** partial records were already yielded, and the embedded items the pre-PR code would have emitted are lost for that record. Verified: `RAISED RuntimeError: 404 invoice deleted -> embedded items NOT emitted`, and in a full read a 404 from the sub-endpoint produced `records: 0` for the whole parent stream.

For Stripe this is a live shape: `invoice.deleted` is one of the three event types at `source-stripe/manifest.yaml:1490`, and `/v1/invoices/{id}/lines` for a deleted invoice won't resolve — so one deleted invoice with >10 lines fails the entire `invoice_line_items` stream where before it silently emitted 10. There's also no way to make the fetch conditional; the retriever fires for every truncated record, including ones where the fetch cannot succeed.

The error *surfacing* is correct, for what it's worth — a proper `AirbyteTracedException` with `stream_descriptor=StreamDescriptor(name='events')`, attributed to the real parent stream rather than the phantom sub-stream.

Failing loudly may well be the right default, but the schema sentence reads as blanket resilience. Either scope it to "returns no records" and state that a request failure surfaces through the retriever's own `error_handler` and fails the stream, or add opt-in tolerance — and if you add tolerance, fix the missing incomplete-fetch warning first, otherwise an `error_handler` with `action: IGNORE` degrades straight into silent loss. Either way it needs a test; there is none today.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -89,8 +126,20 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
...
+            first_fetched = next(fetched_records, None)
```
tolik0: **P2 — request amplification: one full paginated fetch per truncated parent record, no dedup**

`next(...)` is what triggers the request, so it fires from inside the parent's `for current_record in records_generator_fn(response)` loop while the parent `requests.Response` is live. Cost is O(truncated parent records) × sub-pagination pages, serial, with no cap, batching or memoisation.

Stripe's events stream emits one event per invoice status change, so the same invoice appears in many events within a 30-day window and each occurrence re-fetches `/v1/invoices/{id}/lines` in full — an invoice with 20 status events and 3 pages of lines costs 60 requests to produce the same 15 line items 20 times over. Retries multiply it: a single 500 burned 6 attempts and 31s of backoff on one parent record.

Measured: 4 HTTP requests for 3 truncated parent records (1 parent page + 3 sub-fetches), `urls: ['.../v1/events', '.../v1/invoices/in_0/lines', '.../v1/invoices/in_1/lines', '.../v1/invoices/in_2/lines']`.

Related and pre-existing rather than new: `_apply_parent_context` runs `copy.deepcopy(parent_record)` per item when `remain_original_record` is set (source-stripe sets it). That already happened per embedded item, but a parent with `total_count: 500` and 10 embedded items goes from 10 deep copies to 500 — a ~50x amplification of an existing cost.

Checked and clean, so nobody over-claims: the sub-requester **does** receive the source-level `api_budget`, so sub-calls count against the shared rate limit; and it gets its own `requests.Session` (distinct object ids), so there's no nested same-session reentrancy and no deadlock.

Suggestion: document the amplification in the field description, point adopters at `use_cache: true`, and consider a per-sync cap plus a small LRU keyed on the interpolated sub-request URL.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2258,6 +2258,22 @@ definitions:
...
+        items:
+          type: string
+        interpolation_context:
+          - config
+        examples:
+          - ["data", "object", "lines", "has_more"]
+      truncated_list_retriever:
+        title: Truncated List Retriever
+        description: Retriever used to fetch the complete list of items when the field at `truncation_indicator_path` is truthy on a record. The record being expanded is exposed to the retriever's interpolation context as `stream_slice['parent_record']`. If the retriever returns no records, the embedded items are expanded as a fallback. Requires `truncation_indicator_path`.
+        anyOf:
```
tolik0, line 2274: **P2 — in test-read, this retriever's pagination is capped per parent record, and `test_read_limit_reached` stays False**

`create_default_paginator` wraps unconditionally on the factory-level flag (`model_to_component_factory.py:2471-2472`):

```python
if self._limit_pages_fetched_per_slice:
    return PaginatorTestReadDecorator(paginator, self._limit_pages_fetched_per_slice)
```

Nothing distinguishes this sub-retriever's paginator from a stream's. Verified with a 20-item endpoint at `page_size=2` and `max_pages_per_slice=5`:

- 1 parent record → 5 sub-requests, **10 of 20 items**, `test_read_limit_reached: False`
- 2 parent records → 11 requests total; the page budget resets **per parent record**

So a Builder user testing a truncation-recovery config sees a silently short list with the limit flag explicitly false, and a test read can issue `max_pages_per_slice` extra requests per parent record rather than per slice. Normal syncs are unaffected (`_limit_pages_fetched_per_slice` is `None` outside test-read).

At minimum this makes the PR body's "including full pagination via the retriever's own paginator" true for syncs but not test-reads. Better: feed the sub-retriever's page count into `test_read_limit_reached`, or exempt its paginator from the decorator.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -116,10 +166,72 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
...
+    ) -> None:
+        if self._warned_truncation_without_retriever:
+            return
+        self._warned_truncation_without_retriever = True
+
+        indicator_path = [path.eval(self.config) for path in self._truncation_indicator_path]
+        expand_path = [path.eval(self.config) for path in self._expand_path]
+        total_count = self._get_sibling_total_count(parent_record, indicator_path)
+        total_fragment = f" of {total_count} total" if total_count is not None else ""
+        logger.warning(
```
tolik0: **P2 — this WARNING never reaches Builder logs or the manifest-server response**

`StreamRead.logs` is assembled only from AirbyteMessages in the read generator plus deprecation warnings (`connector_builder/test_reader/reader.py:136-149`). A stdlib `logging` call isn't in that generator.

Verified: in an indicator-only run `StreamRead.logs` held only the two `url_base`/`path` deprecation warnings, and this warning appeared solely as a separate top-level `{"type":"LOG","log":{"level":"WARN",...}}` line on stdout — which `connector_builder/main.py` doesn't include in its single serialized response. For the manifest server it's lost outright, going to the server process's stdout.

So it *does* reach platform sync logs and a Cloud user sees it — but the Builder user, whom the message addresses by name ("Configure `truncated_list_retriever`"), does not.

Suggestion: emit through the message repository, the pattern `models/base_model_with_deprecations.py:135` already uses to build a `ConnectorBuilderLogMessage`.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -3485,6 +3496,7 @@ class DynamicDeclarativeStream(BaseModel):
 FileUploader.update_forward_refs()
 DeclarativeStream.update_forward_refs()
 SessionTokenAuthenticator.update_forward_refs()
+RecordExpander.update_forward_refs()
```
tolik0, line 3499: **P2 — hand-edited generated file: functionally correct, but unverified, and the stated reason is wrong**

Two independent reproductions of the pinned pipeline (`datamodel_code_generator==0.26.3`, replaying `bin/generate_component_manifest_files.py`'s flags and post-processing) agree:

- the generated `class RecordExpander` body is **byte-identical** to the hand edit;
- real codegen would additionally relocate `RecordExpander`, `DpathExtractor` and `RecordSelector` and emit `DpathExtractor.update_forward_refs()` as well;
- that second call is required **only by codegen's ordering** (codegen puts `DpathExtractor` *before* `RecordExpander`; the checked-in file has it *after*), so omitting it is correct here. Runtime verified — `RecordExpander.parse_obj` and nested `DpathExtractor.parse_obj` both resolve `SimpleRetriever`.

So no functional defect. Two things remain:

1. **The PR body's causal claim doesn't hold.** A regen-vs-regen experiment holds the pipeline constant, so drift cancels — and the reordering still appears. It's caused by *this PR* adding a forward `$ref` to `SimpleRetriever`, and the classes that move are precisely this PR's own dependency chain. Nothing unrelated moves. As written, "reorders many unrelated classes because the checked-in file has drifted" will read to the next maintainer as license to hand-edit generated files.
2. **Pre-existing drift on `main` is separately real** — regenerating from the base schema and diffing against the checked-in file at base yields 992 diff lines / 26 hunks / 23 class relocations. Version-proof evidence of earlier hand-edits: the schema declares `minItems: 1` on `RateLimitedMultipleTokenAuthenticator.tokens` but the checked-in model has no `min_items`, which no codegen version emits.

No CI check verifies any of this — `grep -rn "assemble\|generate_component" .github/workflows/` returns nothing, so regeneration only happens on a manual `/poe build`.

Suggestion: run `/poe build` on this PR and take the reordering (it's this PR's own consequence, so it belongs in this diff); file a separate regeneration PR for the pre-existing drift; and fix the body sentence. A `poe assemble && git diff --exit-code` gate would retire this whole class of review.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -2258,6 +2258,22 @@ definitions:
...
+        title: Truncation Indicator Path
+        description: Path within each record to a field indicating that the embedded nested list is truncated (e.g. a `has_more` flag on the list object). When the field evaluates to a truthy value and `truncated_list_retriever` is configured, the retriever is used to fetch the complete list instead of expanding the embedded items. When the field is truthy and no retriever is configured, the embedded items are expanded as normal and a WARNING is logged once per stream so the truncation is visible instead of silent. Wildcards (*) are not supported in this path, nor in `expand_records_from_field` when a retriever is configured.
+        type: array
+        items:
+          type: string
+        interpolation_context:
+          - config
+        examples:
+          - ["data", "object", "lines", "has_more"]
+      truncated_list_retriever:
```
tolik0, line 2271: **P2 — both fields degrade silently on an older pinned SDM or CDK**

`RecordExpander` declares no `additionalProperties: false`. Verified against the base commit: a manifest carrying both new fields **validates** against the base schema, and the base pydantic model parses it with `extra fields present? False` — both fields are dropped.

So a Builder connector authored against the new SDM but executed on a pinned older image gets no sub-fetch **and** no warning: straight back to silent truncation. That's unlike a new component `type:`, which fails loudly because component types are enumerated.

Suggestion: note the minimum CDK version in the PR body and in the field description.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,16 @@ def create_record_expander(
...
+                config=config,
+                name="record_expander_truncated_list",
+                primary_key=None,
+                stream_slicer=None,
```
tolik0: **P2 — a `CustomRetriever` in this slot silently loses its manifest values**

`create_custom_component` gives caller kwargs precedence over the manifest — `model_args[key] = arg` at `:1871-1872`, with the comment *"When there are field collisions we defer to these arguments over the component's definition"* — and only afterwards filters to the class's declared type hints (`:1921-1925`).

So extras don't raise (a trivial custom retriever builds and runs fine, with `stream_slice["parent_record"]` correctly populated), but a custom retriever that type-hints `name`, `primary_key`, `transformations` or `stream_slicer` silently receives `"record_expander_truncated_list"` / `None` / `[]` / `None` instead of what its manifest block declares. The schema permits `CustomRetriever` and there's no test for that arm.

Separately, `stream_slicer=None` on this line is dead: it isn't a declared parameter of `create_simple_retriever` (it's swallowed by `**kwargs`), and inside, the retriever is always built with `stream_slicer=_NO_STREAM_SLICING` (`:3660`, `:3681`). It reads as "suppress slicing" and does nothing.

Suggestion: drop `stream_slicer=None`, pass `primary_key`/`transformations` only for the `SimpleRetriever` arm, and add a `CustomRetriever` happy-path test.

tolik0 (commented):
Review of `87ab6325` — posting the P1 and P2 findings inline. Full write-up kept locally; summary of the verdict below.

**Breaking change: NON_BREAKING.** Both fields default to `None`, `truncated = bool([]) and ...` short-circuits, and every new branch is dead on the unset path — so no `cdk-migrations.md` entry and no manifest migration are owed, and `feat(low-code):` → minor bump is right. `RecordExpanderModel` is already in `PYDANTIC_MODEL_TO_CONSTRUCTOR` since #859.

The wiring is sound and the hand-edited generated class body is provably byte-identical to codegen. What holds this back is behavioural, and the headline is uncomfortable for a PR that exists to stop silent data loss: **configuring `truncated_list_retriever` makes the truncation warning unreachable**, and since the schema doesn't require a `paginator` and `create_simple_retriever` defaults to `NoPagination`, forgetting one is the easy mistake — 10 records emitted, 0 warnings, 90 line items lost, while the component already holds `total_count: 100`.

**5 P1** — the above; a glob guard that only rejects the literal `"*"` while dpath honours `?`/`[ab]`/`*`-substrings, throwing an uncaught `ValueError` that kills the stream (data-dependent, so it survives a Builder test read); `$parameters` leaking into the nested requester, which airbytehq/airbyte#85087 already had to restructure its manifest to dodge; no CDK test reaching a real sub-request (all 12 mock the retriever; 3 of 13 mutants survive); and the constant sub-retriever name making every truncated-list request vanish from Connector Builder.

**9 P2** — `partition_router` schema-legal and silently dropped; non-`Mapping` fetched items dropped with the HTTP call wasted and the truncated list emitted instead; a failing retriever losing the embedded items and aborting the parent stream; request amplification with no dedup; test-read capping sub-pagination per parent record; the warning never reaching Builder; the hand-edited generated file; silent field-drop on an older pinned SDM; and `CustomRetriever` losing manifest values.

Also worth flagging, outside the diff:

- **The linked oncall issue doesn't support the PR body's claim.** oncall#12975 is about the `invoices` *cursor*, and its headline symptom already got a different fix (airbytehq/airbyte#82228). Comment 45 (2026-09-01) came back **negative** on the truncation hypothesis: the reported invoice has 1 line item in Stripe and 1 row in Airbyte. The mechanism is separately supported by your mocked-Stripe repro on 6.0.13, but the issue can't carry "drops line items 11+ of any invoice", and the customer's symptom is still unexplained.
- **`Test Connectors` proves nothing here** — 0 of the 6 matrix connectors reference `RecordExpander`, and source-stripe isn't in the matrix. Both red jobs are environmental (`destination-motherduck` `PermissionError(13)` and `source-shopify` at 6h0m15s are red on #1141/#1142/#1145 too). `Validate PR title` and CodeRabbit are vacuous passes — every step is gated on `draft == false`. The `/prerelease` from this PR already succeeded, so a pinned source-stripe run is the missing validation.
- `subscription_items` (`source-stripe/manifest.yaml:1645`) has identical exposure and isn't covered by #85087; and #85087 bumps to 6.0.16, which #85672 took on 2026-09-08, so it needs a rebase to 6.0.17.

Things I checked that are clean, so they don't cost you time: secret filtering does cover the new warning (`filter_secrets` on the WARNING branch — an interpolated secret renders `['****', 'data']`); no state pollution (`SimpleRetriever` has no `cursor` field, so the sub-retriever's `FinalStateCursor` can never emit); mid-sync failure can't skip records; the parent cursor still resolves because source-stripe derives `invoice_updated` from `original_record`, which the fetched path preserves; the sub-retriever counts against the shared `api_budget` and holds its own session; paginators are stateless at head; and the `anyOf` cycle isn't novel (11 cycles at base, 12 at head).

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
ZaneHyattAB: Correcting the reply above. That test name doesn't exist. The test that covers this is test_stream_parameters_propagate_into_truncated_list_requester.
It also does the opposite of what the name suggested. It asserts that the propagation happens, rather than proving it doesn't. It removes request_parameters from the events requester, puts it on the stream's $parameters, and asserts the nested call goes out as GET /v1/invoices/in_1/lines?types[]=invoice.created. That is the stray param you found, locked into a test so it can't change without failing.
The constraint is written into the schema description, the generated model and the docstring: "$parameters of the enclosing stream propagate into this retriever's components (including request_parameters on its requester); move request-shaping parameters to the outer requester's request_parameters when adopting this field."

airbyte_cdk/sources/declarative/expanders/record_expander.py
ZaneHyattAB: **Deleted invoices don't break it.** Tested live, then end to end.

Stripe really does 404 `/lines` for a deleted invoice, and the event really is
truncated — your read was right. But on `eba756f7` + #85087's manifest the sync
survives: 22 records, 0 errors, the healthy invoice fully recovered, the deleted
one falling back to its 10 embedded lines with one warning.

Reason: `base_requester` already IGNOREs 404, and the retriever inherits it, so
the fetch never raises. Remove that and your abort reproduces exactly — 0 of 12,
no warning. Real bug, just not one source-stripe can hit.

Your worry that IGNORE turns into silent loss might not hold either —
`_warn_if_fetch_incomplete` fires on the zero-record path (`ebab4e7`).

**A 500 does break it:** 6 retries, then the stream aborts, 0 of 12, no warning.
That's new.

So one thing to decide, should 5xx be IGNOREd on the sub-requester too? If yes,
a 500 falls back to the embedded lines and warns, like the 404 case. If no, it
fails the stream and the next sync retries. I'd say no: you'd need 6 straight
failures on one call, and if Stripe is down `/v1/events` fails anyway. But it's
a one-line manifest change either way.

Otherwise: fail-loud stays, and #85087 needs no `error_handler` change.

airbyte_cdk/sources/declarative/models/declarative_component_schema.py
ZaneHyattAB: I ran the generator at the merge base and at head to see what this PR actually
causes.

The answer is nothing. The same set of classes is reordered in both runs and the
difference between them is empty. RecordExpander, DpathExtractor and
RecordSelector are in that set, so you're right that the movers are this PR's
dependency chain, but they move at the base too, with the PR absent. CheckStream
doesn't move in either run; its body is byte identical and the one place shift is
BlockSimultaneousSyncsAction moving past it.

Agreed the drift on main is real and separate. Our counts differ, probably
methodology, and it doesn't change the point either way.

So the reordering is all pre existing and regenerating belongs in a separate PR,
which is option (b) you suggested. I'll fix the PR body, which currently claims
the opposite.

On the gate: `poe assemble && git diff --exit-code` can't pass on this PR
whatever it does, for the same reason. The regeneration PR would need to land
first.

airbyte_cdk/sources/declarative/declarative_component_schema.yaml
ZaneHyattAB: Reproduced it, and it was worse than a short list.

Correct manifest, 250 line items, page_size 100, max_pages_per_slice 2. Test
read returns 200 of 250 with test_read_limit_reached False, as you said. The new
truncation warning also fired, telling the user to configure a paginator that
was already there. So we were turning your silent short list into a false bug
report.

Fixed in 95b7b2ac. The warning is now skipped only when the nested paginator is
actually wrapped in PaginatorTestReadDecorator, not whenever a test read runs.
Skipping on the test read flag alone would have silenced a retriever with no
paginator, which is the silent case you raised separately. Checked both ways:
gone under the cap, still firing in a normal sync.

The cap itself is untouched. Your two suggestions both reach into Builder
plumbing, so I'd keep those separate unless you want them here.

Separately, auxiliary_requests shows both nested calls, so sub requests being
invisible in Builder is fixed.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,42 @@ def create_record_expander(
...
+                    log_formatter=log_formatter,
+                )
+                # Only a capped paginator makes a shortfall expected; `NoPagination` is never capped.
+                suppress_incomplete_fetch_warning = isinstance(
+                    truncated_list_retriever.paginator, PaginatorTestReadDecorator
+                )
+            else:
+                truncated_list_retriever = self._create_component_from_model(
+                    model=retriever_model, config=config, log_formatter=log_formatter
+                )
```
tolik0: **P1 — regression: a `CustomRetriever` in this slot no longer builds at all**

Round 1 passed `name`/`primary_key`/`transformations` on a single call. Splitting the arms left this one with only `model`, `config` and `log_formatter` — and those three kwargs are exactly what `create_custom_component` forwards into `_create_nested_component` (`:1891`, `:1909`), which selects from them by the nested constructor's keyword-only args (`:2035-2037`). `create_record_selector` requires `name`; `create_simple_retriever` requires `name`, `primary_key` **and** `transformations`. So any `CustomRetriever` that owns a nested `requester` / `record_selector` / `retriever` fails at source-construction time.

Reproduced at both heads with the same manifest, no `$parameters` injected:

```
round-1 head 87ab6325:  BUILD OK -> MyLinesRetriever
round-2 head 95b7b2ac:  ValueError: Error creating component 'RecordSelector' with parent
                        custom component mycomp.MyLinesRetriever:
                        Please provide RecordSelector.$parameters.name
```

and for a nested `SimpleRetriever`: `Please provide SimpleRetriever.$parameters.name, …primary_key, …transformations`. Narrowing which kwarg is load-bearing:

```
model+config only                   -> ValueError
+ name                              -> OK   (nested RecordSelector)
+ name,primary_key,transformations  -> OK   (nested SimpleRetriever)
```

**Why CI is green:** `test_create_record_expander_with_custom_truncated_list_retriever` passes only because the shared fixture `_record_expander_selector` (`unit_tests/sources/declarative/parsers/test_model_to_component_factory.py:2113-2114`) injects `$parameters: {name: "lists"}`, which `propagate_types_and_parameters` pushes into every nested dict. Delete those two lines and that one test fails while the other seven still pass (1 failed, 7 passed, via `pytest unit_tests/sources/declarative/parsers/test_model_to_component_factory.py -q -k record_expander`).

Blast radius is zero shipped connectors — the field is new — but the `anyOf` advertises `CustomRetriever` and the arm is non-functional for any non-trivial one.

**Fix** — collapse the branch back to one call, which is also what the three other nested-retriever factories accepting `SimpleRetriever | CustomRetriever` do, unconditionally and without branching on model type: `create_dynamic_schema_loader:2747-2766` (closest precedent — it also builds an `is_auxiliary=True` formatter), `create_properties_from_endpoint:3230-3238`, `create_http_components_resolver:4436-4443`.

```python
truncated_list_retriever = self._create_component_from_model(
    model=retriever_model, config=config, name=name,
    primary_key=None, transformations=[], log_formatter=log_formatter,
)
suppress_incomplete_fetch_warning = isinstance(
    getattr(truncated_list_retriever, "paginator", None), PaginatorTestReadDecorator
)
```

This does re-expose round 1's P2 about a custom retriever's manifest-declared `name`/`primary_key` being overridden — but that was a minor nit and all three precedent call sites share the property. An unusable arm is strictly worse than an overridden `name`; if the manifest values must survive, the fix belongs in `create_custom_component` file-wide rather than in a one-off branch here.

Please also change the test fixture so `$parameters` does not supply `name`, otherwise the regression stays invisible.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,42 @@ def create_record_expander(
         config: Config,
         **kwargs: Any,
     ) -> RecordExpander:
+        truncated_list_retriever = None
+        suppress_incomplete_fetch_warning = bool(self._limit_pages_fetched_per_slice)
```
tolik0: **P2 — suppression stays broad for `CustomRetriever`, contradicting the schema text this PR adds, and the expression is untested**

The flag is seeded from `self._limit_pages_fetched_per_slice` and narrowed only inside the `isinstance(retriever_model, SimpleRetrieverModel)` branch below. For a `CustomRetriever` the seed is final, so in **every** Builder / manifest-server test read the incomplete-fetch warning is suppressed — including for a custom retriever with no pagination at all, which is precisely the case the new schema description promises will still warn:

> "…the incomplete-fetch warning is not emitted in test reads when a `paginator` is configured (**without one the retriever is not capped, so the warning still applies**)."

This is also the half of CodeRabbit's suppression comment that is still open: it asked to "apply the same check to `CustomRetriever` only when its constructed pagination path is capped", and that comment is marked addressed in `95b7b2a`.

Measured matrix (`limit_pages` = `None` / `5`):

| retriever | `suppress` | `.paginator` |
|---|---|---|
| SimpleRetriever, no paginator | False / **False** | `NoPagination` |
| SimpleRetriever + paginator | False / True | `DefaultPaginator` / `PaginatorTestReadDecorator` |
| CustomRetriever | False / **True** | (absent) |

End-to-end: custom retriever returning 2 records, parent `total_count: 5` — production read warns; Builder test read with `max_pages_per_slice=1` gives `incomplete-fetch WARN present? False`.

Both mutants survive (forcing the seed to `False`, and to `True`), so the expression is entirely unexercised — while its sibling `test_create_record_expander_rejects_unsupported_retriever_options` *was* correctly parametrized over both retriever types.

**Fix**: seed `False` and set it only on the `SimpleRetrieverModel` arm — a retriever the CDK cannot cap should keep warning. Or, if the broad behaviour is intended, narrow the schema sentence to `SimpleRetriever`. Either way it needs the two parametrized tests.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,42 @@ def create_record_expander(
...
+                    config=config,
+                    name=name,
+                    primary_key=None,
+                    transformations=[],
+                    log_formatter=log_formatter,
+                )
+                # Only a capped paginator makes a shortfall expected; `NoPagination` is never capped.
+                suppress_incomplete_fetch_warning = isinstance(
+                    truncated_list_retriever.paginator, PaginatorTestReadDecorator
+                )
```
tolik0: **P2 — suppression is decided at build time from the paginator's *type*, not from whether the cap was actually hit**

`isinstance(..., PaginatorTestReadDecorator)` is evaluated once at construction. `PaginatorTestReadDecorator` already tracks whether the cap was reached (`_page_count` vs `_maximum_number_of_pages`, `default_paginator.py:237,244,251`) and that information is discarded here.

Failure scenario: Builder test read with `max_pages_per_slice=5`, and the complete-list endpoint pointed at the wrong path so it returns one page of 2 records with `has_more: false` while `total_count` is 500. The cap was never hit; the shortfall is a real config bug the user needs to see. With `suppress=True` there is no warning, and they ship a connector that silently loses 498 items per parent record.

**Fix**: expose something like `PaginatorTestReadDecorator.reached_limit` and skip the warning only when the cap actually bit on the fetch that fell short.

airbyte_cdk/sources/declarative/expanders/record_expander.py
```diff
@@ -89,37 +181,153 @@ def expand_record(self, record: MutableMapping[Any, Any]) -> Iterable[MutableMap
...
+        )
+
+    def _warn_if_fetch_incomplete(
+        self, parent_record: Mapping[str, Any], expand_path: list[Any], fetched_count: int
+    ) -> None:
+        if self.suppress_incomplete_fetch_warning:
+            return
+        indicator_path = self._evaluated_indicator_path()
+        total_count = self._get_sibling_total_count(parent_record, indicator_path)
+        if total_count is None or fetched_count >= total_count:
```
tolik0, line 257: **P2 — a zero-record fetch is still silent when the parent record has no `total_count` sibling**

`ebab4e75` is titled "warn on empty fetch", and it does — but only when an integer `total_count` sibling exists. With the indicator truthy, a retriever configured, and the complete-list endpoint returning `{"data": []}`, this line returns early, `_warn_truncated_without_retriever` is unreachable (a retriever *is* configured), and `expand_record` falls through and emits the **truncated** embedded list. The `fallback_fragment` branch for `fetched_count == 0` is dead in that case.

Verified on a production read — indicator `has_more: true`, no `total_count`, nested endpoint returns `{"data": []}`:

```
records: ['il_1']  WARNs: []
```

**Fix**: warn unconditionally when `fetched_count == 0` while the indicator is truthy — that state is always either a misconfiguration or a fallback — and keep the `total_count` comparison for the partial case.

Related: `test_no_incomplete_fetch_warning_when_count_matches_or_total_unknown` parametrizes `total_count=True`, but since the retriever returns 15 records the real code (`None`, no warning) and the `isinstance(total, bool)` mutant (`total_count=1`, `15 >= 1`, no warning) agree, so that guard's mutant survives. A case with `fetched_count == 0` kills it.

tolik0 (commented):
Re-review at `95b7b2ac` (round 1 was `87ab6325`). Verdict: **fix before merge — one blocking regression, one line.** Still **non-breaking**; `feat(low-code):` is the right type.

All 5 round-1 P1s and 6 of 9 P2s are genuinely resolved, and I checked each by running code at this head rather than reading the commit messages: the round-1 silent-loss repro now warns (`records=10, warnings=1`), all four glob bypasses are rejected at construction, Builder `auxiliary_requests` went 0 → 2, and the hand-edited `class RecordExpander` is byte-identical to pinned `datamodel_code_generator==0.26.3` output. The expander suite went 12 → 44 with a 71% mutation kill rate on the new behaviour. Thank you for the two self-caught gaps (`CustomRetrieverModel` being `Extra.allow`, and the warning double-emission) and for escalating the three hard questions instead of silently deciding them.

One item blocks, and it is an own-goal from fixing my own round-1 nit: splitting the factory arms dropped `name`/`primary_key`/`transformations` from the `CustomRetriever` arm, so a `CustomRetriever` with any nested component no longer builds. Its test passes only because the shared fixture injects `$parameters: {name: "lists"}`.

Four non-blocking P2s are inline below, plus a reply in the concurrency thread. Two of them compound: in the manifest-server path a test read can truncate a fetch *and* suppress the warning about it.

Also worth relaying, not a CDK finding: airbytehq/airbyte#85087 bumps `source-stripe` to `6.0.16`, which #85672 already took — it needs a rebase to `6.0.17`.

A fresh `/prerelease` plus a pinned `source-stripe` run is the outstanding fleet validation: the round-1 prerelease predates all 10 commits in this round, and none of the six `Test Connectors` connectors exercise `RecordExpander`.

unit_tests/sources/declarative/expanders/test_record_expander.py
```diff
@@ -0,0 +1,574 @@
...
+    expander = _retriever_expander(_make_retriever([]))
+
+    with caplog.at_level("WARNING", logger="airbyte"):
+        records = list(expander.expand_record(_event(embedded, has_more=True, total_count=None)))
+
+    assert len(records) == 10
+    assert not [r for r in caplog.records if r.levelname == "WARNING"]
+
+
+def test_warning_is_emitted_once_across_concurrent_expansions():
```
tolik0, line 436: **P2 — this concurrency test is decorative: it passes with the lock removed**

I replaced `self._warning_lock = threading.Lock()` (`record_expander.py:151`) with `contextlib.nullcontext()` and the whole suite still passes:

```
151:        self._warning_lock = __import__("contextlib").nullcontext()
======================= 44 passed, 19 warnings in 0.56s ========================
```

The barrier synchronises thread *start*; the window the lock protects is the read-then-write of the flag at `:234-236`, two attribute accesses the GIL will not preempt in practice. A hardened variant (16 threads, `sys.setswitchinterval(1e-9)`) also failed to kill it, 0 for 5.

The lock is *correct* and worth keeping — the point is that nothing would notice if someone removed it, and `_warned_incomplete_fetch` has no concurrency test at all.

**Fix**: keep this as a smoke test and add a white-box one that forces the window open — a data descriptor on the flag whose `__set__` sleeps (verified: passes on HEAD, fails `assert 4 == 1` under `nullcontext`). Minimum acceptable substitute is `assert isinstance(expander._warning_lock, type(threading.Lock()))`.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -2506,6 +2506,38 @@ def create_record_expander(
...
+                if retriever_fields.get(unsupported_option):
+                    raise ValueError(
+                        f"`{unsupported_option}` is not supported on `truncated_list_retriever`."
+                    )
+            if isinstance(retriever_model, SimpleRetrieverModel):
+                truncated_list_retriever = self._create_component_from_model(
+                    model=retriever_model,
+                    config=config,
+                    name=name,
+                    primary_key=None,
```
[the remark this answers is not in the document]
tolik0: **P2 — the rationale for declining does not hold: the manifest server is not single-threaded.**

The first half is right, and `connector_builder_handler.py:73-76` does force `concurrency_level.default_concurrency = 1` unconditionally. But `create_source` is not the only caller. `manifest_server/command_processor/utils.py:77-93` (`build_source`) constructs `ConcurrentDeclarativeSource(emit_connector_builder_messages=True, limits=TestLimits(...))` and **never touches `concurrency_level`**; `routers/manifest.py:109` → `command_processor/processor.py:42-48` then runs `TestReader.run_test_read` on it. The manifest server is the production Connector Builder backend, so that is the path most test reads actually take.

Measured, with a manifest declaring `default_concurrency: 10`:

```
manifest_server build_source     -> workers: 10 ; emit_connector_builder_messages: True ; limit_pages_fetched_per_slice: 5
connector_builder create_source  -> workers: 1
```

So there the nested retriever *is* wrapped in `PaginatorTestReadDecorator` — whose own docstring (`default_paginator.py:231-233`) says "This decorator is not currently thread-safe" — and the single shared `RecordExpander` (`stream_slicers/declarative_partition_generator.py:50`, `:59`) hands one nested retriever and one paginator to up to 10 threads. Reproduced against the real class with cap = 3:

```
A page2 ✓  A page3 ✓  A page4 None
B page2 ✓  B page3 None   <-- B stopped after 2 pages though the cap is 3
```

Partition B's fetch is silently cut short because A's `get_initial_token()` reset the shared `_page_count`.

On scope, you are right that this is pre-existing: the stream's own retriever and paginator are already shared this way in the manifest server, and this PR adds one more instance of the same hazard rather than creating it. I am not asking for a per-fetch clone either — your objection about duplicating the session and defeating `use_cache` stands.

What I am flagging is narrower and I think still actionable:

1. The stated reason is wrong, so the conclusion "cannot occur under the current Builder wiring" does not follow.
2. The new schema description now asserts something untrue about test reads.
3. It compounds with the suppression finding I left on the seed line above: in exactly this configuration `suppress_incomplete_fetch_warning` is `True`, so the resulting shortfall is also **silent**.

Either have the manifest-server path force `default_concurrency=1` the way `create_source` does, or stop claiming test reads are single-threaded. Fixing the suppression finding independently removes the silence, which is the part that actually costs a user data.

I did not drive this through a live manifest-server HTTP request — the race is reproduced against the real `PaginatorTestReadDecorator` and the wiring that makes it reachable was read and confirmed.

airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
tolik0: Reopening this one: the fix for it caused a regression. Splitting the arms so the `SimpleRetriever` branch keeps `name`/`primary_key`/`transformations` left the `CustomRetriever` branch with only `model`/`config`/`log_formatter`, and those kwargs are what `_create_nested_component` feeds to the custom retriever's *sub*-components — so a `CustomRetriever` with a nested requester or selector no longer builds at all. Details, repro and the one-line fix are in the P1 comment on `:2542-2544`.

Dropping the dead `stream_slicer=None` was right and should stay.

tolik0 (approved):
Approving at `4332e3c2`.

`4332e3c2` is the fix I asked for, taken whole, and it closes round 2's P1 and P2-1 together: the `isinstance` branch is collapsed back into one `_create_component_from_model` call passing `name`/`primary_key`/`transformations` for both retriever types, the suppression seed drops from `bool(self._limit_pages_fetched_per_slice)` to `False`, and the fixture's `$parameters: {name: "lists"}` injection — the thing that masked the regression — is gone, with the assertions strengthened to check the nested `requester` and `record_selector` were actually built.

Verified by running it, not by reading the commit message:

- A genuinely non-`SimpleRetriever` custom retriever now builds with a nested `RecordSelector` and with a nested `SimpleRetriever`: `BUILD OK -> MyLinesRetriever  name='record_expander_truncated_list'` in both cases, with no `$parameters` anywhere. That distinction matters because the in-tree `TestingCustomRetriever` *subclasses* `SimpleRetriever`, so the repo's own custom-retriever tests never leave the `SimpleRetriever` path.
- Suppression matrix now matches the intent: `CustomRetriever` → `False` at `limit_pages` both `None` and `5`; `SimpleRetriever` → `True` only at `limit_pages=5` **and** with a paginator, where it is a `PaginatorTestReadDecorator`, `False` with `NoPagination`.
- The regression is genuinely covered now: dropping `primary_key`/`transformations` from the single call fails 5 tests, dropping all three (the exact round-2 state) fails 8. Under round 2 the equivalent change failed nothing.
- CI at this head: every real gate green, including `CodeRabbit`, which was rate-limited last round, and `Check: source-shopify` at 1h4m. `Check: destination-motherduck` is the only red and is the same environmental MotherDuck-token failure seen on all 12 PRs sampled. Local sweep 313 passed / 1 failed, and that failure (`test_full_resolve_manifest`) is also red at the merge base.

Still non-breaking, and `feat(low-code):` is the right type — both fields are optional with `None` defaults and the unset path is byte-identical to base.

**Not blocking, your call whether any of it rides along.** Four P2s stay open from round 2, and none touches a production sync path of a shipped connector: build-time suppression decided from the paginator's type rather than whether the cap actually bit; the zero-record fetch that is still silent when the record has no `total_count` sibling; the manifest-server concurrency rationale; and the decorative concurrency test, which I re-checked at this head and it still passes with the lock replaced by `nullcontext()`. The zero-record one does not reach `source-stripe` — your live deleted-invoice test saw the warning fire on that path, which means Stripe's embedded `lines` object does carry the `total_count` sibling that finding depends on being absent.

Two one-liners if you want them in this PR rather than a follow-up:

1. Warn unconditionally when `fetched_count == 0` while the indicator is truthy, keeping the `total_count` comparison for the partial case.
2. Qualify the schema sentence about test-read suppression to `SimpleRetriever`. After `4332e3c2` a `CustomRetriever` always keeps the warning, so the current wording is now wrong for that arm. Copilot flagged the same mismatch on `:2542` and proposed the opposite cure — reverting to `bool(self._limit_pages_fetched_per_slice)` for opaque custom retrievers. I would not: a spurious warning in a test read costs a log line, a suppressed real shortfall costs the user data silently, and that asymmetry is why the broad seed was a finding in the first place.

One new low-severity note for whoever touches this next: the new `isinstance(truncated_list_retriever, SimpleRetriever)` guard is untested — dropping it entirely leaves all 211 tests green, because `TestingCustomRetriever` subclasses `SimpleRetriever`, so the `custom_retriever` parametrization added here never exercises the non-`SimpleRetriever` path. The behaviour is correct; nothing in the suite pins it. Relatedly, the seed at `:2510` is now dead for every path that consults the flag, so forcing it to `True` is an equivalent mutant rather than a coverage gap.

Two things to line up outside this PR: the round-1 `/prerelease` predates all 11 commits since, so a fresh one plus a pinned `source-stripe` run is the fleet validation I would want before the connector release — none of the six `Test Connectors` connectors exercise `RecordExpander`. And airbytehq/airbyte#85087 still bumps to `6.0.16`, which #85672 already took, so it needs a rebase to `6.0.17`.

Thanks for the turnaround on this one, and for engaging with every finding on the merits across three rounds.

PR #1138, opened 2026-08-28
fix(oauth): preserve provider error detail on refresh token failure

tolik0:
## What

When an OAuth refresh request is rejected, `AbstractOauth2Authenticator._make_handled_request` raised an `AirbyteTracedException` whose user-facing message was one fixed sentence, and it threw away the provider's own diagnostic. `_wrap_refresh_token_exception` had already parsed the error body with `exception.response.json()` in order to decide whether the failure was a refresh-token failure, and then discarded that parsed body.

That matters most for Microsoft Entra, where the body carries an AADSTS code that separates root causes needing completely different fixes:

- `AADSTS50173` — the grant was revoked, typically because the user changed or reset their password. Re-authenticating is the fix.
- `AADSTS7000218` / `AADSTS700025` — the app registration's client type or client credential is misconfigured. Re-authenticating does not help; the app registration has to change.
- `AADSTS50076` / `AADSTS50078` / `AADSTS700082` — Conditional Access requires an interactive sign-in or MFA. Again a different fix, and one the workspace admin has to make.

Today all three collapse into the same string. Across 300 sampled production failure summaries for source-bing-ads, zero contain an AADSTS code, so support has no way to tell these apart from the failure summary. See airbytehq/oncall#12835.

## How

- The error body is parsed once, in `_make_handled_request`, and passed into `_wrap_refresh_token_exception` through a new optional `response_content` argument, so the response is no longer parsed twice. The argument defaults to `None` and the method parses on demand when it is absent, so existing callers keep working.
- `internal_message` carries the full provider response (`HTTP <status>: <body>`, truncated at 1000 characters), so the whole payload lands in the logs.
- The user-facing `message` deliberately keeps the existing actionable sentence as its lead: "Refresh token was rejected by the OAuth provider (invalid, expired, or already used). Re-authenticate this source's credentials in its connection settings." Only after that is a short provider detail appended, as `Provider error: <error>: <error_description>`, built from the standard OAuth 2.0 `error` and `error_description` fields, collapsed to a single line and truncated at 200 characters. Two hundred characters is enough to keep the `AADSTS<code>` and the beginning of its description, since Entra puts the code at the front of `error_description`, while staying short enough that the failure summary is still readable. No raw provider blob becomes the primary message, and when the body has no usable `error` / `error_description` the message is byte-for-byte what it was before.
- Bodies that are empty, non-JSON, or valid JSON but not an object now resolve to "no parsed content" rather than raising. Previously a JSON array body would have hit `AttributeError` on `.get(...)` inside the already-failing error path.
- Everything surfaced, in both the user-facing and the internal message, is run through `filter_secrets`, and the authenticator's own refresh token and client secret are redacted explicitly on top of that, in case a provider echoes submitted credentials back in its payload. Only response bodies are read, so request headers, including `Authorization`, are never echoed.

No changelog entry or version bump is included: `CONTRIBUTING.md` states releases are drafted automatically by `semantic-pr-release-drafter` from the PR title, and the package version is computed by `poetry-dynamic-versioning`. `CHANGELOG.md` is frozen and points at GitHub Release Notes.

## Test plan

New tests in `unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py`:

- a parametrized test over the three Entra failure modes above, asserting that the AADSTS code and the full response body reach `internal_message` while `message` still starts with the re-authenticate guidance and then carries the provider code on a single line;
- truncation of an oversized `error_description`;
- redaction, using a payload that echoes the refresh token and client secret back;
- empty, HTML, and JSON-array bodies falling back to the raw `RequestException` instead of raising something new;
- a provider whose body has no `error` / `error_description`, asserting the user-facing message is exactly the unchanged sentence.

The existing `test_refresh_access_token_wrapped` assertion on `message` was relaxed from equality to `startswith`, since the wrapped case now appends `Provider error: invalid_grant`.

This repo is Poetry-managed, so `uv run pytest` cannot resolve the dev dependencies; the tests were run with the project's Poetry virtualenv:

```
$ python -m pytest unit_tests/sources/streams/http/requests_native_auth/ -q
======================== 49 passed, 3 warnings in 0.86s ========================

$ python -m pytest unit_tests/sources/declarative/auth -q
======================= 166 passed, 3 warnings in 34.80s =======================

$ ruff check airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
All checks passed!

$ ruff format --check airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
2 files already formatted

$ mypy --config-file mypy.ini airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
Success: no issues found in 1 source file
```

---

## Update — 2026-09-03 (added by @ZaneHyattAB on @tolik0's behalf)

@tolik0 is out, and this PR now carries commit `6d5e37c9` applying
@bazarnov's review feedback. **Everything above is @tolik0's original text,
left as written.** The points below supersede it where they conflict; the
original wording is kept for context rather than rewritten.

1. **The user-facing detail is now the first line of `error_description`
   only, capped at 120 characters** (was: single-line collapse, 200
   characters). Entra appends `Trace ID`, `Correlation ID` and `Timestamp` on
   later lines and embeds per-request issue dates in the first sentence, so
   collapsing whitespace pulled per-request GUIDs into the message. Every
   attempt then produced a distinct ~375-character string, which defeats
   grouping by failure summary — the capability this PR exists to provide.
   After the change the message is a stable ~295 characters per error code,
   with the full body still in `internal_message`.

2. **`AADSTS700082` is regrouped.** It is
   `ExpiredOrRevokedGrantInactiveToken` — refresh-token expiry after
   inactivity — so it belongs with the expired/revoked group above, not with
   Conditional Access.

3. **Reachability caveat added.** On the declarative defaults
   `source-bing-ads` runs (`(400,)` / `"error"` /
   `("invalid_grant", "invalid_permissions")`), `invalid_client`
   (`AADSTS7000218`, `AADSTS700025`) and `interaction_required`
   (`AADSTS50078`) never reach this path and surface as raw HTTP errors.

4. **Only string values are read**, so a provider nesting the error object no
   longer renders a dict repr into the message.

5. **The redaction test now actually tests the redaction.** It uses distinct
   credential values and calls `update_secrets([])` first, because secrets
   registered by earlier tests were masking the credentials on their own —
   the previous test passed even with the explicit pair redaction disabled.
   Verified by disabling that redaction and confirming the test fails.

6. **New declarative-authenticator test** at
   `unit_tests/sources/declarative/auth/test_oauth.py`, covering the path
   manifest-only connectors such as `source-bing-ads` actually run.

**Retracted:** the "300 sampled production failure summaries" figure above is
not sourced in airbytehq/oncall#12835 and should not be relied on. The
verified claim is narrower: a deliberately revoked Bing Ads token on `3.0.7`
produced a 95-line `check` log containing **zero** AADSTS codes, and the same
token on a preview build of this PR surfaced the code.

Validation on `6d5e37c9`: 216 targeted tests pass, `ruff` and `mypy` clean.
Additionally validated end to end in Airbyte Cloud on a genuinely revoked
token — see airbytehq/oncall#12835.

@tolik0, please adjust or fold this in as you see fit when you're back.

## Update — 2026-09-09: provider detail is now the code only (`1216c9d2`)

Supersedes the `<error>: <error_description>` format in "How" above (single line, 200 characters) and point 1 of the 2026-09-03 update (first line, 120 characters). Neither was deterministic for every Entra code: `AADSTS700082` carries the token issue timestamp in its *first sentence*, so no line rule or character cap can make it stable ([analysis](https://github.com/airbytehq/airbyte-python-cdk/pull/1138#discussion_r3936030689)).

What ships in `1216c9d2`:

- The user-facing detail is `Provider error: <error>[: <provider code>]`, where the code is `^[A-Za-z]{3,}\d{4,}\b` matched against the start of `error_description` (e.g. `AADSTS700082`). No description prose is ever appended, so the same failure produces a byte-identical message on every attempt ([rationale](https://github.com/airbytehq/airbyte-python-cdk/blob/1216c9d2a9ce322d52799e722790afcb01721ef0/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L299-L313)).
- Providers whose `error_description` is prose (Google, Salesforce, GitLab, Airtable, Harvest, ...) get `Provider error: <error>` only; the full body is still in `internal_message`, now redacted as well.
- `_PROVIDER_ERROR_DETAIL_MAX_LENGTH` is 64 (both `error` and the code are provider-controlled). The cap test asserts the literal length so the constant is pinned.
- New tests: first-sentence timestamp with no later lines, `\r\n` trace id under the cap, the same code with two different issue dates asserting equal `message`, and a word-boundary check for the code pattern.

Residual theoretical case, no change requested: a provider whose `error_description` leads with a hex/GUID-like token would surface that token as the "code"; none of the providers with fixtures in the monorepo does, and the value is still redacted and capped.

Link to Devin session: https://app.devin.ai/sessions/51216c43280840fba9342c4cec3cd435
Open in Devin Desktop: https://app.devin.ai/desktop/session/51216c43280840fba9342c4cec3cd435?variant=devin

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth refresh-token error messages by showing concise provider error codes and relevant details.
  - Redacted refresh tokens, client secrets, and other credentials from both user-facing and diagnostic messages.
  - Prevented lengthy provider responses, trace IDs, and timestamps from cluttering user-facing errors.
  - Improved handling of non-JSON error responses by preserving the underlying request error.

### DIFF
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -27,8 +27,10 @@ _NOOP_MESSAGE_REPOSITORY = NoopMessageRepository()
 
 # How much of the provider's error detail is appended to the user-facing message. Long enough to
 # keep a provider error code and the start of its description (e.g. Microsoft Entra puts the
-# `AADSTS<code>` at the front of `error_description`), short enough to keep the message readable.
-_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 200
+# `AADSTS<code>` at the front of `error_description`), short enough to keep the message readable
+# and to stop before the per-request values (issue dates, trace ids) that some providers embed
+# further into the description, so the same failure produces the same message on every attempt.
+_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 120
 # How much of the raw provider response is kept in the internal message, which is logged and is not
 # shown to the user.
 _PROVIDER_ERROR_RESPONSE_MAX_LENGTH = 1000
@@ -295,24 +297,27 @@ class AbstractOauth2Authenticator(AuthBase):
         """
         Build a short, single-line provider error detail suitable for the user-facing message.
 
-        Only the standard OAuth 2.0 error fields (`error` and `error_description`) are used: they
-        are where providers put the actionable diagnostic, and they are not expected to carry
-        credentials. For Microsoft Entra this is what distinguishes a revoked grant
-        (`AADSTS50173`) from a client-type or secret misconfiguration (`AADSTS7000218`,
-        `AADSTS700025`) from Conditional Access requiring interactive sign-in (`AADSTS50076`,
-        `AADSTS50078`, `AADSTS700082`) - all of which otherwise collapse into the same message.
+        Only the standard OAuth 2.0 `error` and `error_description` fields (RFC 6749 section 5.2)
+        are used, and only the first line of the description: providers put the actionable code
+        there (Microsoft Entra leads with `AADSTS<code>`, which tells a revoked or expired grant
+        such as `AADSTS50173` / `AADSTS700082` apart from a client misconfiguration such as
+        `AADSTS7000218`), while per-request trace ids and timestamps follow on later lines. Which
+        provider errors reach this path at all is set by the authenticator's `refresh_token_error_*`
+        configuration.
         """
         if not response_content:
             return None
         parts = [
-            str(response_content[key])
+            response_content[key].strip()
             for key in ("error", "error_description")
-            if response_content.get(key)
+            if isinstance(response_content.get(key), str) and response_content[key].strip()
         ]
         if not parts:
             return None
-        # Collapse newlines and repeated whitespace so the detail stays on a single line.
-        detail = " ".join(": ".join(parts).split())
+        # Keep the first line only and collapse its whitespace: the actionable code leads the
+        # description, while trace ids and timestamps that differ on every attempt follow on later
+        # lines and would make the same failure read differently each time.
+        detail = " ".join(": ".join(parts).splitlines()[0].split())
         return self._truncate(self._redact_credentials(detail), _PROVIDER_ERROR_DETAIL_MAX_LENGTH)
 
     def _build_provider_response_info(self, exception: requests.exceptions.RequestException) -> str:
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -17,13 +17,21 @@ from airbyte_cdk.models import FailureType, Level
 from airbyte_cdk.sources.http_logger import format_http_message
 from airbyte_cdk.sources.message import MessageRepository, NoopMessageRepository
 from airbyte_cdk.utils import AirbyteTracedException
-from airbyte_cdk.utils.airbyte_secrets_utils import add_to_secrets
+from airbyte_cdk.utils.airbyte_secrets_utils import add_to_secrets, filter_secrets
 from airbyte_cdk.utils.datetime_helpers import AirbyteDateTime, ab_datetime_now, ab_datetime_parse
 
 from ..exceptions import DefaultBackoffException
 
 logger = logging.getLogger("airbyte")
 _NOOP_MESSAGE_REPOSITORY = NoopMessageRepository()
+
+# How much of the provider's error detail is appended to the user-facing message. Long enough to
+# keep a provider error code and the start of its description (e.g. Microsoft Entra puts the
+# `AADSTS<code>` at the front of `error_description`), short enough to keep the message readable.
+_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 200
+# How much of the raw provider response is kept in the internal message, which is logged and is not
+# shown to the user.
+_PROVIDER_ERROR_RESPONSE_MAX_LENGTH = 1000
 
 
 class ResponseKeysMaxRecurtionReached(AirbyteTracedException):
@@ -238,8 +246,91 @@ class AbstractOauth2Authenticator(AuthBase):
         default_token_expiry_duration_hours = 1  # 1 hour
         return ab_datetime_now() + timedelta(hours=default_token_expiry_duration_hours)
 
+    @staticmethod
+    def _parse_error_response_content(
+        response: Optional[requests.Response],
+    ) -> Optional[Mapping[str, Any]]:
+        """
+        Best-effort parse of an error response body as a JSON object.
+
+        Returns `None` when the response is missing, empty, not valid JSON, or not a JSON object,
+        so that callers can degrade gracefully instead of raising a new exception while they are
+        already handling an error.
+        """
+        if response is None:
+            return None
+        try:
+            content = response.json()
+        except (JSONDecodeError, ValueError):
+            return None
+        return content if isinstance(content, Mapping) else None
+
+    def _redact_credentials(self, value: str) -> str:
+        """
+        Redact credential material from a string before it is logged or surfaced to the user.
+
+        Only response bodies are passed here, so request headers (including `Authorization`) are
+        never echoed. On top of the config secrets already tracked by the CDK, the authenticator's
+        own refresh token and client secret are redacted explicitly, in case a provider echoes the
+        submitted credentials back in its error payload.
+        """
+        redacted = filter_secrets(value)
+        for get_credential in (self.get_refresh_token, self.get_client_secret):
+            try:
+                credential = get_credential()
+            except Exception:
+                # Never let redaction itself fail the error path we are already in.
+                continue
+            if credential and isinstance(credential, str):
+                redacted = redacted.replace(credential, "****")
+        return redacted
+
+    @staticmethod
+    def _truncate(value: str, max_length: int) -> str:
+        return value if len(value) <= max_length else value[:max_length] + "..."
+
+    def _build_provider_error_detail(
+        self, response_content: Optional[Mapping[str, Any]]
+    ) -> Optional[str]:
+        """
+        Build a short, single-line provider error detail suitable for the user-facing message.
+
+        Only the standard OAuth 2.0 error fields (`error` and `error_description`) are used: they
+        are where providers put the actionable diagnostic, and they are not expected to carry
+        credentials. For Microsoft Entra this is what distinguishes a revoked grant
+        (`AADSTS50173`) from a client-type or secret misconfiguration (`AADSTS7000218`,
+        `AADSTS700025`) from Conditional Access requiring interactive sign-in (`AADSTS50076`,
+        `AADSTS50078`, `AADSTS700082`) - all of which otherwise collapse into the same message.
+        """
+        if not response_content:
+            return None
+        parts = [
+            str(response_content[key])
+            for key in ("error", "error_description")
+            if response_content.get(key)
+        ]
+        if not parts:
+            return None
+        # Collapse newlines and repeated whitespace so the detail stays on a single line.
+        detail = " ".join(": ".join(parts).split())
+        return self._truncate(self._redact_credentials(detail), _PROVIDER_ERROR_DETAIL_MAX_LENGTH)
+
+    def _build_provider_response_info(self, exception: requests.exceptions.RequestException) -> str:
+        """
+        Build the full provider response detail for the internal message, which goes to the logs.
+        """
+        if exception.response is None:
+            return self._redact_credentials(str(exception))
+        body = self._truncate(
+            self._redact_credentials(exception.response.text),
+            _PROVIDER_ERROR_RESPONSE_MAX_LENGTH,
+        )
+        return f"HTTP {exception.response.status_code}: {body}"
+
     def _wrap_refresh_token_exception(
-        self, exception: requests.exceptions.RequestException
+        self,
+        exception: requests.exceptions.RequestException,
+        response_content: Optional[Mapping[str, Any]] = None,
     ) -> bool:
         """
         Wraps and handles exceptions that occur during the refresh token process.
@@ -249,16 +340,20 @@ class AbstractOauth2Authenticator(AuthBase):
 
         Args:
             exception (requests.exceptions.RequestException): The exception raised during the request.
+            response_content (Optional[Mapping[str, Any]]): The already-parsed response body, when
+                the caller has one, so the body is not parsed twice. Parsed on demand otherwise.
 
         Returns:
             bool: True if the exception is related to a refresh token error, False otherwise.
         """
-        try:
-            if exception.response is not None:
-                exception_content = exception.response.json()
-            else:
-                return False
-        except JSONDecodeError:
+        if exception.response is None:
+            return False
+        exception_content = (
+            response_content
+            if response_content is not None
+            else self._parse_error_response_content(exception.response)
+        )
+        if exception_content is None:
             return False
         return (
             exception.response.status_code in self._refresh_token_error_status_codes
@@ -333,15 +428,25 @@ class AbstractOauth2Authenticator(AuthBase):
                         response=e.response,
                         failure_type=FailureType.transient_error,
                     )
-            if self._wrap_refresh_token_exception(e):
-                response_info = (
-                    f"HTTP {e.response.status_code}: {e.response.text[:1000]}"
-                    if e.response is not None
-                    else str(e)
+            error_content = self._parse_error_response_content(e.response)
+            if self._wrap_refresh_token_exception(e, response_content=error_content):
+                message = (
+                    "Refresh token was rejected by the OAuth provider (invalid, expired, or "
+                    "already used). Re-authenticate this source's credentials in its connection "
+                    "settings."
                 )
+                provider_error_detail = self._build_provider_error_detail(error_content)
+                if provider_error_detail:
+                    # The provider's own diagnostic is what tells apart otherwise identical-looking
+                    # failures (revoked grant vs. misconfigured client vs. Conditional Access), so
+                    # a short form of it is appended after the actionable guidance.
+                    message = f"{message} Provider error: {provider_error_detail}"
                 raise AirbyteTracedException(
-                    internal_message=f"Refresh token rejected by the OAuth token endpoint. {response_info}",
-                    message="Refresh token was rejected by the OAuth provider (invalid, expired, or already used). Re-authenticate this source's credentials in its connection settings.",
+                    internal_message=(
+                        "Refresh token rejected by the OAuth token endpoint. "
+                        f"{self._build_provider_response_info(e)}"
+                    ),
+                    message=message,
                     failure_type=FailureType.config_error,
                 ) from e
             raise
--- a/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
+++ b/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
@@ -3,6 +3,7 @@
 #
 
 import logging
+import re
 import threading
 from abc import abstractmethod
 from datetime import timedelta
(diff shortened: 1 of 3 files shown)


ZaneHyattAB:
/prerelease
> **Prerelease Job Info**
>
>

[1]: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/33676749491
>
> View the publish workflow run: https://github.com/airbytehq/airbyte-python-cdk/actions/runs/33676771546

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -17,14 +17,22 @@
...
 
 from ..exceptions import DefaultBackoffException
 
 logger = logging.getLogger("airbyte")
 _NOOP_MESSAGE_REPOSITORY = NoopMessageRepository()
 
+# How much of the provider's error detail is appended to the user-facing message. Long enough to
+# keep a provider error code and the start of its description (e.g. Microsoft Entra puts the
+# `AADSTS<code>` at the front of `error_description`), short enough to keep the message readable.
+_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 200
```
bazarnov: With the first-line rule suggested below, 120 still keeps the whole `AADSTS50173` sentence (113 characters with the `invalid_grant: ` prefix) but stops before the two issue timestamps Entra puts later on that same line, so the message is identical for every occurrence of a code.

```suggestion
# How much of the provider's error detail is appended to the user-facing message. Long enough to
# keep a provider error code and the start of its description (e.g. Microsoft Entra puts the
# `AADSTS<code>` at the front of `error_description`), short enough to keep the message readable
# and to stop before the per-request values (issue dates, trace ids) that some providers embed
# further into the description, so the same failure produces the same message on every attempt.
_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 120
```

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -238,8 +246,91 @@ def _default_token_expiry_date(self) -> AirbyteDateTime:
...
+    ) -> Optional[str]:
+        """
+        Build a short, single-line provider error detail suitable for the user-facing message.
+
+        Only the standard OAuth 2.0 error fields (`error` and `error_description`) are used: they
+        are where providers put the actionable diagnostic, and they are not expected to carry
+        credentials. For Microsoft Entra this is what distinguishes a revoked grant
+        (`AADSTS50173`) from a client-type or secret misconfiguration (`AADSTS7000218`,
+        `AADSTS700025`) from Conditional Access requiring interactive sign-in (`AADSTS50076`,
+        `AADSTS50078`, `AADSTS700082`) - all of which otherwise collapse into the same message.
```
bazarnov: `AADSTS700082` is `ExpiredOrRevokedGrantInactiveToken` - the refresh token expired after inactivity ([reference-error-codes](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes)), so it belongs with the expired/revoked group, not with Conditional Access (the PR description has the same grouping). Also worth stating that reachability depends on `refresh_token_error_*`: on the declarative defaults `(400,) / "error" / ("invalid_grant", "invalid_permissions")` that source-bing-ads runs on, `invalid_client` (AADSTS7000218, AADSTS700025) and `interaction_required` (AADSTS50078) never get here.

```suggestion
        Only the standard OAuth 2.0 `error` and `error_description` fields (RFC 6749 section 5.2)
        are used, and only the first line of the description: providers put the actionable code
        there (Microsoft Entra leads with `AADSTS<code>`, which tells a revoked or expired grant
        such as `AADSTS50173` / `AADSTS700082` apart from a client misconfiguration such as
        `AADSTS7000218`), while per-request trace ids and timestamps follow on later lines. Which
        provider errors reach this path at all is set by the authenticator's `refresh_token_error_*`
        configuration.
```

bazarnov (commented):
Thanks for the change, and for the end-to-end validation on a revoked Bing Ads token. Review only, nothing pushed to the branch; the inline suggestions below are validated as a whole on commit de26718c360bd1ef00844d4a0fd2188093a48404 (merge of `main` + the suggested edits, CI run on https://github.com/airbytehq/airbyte-python-cdk/pull/1144).

Verified:
- Adequacy: source-bing-ads sets no `refresh_token_error_*`, so it runs on the declarative defaults `(400,) / "error" / ("invalid_grant", "invalid_permissions")` ([model_to_component_factory.py#L3068](https://github.com/airbytehq/airbyte-python-cdk/blob/main/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L3068)), which is exactly the path Microsoft's `400 invalid_grant` with `AADSTS50173` takes. The code reaches the failure summary, the full body reaches `internal_message`.
- Compatibility: `_wrap_refresh_token_exception` only gains an optional kwarg; no connector in airbytehq/airbyte overrides it or `_make_handled_request`, and nothing asserts on the exact previous message. Not a breaking change.
- Redaction order is sound (`filter_secrets` first, then the authenticator's own refresh token and client secret, before truncation), and only the response body is ever read.
- The new tests are discriminating: 6 of the 9 cases fail on `main`, including the JSON-array body that previously raised `AttributeError` inside the error handler.

Suggested changes (inline):
1. Keep only the first line of `error_description` and cap at 120. Entra appends `Trace ID`, `Correlation ID`, `Timestamp` and issue dates to the description, so today the same failure produces a different user-facing message on every attempt (up to 375 characters, GUIDs inside), which defeats grouping by failure summary - the reason this PR exists. With the change it is one string per failure class, about 295 characters at most, and the full body is still in `internal_message`.
2. Docstring: `AADSTS700082` is refresh-token expiry due to inactivity, not Conditional Access ([reference-error-codes](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes)); the PR description carries the same grouping.
3. Tests: the redaction test uses credential values equal to their field names and is satisfied by a secret leaked from an earlier test (`add_to_secrets("token")`); the `trailing_marker` comment is not true for the `AADSTS7000218` fixture (181 characters, under the cap). Worth adding the declarative-authenticator path that manifest connectors actually run; a ready-made test is in the commit above (`unit_tests/sources/declarative/auth/test_oauth.py`).

Notes for merge and follow-up:
- Under the defaults, `invalid_client` (AADSTS7000218 / AADSTS700025, the latter arriving as HTTP 401) and `interaction_required` (AADSTS50078) do not enter this path for source-bing-ads; they surface as the raw HTTP error. If those need the same treatment, set `refresh_token_error_status_codes` / `refresh_token_error_values` in the connector manifest.
- source-bing-ads pins `source-declarative-manifest:7.25.1`; this fix and #1130 (`v7.28.3`) reach users only after the next CDK patch release and a base-image bump of the connector.
- The "300 sampled production failure summaries" figure in the description is not in the oncall thread; worth citing the query or softening.
- The PR is still a draft.

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -238,8 +246,91 @@ def _default_token_expiry_date(self) -> AirbyteDateTime:
...
+            return None
+        parts = [
+            str(response_content[key])
+            for key in ("error", "error_description")
+            if response_content.get(key)
+        ]
+        if not parts:
+            return None
+        # Collapse newlines and repeated whitespace so the detail stays on a single line.
+        detail = " ".join(": ".join(parts).split())
```
bazarnov: Entra's `error_description` is `AADSTS<code>: <sentence>\r\nTrace ID: <guid>\r\nCorrelation ID: <guid>\r\nTimestamp: ...` ([token endpoint error response](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#error-response-1)). Collapsing whitespace pulls those onto the user-facing line: for `AADSTS7000218` the whole detail is 181 characters, so both GUIDs land in `message`; for `AADSTS50173` the 200 cut lands after the first issue timestamp. Every attempt then carries a distinct 375-character message, which defeats grouping by failure summary (the 4,074-row count in the oncall issue keys on exactly that). First line only plus the 120 cap gives one string per code, about 295 characters in total, with the full body still in `internal_message`. The `isinstance` guard also avoids rendering a dict repr when a provider nests the error object.

```suggestion
        parts = [
            response_content[key].strip()
            for key in ("error", "error_description")
            if isinstance(response_content.get(key), str) and response_content[key].strip()
        ]
        if not parts:
            return None
        # Keep the first line only and collapse its whitespace: the actionable code leads the
        # description, while trace ids and timestamps that differ on every attempt follow on later
        # lines and would make the same failure read differently each time.
        detail = " ".join(": ".join(parts).splitlines()[0].split())
```

unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
```diff
@@ -648,9 +648,171 @@ def test_refresh_access_token_wrapped(
...
+            TestOauth2Authenticator.client_id,
+            TestOauth2Authenticator.client_secret,
+            TestOauth2Authenticator.refresh_token,
+            refresh_token_error_status_codes=(400,),
+            refresh_token_error_key="error",
+            refresh_token_error_values=("invalid_grant", "invalid_client"),
+        )
+
+    # Sits at the very end of each provider payload below, past the point where the user-facing
+    # message is truncated, so it can only be found in the internal message.
```
bazarnov: This is not true for the `AADSTS7000218` fixture: its collapsed detail is 181 characters, under the 200 cap, so the marker lands in the user-facing message as well, and nothing catches it because the test only asserts the marker is in `internal_message`. With the first-line change the comment becomes accurate; either way, `assert TestOauth2Authenticator.trailing_marker not in exc_info.value.message` next to the `internal_message` assertion pins the user-facing side (it fails on the current code, passes with the change).

```suggestion
    # Sits on a later line of each provider payload below, like the trace ids and timestamps
    # Microsoft Entra appends, so it must reach the internal message and never the user-facing one.
```

unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py
```diff
@@ -648,9 +648,171 @@ def test_refresh_access_token_wrapped(
...
+            },
+        )
+
+        with pytest.raises(AirbyteTracedException) as exc_info:
+            oauth.refresh_access_token()
+
+        for message in (exc_info.value.message, exc_info.value.internal_message):
+            assert TestOauth2Authenticator.refresh_token not in message
+            assert TestOauth2Authenticator.client_secret not in message
+            assert "****" in message
```
bazarnov: `TestOauth2Authenticator.refresh_token` is the literal `"refresh_token"` and `client_secret` is `"client_secret"`, so these assertions cannot tell a redacted value from a redacted field name. On top of that, `test_refresh_access_token_retry` runs earlier and calls `add_to_secrets("token")` on its successful refresh, after which `filter_secrets` turns `refresh_token` into `refresh_****` on its own: with the explicit pair redaction disabled, `refresh_token not in message` and `"****" in message` still pass. Distinct values plus a reset of the config-level secrets make the test exercise the redaction this PR adds (needs `from airbyte_cdk.utils.airbyte_secrets_utils import update_secrets` at the top).

```suggestion
        # Start from no config-level secrets: earlier tests register values such as "token" through
        # add_to_secrets, which would otherwise mask the credentials on their own.
        update_secrets([])
        refresh_token = "0.AXoA-rt-9f3ZqW7kP2"
        client_secret = "s3cr3t~Xyz-1Qp"
        oauth = Oauth2Authenticator(
            f"https://{TestOauth2Authenticator.refresh_endpoint}",
            TestOauth2Authenticator.client_id,
            client_secret,
            refresh_token,
            refresh_token_error_status_codes=(400,),
            refresh_token_error_key="error",
            refresh_token_error_values=("invalid_grant",),
        )
        requests_mock.post(
            f"https://{TestOauth2Authenticator.refresh_endpoint}",
            status_code=400,
            json={
                "error": "invalid_grant",
                "error_description": f"AADSTS50173: rejected rt={refresh_token} cs={client_secret}",
            },
        )

        with pytest.raises(AirbyteTracedException) as exc_info:
            oauth.refresh_access_token()

        for message in (exc_info.value.message, exc_info.value.internal_message):
            assert refresh_token not in message
            assert client_secret not in message
            assert message.count("****") == 2
```

airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
tolik0: Following up in this thread rather than opening a new one, because the resolution above holds for `AADSTS50173` but not for every code the docstring now groups with it.

**`AADSTS50173` is safe**, and the arithmetic above checks out. In `f"{error}: {error_description}"` with `error="invalid_grant"`, the first per-request value (the `'<issued-on>'` timestamp) starts at offset **191**, so the 120 cut lands mid-word at `...a fresh auth token is needed. The us...` — identical on every attempt.

**`AADSTS700082` is not.** Entra puts the issue timestamp in the *first sentence* of that description, not on a later line:

> `AADSTS700082: The refresh token has expired due to inactivity. The token was issued on <ISO timestamp> and was inactive for <duration>.`

The prefix up to the timestamp is **102** characters, so `.splitlines()[0]` does not remove it and the 120 cap admits ~18 characters of it. Two attempts on the same error code:

```
... Provider error: invalid_grant: AADSTS700082: The refresh token has expired due to inactivity. The token was issued on 2026-06-11T03:14:0...
... Provider error: invalid_grant: AADSTS700082: The refresh token has expired due to inactivity. The token was issued on 2026-07-29T22:01:5...
```

Both are ~295 characters, matching the estimate above, but they are different strings. On `main` those same two responses produced one byte-identical message, so for this code the change moves failure-summary cardinality from exactly 1 to unbounded — the grouping regression this thread set out to fix, on the code that https://github.com/airbytehq/airbyte-python-cdk/pull/1138#discussion_r3926561915 deliberately regrouped into the expired/revoked bucket.

**And no test can catch it.** Mutating the file one change at a time against the 86 tests in the two touched modules:

- delete `.splitlines()[0]` → **86 passed**. Every fixture's trace id sits past character 120, so the cap alone removes it, which makes both `trailing_marker not in exc_info.value.message` and `"Trace ID" not in ...message` vacuous.
- `_PROVIDER_ERROR_DETAIL_MAX_LENGTH = 120` → `2000` → **86 passed**. The truncation test asserts `len(provider_detail) == _PROVIDER_ERROR_DETAIL_MAX_LENGTH + len("...")` against the *imported* constant, so it can never pin its value, and its `"x" * 5000` fixture passes for any cap below ~5028.

Only the combined mutant fails, so the two guards mask each other and the realistic future regression — raising the cap to fit more description while keeping the first-line rule — ships green.

Options, in the order I'd prefer them:

1. **Append the provider code, not the prose.** `^[A-Za-z]{3,}\d{4,}` against the description yields `AADSTS700082`, which is the entire grouping key. Immune to sentence shape by construction, and it also bounds how much provider-controlled text lands in the persisted failure summary.
2. **Keep the prose, strip per-request values before truncating** — `re.sub(r"\d{4}-\d{2}-\d{2}T[\d:.]+Z?", ...)` plus a GUID pattern.
3. **Cut at the last sentence boundary at or before the cap**, so a partial timestamp can never terminate the string.

Whichever you pick, two fixtures would pin the property this thread is about, each verified to fail on exactly one of the mutants above:

- both timestamps in the first sentence and **no** later lines (first line 335 chars): assert `"AADSTS50173" in message` and `re.search(r"\d{4}-\d{2}-\d{2}T", message) is None`. Passes at 120, fails at 2000 regardless of `splitlines`.
- `error_description = "AADSTS50173: grant revoked.\r\nTrace ID: 00000000-0000-0000-0000-000000000000"` (collapsed length 89, under the cap): assert `"Trace ID" not in message`. Passes with `splitlines`, fails without it at any cap.

Plus the property itself, asserted directly: the same `AADSTS700082` payload twice with different issue dates, asserting the two `message` values are equal.

Reviewed at `6d5e37c9`. The offsets are against Microsoft's published `error_description` wording; the mechanism — a character cap cannot separate stable text from per-request text — does not depend on it.

bazarnov (commented):
Re-review of 1216c9d2 with `main` merged locally (clean - the three commits since the base do not touch the OAuth code). Review only, nothing pushed to the branch, no approval on this one by design.

Verified on this head:
- All five earlier threads are addressed; resolving them. The first-line/120 rule from my round is superseded by the code-only detail, and that is the right call: `AADSTS700082` carries the token issue timestamp in its first sentence, so no line rule or character cap could make it deterministic ([analysis](https://github.com/airbytehq/airbyte-python-cdk/pull/1138#discussion_r3936030689)).
- Independent mutation check against the two touched modules, one mutant at a time: drop `\b` >> 1 failed; unanchor the pattern (`.search`, no `^`) >> 1 failed; drop `_redact_credentials` from the detail >> 2 failed; cap 64 >> 2000 >> 1 failed; first-line prose instead of the code >> 4 failed; drop the `error` token >> 12 failed. Every guard is pinned by at least one test.
- Fleet check in airbytehq/airbyte: the only provider whose `error_description` leads with a code is Microsoft Entra (`AADSTS<n>:`). Google, Salesforce, GitLab, Airtable and Harvest return prose, so they degrade to `Provider error: invalid_grant` - one string per failure. No connector sets `refresh_token_error_key` to anything but `error` (gitlab, airtable only set it explicitly), so the ungated `error` read is bounded to the two default literals today. Nothing asserts on the exact wrapped message; nothing overrides `_wrap_refresh_token_exception`.
- Redaction precedes truncation in both builders, and `internal_message` is now redacted where the previous `e.response.text[:1000]` was not.
- CI: all required checks green on 1216c9d2. `Check: destination-motherduck` fails on every CDK PR (MotherDuck sandbox token), non-required.
- Local: 86 passed in the two modules, ruff 0.11.5 clean. The 9 errors in `TestSingleUseRefreshTokenOauth2Authenticator` are a local fixture-resolution problem identical on `main`; CI runs them green.

One item left, description only:
- The PR description still documents the two superseded formats: "How" describes `<error>: <error_description>` collapsed at 200 characters, and point 1 of the 2026-09-03 update describes the first line at 120. A short third note saying the detail is now `<error>[: <provider code>]`, capped at 64, with no prose fallback ([rationale](https://github.com/airbytehq/airbyte-python-cdk/blob/1216c9d2a9ce322d52799e722790afcb01721ef0/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L299-L313)) would make the description match what ships.

Notes:
- Residual theoretical case: a provider whose `error_description` leads with a hex or GUID-like token would surface that token as the "code" ([pattern](https://github.com/airbytehq/airbyte-python-cdk/blob/1216c9d2a9ce322d52799e722790afcb01721ef0/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py#L32)). None of the providers with fixtures in the monorepo does this, and the value is still redacted and capped, so no change requested - recording it for the next reader.
- Still a draft. #1145 edits the same test module in the `SingleUseRefreshToken` region plus `oauth.py`; no overlap with this diff.

bazarnov (approved):

PR #1167, opened 2026-09-17
fix: flush `sys.__stdout__` in `PrintBuffer.flush()`

ZaneHyattAB:
## Overview

<table><tr><td>

**:point_right: TL;DR:** Connector output written through the CDK's print buffer could sit inside the process for hours before the platform saw it, because the buffer wrote to the real stdout but never asked the interpreter to push those bytes into the pipe. This made healthy, progressing sources look silent to the platform heartbeat and to anyone reading the logs.

Specifically, `PrintBuffer.flush()` now calls `sys.__stdout__.flush()` after `sys.__stdout__.write()`.

</td></tr></table>

Related: airbytehq/oncall#12977 (internal, `source-stripe`). Requested by Zane Hyatt (@ZaneHyattAB).

## Changes

- `PrintBuffer.flush()` flushes the underlying `sys.__stdout__` `TextIOWrapper` after writing the batched content.
- New `unit_tests/utils/test_print_buffer.py`:
  - unit test asserting `flush()` writes to and flushes `sys.__stdout__`;
  - real-pipe regression test: a child process using the real CDK logger + `PRINT_BUFFER` (with `PYTHONUNBUFFERED` unset) emits a RECORD and a `logger.info` line, then sleeps; the parent asserts both lines arrive on the pipe while the child is still alive. Without the fix this test times out with zero bytes received.

## Review Spotlight

Reviewers with limited time, please review first:
- [`print_buffer.py#L54-L64`](https://github.com/airbytehq/airbyte-python-cdk/blob/devin/1789662473-print-buffer-flush-stdout/airbyte_cdk/utils/print_buffer.py#L54-L64): the one-line fix
- [`test_print_buffer.py#L27-L86`](https://github.com/airbytehq/airbyte-python-cdk/blob/devin/1789662473-print-buffer-flush-stdout/unit_tests/utils/test_print_buffer.py#L27-L86): the real-pipe regression test

## Why this is a correctness bug, not just a debuggability one

When stdout is a pipe (always, on the platform), `sys.__stdout__` is a block-buffered `TextIOWrapper` with an ~8 KiB pending buffer. `PrintBuffer.flush()` wrote into that buffer and returned, so bytes only reached the pipe when cumulative output crossed 8 KiB or the process exited. Measured with the real CDK logger and a real pipe: 230 B writes arrive in 8,280 B bursts; a RECORD + two `logger.info` lines emitted at t+1 s reached the reader only at process exit (t+9 s); with `PYTHONUNBUFFERED=1` they arrived at t+1.5 s.

The platform's source heartbeat counts messages it *receives*. For a low-volume or keepalive-only stream (e.g. one 213 B interim STATE per 600 s state-throttle window) filling 8 KiB takes ~36 messages ≈ 6 h, so with the default `heartbeat-max-seconds-between-messages` of 3 h a source that is progressing can be failed as silent. In the airbytehq/oncall#12977 investigation, one production job's observed 6h03m gap between two adjacent `logger.info` calls in `_on_stream_is_done` is exactly this buffer filling (36 × 600 s), not a stall — and three independent analyses of that job reached wrong conclusions from the log because of it.

## Design notes: why flush in `PrintBuffer.flush()` rather than `write_through` / `PYTHONUNBUFFERED`

<details>
<summary>Show/Hide Content</summary>

Options considered:

1. **`sys.__stdout__.flush()` inside `PrintBuffer.flush()` (chosen).** `PrintBuffer` is already the single choke point for stdout in `launch()`; its existing 0.1 s interval keeps batching for the `print()` path, and `logging.StreamHandler.emit` already calls `PrintBuffer.flush()` after each log line. Adding the fd flush there makes "left `PrintBuffer`" mean "reached the pipe", with syscall count still bounded by the interval (≤ 10/s for records) plus one per log line. Smallest, most contained change; nothing outside the entrypoint path is affected.
2. **`sys.stdout.reconfigure(write_through=True)` / line buffering at the entrypoint.** Equivalent syscall profile for the `PrintBuffer` path (each `PrintBuffer.flush()` already writes one combined string), but it is process-global: it also affects any code that writes to `sys.__stdout__` directly, and it must be applied in every entrypoint (`launch`, `source_declarative_manifest`, connector builder). Rejected as less contained for no benefit.
3. **`PYTHONUNBUFFERED=1` in the connector base image.** Correct and complementary, but outside this repository (base image / `airbyte` monorepo) and would not fix connectors pinned to older images. Suggested as a follow-up.

Residual behaviour after this change: messages written via `print()` within 0.1 s after the last flush stay in `PrintBuffer` until the next `write()` or `logger` call (the existing design). A message written after ≥ 0.1 s of idleness is flushed immediately, so a sparse keepalive stream now reaches the platform on emission.

</details>

## Throughput and syscall measurements

Benchmark: child process with stdout = real pipe, `init_logger("airbyte")` + `with PRINT_BUFFER:`, `print()` of 1,000,000 ~270 B RECORD lines plus one `logger.info` per 100k records; parent drains the pipe. Python 3.12, same machine, 3 runs each (child-side records/sec = records / time inside the `with` block).

| mode | records/sec (child), 3 runs | `write()` syscalls (strace -f -c, 1M records) |
|---|---|---|
| before (no fd flush) | 309,990 / 287,919 / 322,289 | 44 |
| after (this PR) | 321,574 / 316,277 / 307,660 | 45 |

Difference is within run-to-run noise; syscall count is unchanged (+1 per `logger` line, because those are now pushed through instead of coalescing with the next record batch). Per-line flushing was therefore not needed and the 0.1 s batching is preserved.

## Risks

- Each `logger.*` call now costs one `write()` syscall (previously it could coalesce into the 8 KiB buffer). With `LOG_LEVEL=debug` (request/response logging) this is one extra syscall per HTTP request — negligible against the request itself.
- Behaviour on a *full* pipe is unchanged: `write()` already blocked there when the 8 KiB buffer spilled; the flush just moves the block to the same call.

## Follow-ups

- Set `PYTHONUNBUFFERED=1` (or `-u`) in the Python connector base image so stderr and any direct `sys.__stdout__` writers get the same guarantee (outside this repo).
- Separate PR: HTTP connect/read timeouts in `HttpClient` (the other finding from airbytehq/oncall#12977).

## Test plan

- `poetry run pytest unit_tests/utils/test_print_buffer.py unit_tests/test_logger.py unit_tests/test_entrypoint.py -q` → 61 passed
- Verified the real-pipe test fails on `main` (times out after 5 s with `b''` received) and passes with this change.
- `poetry run ruff check` / `ruff format --check` on changed files; `poetry run mypy --config-file mypy.ini airbyte_cdk` → no issues.

## Suggested reviewers

@maxi297, @artem1205 (recent authors of `print_buffer.py` / `logger.py`). Not assigned on purpose; Zane will assign.

Link to Devin session: https://app.devin.ai/sessions/0e315592fd95478b8b4db2e122b927a8
Open in Devin Desktop: https://app.devin.ai/desktop/session/0e315592fd95478b8b4db2e122b927a8?variant=devin


airbyte_cdk/utils/print_buffer.py
```diff
@@ -52,9 +52,15 @@ def write(self, message: str) -> None:
...
     def flush(self) -> None:
+        """Writes the buffered content to the real stdout and pushes it through to the file descriptor.
+
+        `sys.__stdout__` is block-buffered when stdout is a pipe, so without the explicit
+        `flush()` the messages would stay in the interpreter's ~8 KiB text buffer until it fills up.
+        """
         with self.lock:
             combined_message = self.buffer.getvalue()
             sys.__stdout__.write(combined_message)  # type: ignore[union-attr]
+            sys.__stdout__.flush()  # type: ignore[union-attr]
```
darynaishchenko, line 63: 🟡 **low: log lines no longer get batched; each one now costs a `write` syscall**

The root logger's handler is a `logging.StreamHandler` whose stream is `PRINT_BUFFER` (`airbyte_cdk/logger.py:31-34`). `StreamHandler.emit` calls `stream.flush()` after **every** record, so each log line calls `PrintBuffer.flush()` directly, skipping the `flush_interval` check in `write()`. Before this change that was cheap (it copied into the userspace `sys.__stdout__` buffer). Now it's a real `os.write` per log line, and any pending records get pushed out with it. Records written via plain `print()` in `launch()` aren't affected.

I measured it with 1,000 messages, the interval disabled, and `sys.__stdout__` replaced by a TextIOWrapper → BufferedWriter(8 KiB) → counting raw stream (the same layering Python uses for a piped stdout):

| | log lines | records |
|---|---|---|
| before this PR | 8 writes | 1 |
| this PR | **1000 writes** | 1 |
| suggestion below | 8 writes | 1 |

**Suggested fix:** keep the external `flush()` cheap and only flush the fd on the interval and on exit:

```python
def write(self, message: str) -> None:
    with self.lock:
        self.buffer.write(message)
        current_time = time.monotonic()
        if (current_time - self.last_flush_time) >= self.flush_interval:
            self._flush_to_fd()
            self.last_flush_time = current_time

def flush(self) -> None:
    # Called by StreamHandler after every log record; drain without a syscall.
    with self.lock:
        self._drain()

def _drain(self) -> None:
    sys.__stdout__.write(self.buffer.getvalue())  # type: ignore[union-attr]
    self.buffer = StringIO()

def _flush_to_fd(self) -> None:
    self._drain()
    sys.__stdout__.flush()  # type: ignore[union-attr]

# __exit__: call self._flush_to_fd() instead of self.flush()
```

Trade-off: `print(..., flush=True)` (e.g. `http_client.py:529`) would no longer force immediate delivery. It would wait for the next interval, up to 0.1s.

darynaishchenko (commented):
The fix makes sense: without it, batches sit in the ~8 KiB `sys.__stdout__` buffer on a pipe instead of reaching the platform every `flush_interval`. One side effect for logs is noted inline.

airbyte_cdk/utils/print_buffer.py
ZaneHyattAB: @darynaishchenko  I went and measured how often connectors actually log once per record.

Nothing in the fleet logs per record on healthy data at default level. No certified Python connector, and the only 1:1 path in the CDK is TypeTransformer (transform.py:231-236), which needs schema_normalization: Default plus actual schema drift. Realistic ratio is about 0.01 per record, where everything is within noise. The flush itself costs 2.3 µs, and on a mock sync main and this PR are indistinguishable.

The gated variant does fix the throughput, but the quiet period line then only arrives at process exit and it fails the real pipe test, so it brings back the bug this PR is fixing.

So to answer your question directly: I'd like to keep the flush as written, since the 12% only shows up at a log rate nothing actually runs at. There is a way to get both and I'll file it as a follow up rather than grow this PR: keep the batching, add a daemon sweeper that flushes anything pending every 100 ms. That holds baseline throughput and syscalls, and the quiet line lands in 81 ms.

Context for why this one matters: at LOG_LEVEL=debug it does hit 2 to 3 lines per record on substreams that make one request per record, which is the shape of the Stripe issue behind this PR. That issue still has no confirmed root cause because buffered output made a live source look silent, and three separate analyses went down the wrong path as a result.

darynaishchenko (approved):

